PR #74, opened 2015-05-06
Added non-generic HttpOperationResponse to support responses without payload

stankovski:


stankovski:
@azuresdkci test this please

PR #87, opened 2015-06-01
Added serialization method to nodejs clientruntime

amarzavery:


ClientRuntimes/NodeJS/lib/serialization.js
```diff
@@ -0,0 +1,42 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+function serialize(toSerialize) {
```
yugangw-msft: since it is in common, please provide comments

ClientRuntimes/NodeJS/lib/serialization.js
```diff
@@ -0,0 +1,42 @@
...
+  if (!toSerialize) return;
```
yugangw-msft: should be "return null;" so to be more explict ?

ClientRuntimes/NodeJS/lib/validate.js
```diff
@@ -52,19 +52,18 @@ exports.isValidUuid = function(uuid, callback) {
...
 
 /**
-* Creates a anonymous function that check if a given key is base 64 encoded.
+* Validates if the given string is base64 encoded
 *
 * @param {string} key The key to validate.
-* @return {function}
+* @return {bool} 
 */
 exports.isBase64Encoded = function (key) {
```
yugangw-msft: please rename the "key" to be more general naming as we are not deal with any dictionary, so that we have to use "key". The "str" would be enough.

ClientRuntimes/NodeJS/lib/validate.js
```diff
@@ -52,19 +52,18 @@ exports.isValidUuid = function(uuid, callback) {
...
   var isValidBase64String = key.match('^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{4}|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)$');
-
+  var result = false;
   if (isValidBase64String) {
```
yugangw-msft: please simplify.
The entire function only need one line:
   return str,match(.....);

ClientRuntimes/NodeJS/test/serializationTests.js
```diff
@@ -0,0 +1,122 @@
+//
+// Copyright (c) Microsoft and contributors.  All rights reserved.
```
yugangw-msft: i don't think this is a valid license header

ClientRuntimes/NodeJS/test/serializationTests.js
```diff
@@ -0,0 +1,122 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+var should = require('should');
+var msRest = require('../lib/msRest');
+var assert = require('assert');
```
yugangw-msft: please sort

ClientRuntimes/NodeJS/lib/README.md
```diff
@@ -0,0 +1,16 @@
+# MS-Rest
+
+Infrastructure for error handling, tracing, and http client pipeline configuration. Required by nodeJS client libraries generated using AutoRest.
+
+- **Node.js version: 0.6.15 or higher**
```
markcowl: Is this the version we want?  Shouldn't we be requiring 0.10?

ClientRuntimes/NodeJS/lib/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "email": "azsdkteam@microsoft.com",
+    "url": "https://github.com/Azure/AutoRest"
+  },
+  "version": "0.1.1",
+  "description": "Client Runtime for nodeJS client libraries generated using AutoRest",
+  "tags": ["node", "microsoft", "autorest", "clientruntime"],
+  "keywords": ["node", "microsoft", "autorest", "clientruntime"],
+  "main": "./lib/msRest.js",
+  "engines": {
+    "node": ">= 0.6.15"
```
markcowl: Again, do we really work with 0.6.15?  Can we change this in a big way for the new runtime?

ClientRuntimes/NodeJS/lib/serialization.js
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Serializes the JSON Object. It serializes Buffer object to a 
+ * 'base64' encoded string and a Date Object to a string 
+ * compliant with ISO8601 format.
+ * 
+ * @param {Object} toSerialize
+ * 
+ * @returns {Object} serializedObject
+ */
+exports.serializeObject = function (toSerialize) {
```
markcowl: any reason not to just put the method inline here instead?

ClientRuntimes/NodeJS/lib/validate.js
markcowl: Do we need this function, now that there is no general deserialize method?

ClientRuntimes/NodeJS/lib/README.md
amarzavery: yes that is right

ClientRuntimes/NodeJS/lib/serialization.js
amarzavery: this has been fixed

PR #89, opened 2015-06-02
consistent assemblyinfo(s)

matt-gibbs:


ClientRuntimes/CSharp/ClientRuntime/Properties/AssemblyInfo.cs
```diff
@@ -6,18 +6,18 @@
...
 [assembly: AssemblyProduct("Microsoft AutoRest")]
-[assembly: AssemblyCopyright("Copyright (c) Microsoft Corporation")]
+[assembly: AssemblyCopyright("Copyright � Microsoft Corporation")]
+[assembly: AssemblyConfiguration("")]
 [assembly: AssemblyTrademark("")]
 [assembly: AssemblyCulture("")]
 [assembly: NeutralResourcesLanguage("en")]
-
-[assembly: ComVisible(false)]
 [assembly: CLSCompliant(false)]
```
yugangw-msft: It is not related with this PR. Since we are at it, do you know why we mark it at the assembly level that we are not comply with the CLS? For client applications, it doesn't matter, but it is the common library, and would be nice that our library could be used for other .net languages. Or at least we know which type violates it and we comment it here.

yugangw-msft:
lgtm

stankovski:
@matt-gibbs can you please resolve merge conflicts?

ClientRuntimes/CSharp/ClientRuntime/Properties/AssemblyInfo.cs
matt-gibbs: Don't know. Will check.

matt-gibbs:
Replacing this with a separate PR.
https://github.com/Azure/autorest/pull/110

PR #95, opened 2015-06-07
validate date and datetime per ISO8601 format and deserialize to Date() object if valid

amarzavery:


amarzavery:
tests for deserialization cover validation scenarios as well

PR #102, opened 2015-06-12
added retry policy and modifief redirect filter, optimized the serviceClient constructor

amarzavery:


ClientRuntimes/NodeJS/lib/filters/exponentialretrypolicyfilter.js
```diff
@@ -0,0 +1,179 @@
...
+
+/**
+ * Determines if the operation should be retried and how long to wait until the next retry.
+ *
+ * @param {number} statusCode The HTTP status code.
+ * @param {object} retryData  The retry data.
+ * @return {bool} True if the operation qualifies for a retry; false otherwise.
+ */
+function shouldRetry(statusCode, retryData) {
+  // Retry on HTTP TimeOut 408
```
yugangw-msft: update or remove it

ClientRuntimes/NodeJS/lib/filters/exponentialretrypolicyfilter.js
```diff
@@ -0,0 +1,179 @@
...
+  if ((statusCode < 500 && statusCode !== 408) || statusCode === 501 || statusCode === 505) {
+    return false;
+  }
+
+  var currentCount = (retryData && retryData.retryCount) ? retryData.retryCount : 0;
```
yugangw-msft: My understanding is the "retryData" will never be a null since we always init it; if yes, then we should drop the check.  Because when it turns out be null, throwing is better than entering an potential infinite loop.

ClientRuntimes/NodeJS/lib/filters/exponentialretrypolicyfilter.js
```diff
@@ -0,0 +1,179 @@
...
+
+    retryData.error = err;
+  }
+
+  // Adjust retry count
+  retryData.retryCount++;
+
+  // Adjust retry interval
+  var incrementDelta = Math.pow(2, retryData.retryCount) - 1;
+  var boundedRandDelta = this.retryInterval * 0.8 + Math.floor(Math.random() * (this.retryInterval * 1.2 - this.retryInterval * 0.8));
```
yugangw-msft: apply line break after "* 0.8"? too long.

ClientRuntimes/NodeJS/lib/filters/exponentialretrypolicyfilter.js
```diff
@@ -0,0 +1,179 @@
...
+
+  return (currentCount < this.retryCount);
+}
+
+/**
+ * Updates the retry data for the next attempt.
+ *
+ * @param {object} retryData  The retry data.
+ * @param {object} err        The operation's error, if any.
+ * @return {undefined}
```
yugangw-msft: wonder whether this is the convention for java script that using "undefined" if there is no return value. I would just get rid of it

ClientRuntimes/NodeJS/lib/filters/exponentialretrypolicyfilter.js
```diff
@@ -0,0 +1,179 @@
...
+* Creates a new 'ExponentialRetryPolicyFilter' instance.
+*
+* @constructor
+* @param {number} retryCount        The client retry count.
+* @param {number} retryInterval     The client retry interval, in milliseconds.
+* @param {number} minRetryInterval  The minimum retry interval, in milliseconds.
+* @param {number} maxRetryInterval  The maximum retry interval, in milliseconds.
+*/
+function ExponentialRetryPolicyFilter(retryCount, retryInterval, minRetryInterval, maxRetryInterval) {
+  // Implement the new style filter in terms of the old implementation
```
yugangw-msft: what is the "old implementation"?

ClientRuntimes/NodeJS/lib/serviceClient.js
```diff
@@ -8,34 +8,48 @@ var Constants = require('./constants');
...
-function ServiceClient(credentials, filters) {
-  if (!filters) {
-    filters = [];
+ * @class
+ * Initializes a new instance of the ServiceClient class.
+ * @constructor
+ * @param {object} options The parameter options
+ * 
+ * @param {object} [options.credentials]    - BasicAuthenticationCredentials or 
+ * TokenCredentials object used for authentication.
```
yugangw-msft: my suggestion is to drop the specific derived type names. Because in near future if we support new type, or for any reason to rename, likely. This comment will become invalid without notice. Updating common is always not cheap, as we need refresh all dependents downstream.

ClientRuntimes/NodeJS/lib/utils.js
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information. 
 
 var Constants = require('./constants');
+var _ = require('underscore');
```
yugangw-msft: move this above the "Constants" as this is 3rd party ones

ClientRuntimes/NodeJS/package.json
```diff
@@ -25,7 +25,8 @@
     "request": "2.45.0",
     "validator": "~3.1.0",
     "duplexer": "~0.1.1",
-    "through": "~2.3.4"
+    "through": "~2.3.4",
+    "tough-cookie": "*"
```
yugangw-msft: apologize if i miss anything, but i don't see we are using this package in this PR

amarzavery: this is required if someone enables cookie in request

amarzavery: jar - If true and **tough-cookie is installed**, remember cookies for future use (or define your custom cookie jar; see examples section) as per request.js doc

PR #131, opened 2015-06-26
Update LRO code to preserve the latest resource location header value.

ogail:


ogail:

PR #136, opened 2015-06-29
Inherit AzureNodeJSCodeGenerator from generic NodeJSCodeGenerator

amarzavery:


AutoRest/Generators/Azure.Common/Azure.Common.Tests/AzureServiceClientNormalizerTests.cs
```diff
@@ -165,6 +166,31 @@ public void SwaggerODataSpecParsingTest()
...
         [Fact]
+        public void SwaggerResourceExternalFalseTest()
+        {
+            var settings = new Settings
+            {
+                Namespace = "Test",
+                Input = @"Swagger\resource-external-false.json"
+            };
+
+
```
hovsepm: please remove whitespace

PR #140, opened 2015-07-01
Updating FxCop rules and fixing tests

markcowl:


AutoRest/Generators/Azure.Common/Azure.Common.Tests/AutoRest.Generator.Azure.Common.Tests.csproj
```diff
@@ -17,6 +17,9 @@
     <AutoRestTest>true</AutoRestTest>
     <NuGetPackageImportStamp>22573d97</NuGetPackageImportStamp>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)' == 'Net45-Debug'">
```
yugangw-msft: don't need this condition, just make the rule set across all configuration

yugangw-msft: please do the same for all other files

AutoRest/Generators/Azure.Common/Azure.Common.Tests/Properties/AssemblyInfo.cs
```diff
@@ -5,7 +5,7 @@
 using System.Reflection;
 using System.Runtime.InteropServices;
 
-[assembly: AssemblyTitle("AutoRest.Generator.Azure.CSharp.Tests")]
+[assembly: AssemblyTitle("AutoRest.Generator.Azure.CommonS.Tests")]
```
yugangw-msft: Typo? the "trailing S"?

AutoRest/Generators/CSharp/Azure.CSharp.Tests/AutoRest.Generator.Azure.CSharp.Tests.csproj
```diff
@@ -16,6 +16,9 @@
     <AutoRestTest>true</AutoRestTest>
     <NuGetPackageImportStamp>2c7e2ccf</NuGetPackageImportStamp>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)' == 'Net45-Debug'">
+    <CodeAnalysisRuleSet>MinimumRecommendedRules.ruleset</CodeAnalysisRuleSet>
```
yugangw-msft: to avoid the code duplicate.
I suggest we do it below https://github.com/Azure/autorest/blob/dev/tools/AutoRest.Settings.targets#L33
<CodeAnalysisRuleSet Condition=" $(AutoRestTest) == 'true' ">MinimumRecommendedRules.ruleset</CodeAnalysisRuleSet>

Tools/TestRules.ruleset
```diff
@@ -0,0 +1,423 @@
...
+    <Rule Id="CA3147" Action="None" />
+    <!-- Microsoft namespace rules do not apply to test code-->    
+    <Rule Id="CA904" Action="None" />
+    <!-- Microsoft namespace rules do not apply to test code-->
+    <Rule Id="CA905" Action="None" />
+    <!-- Assemblies are not ngen'd, therefore pre-compilation rules for generics do not apply-->
+    <Rule Id="CA908" Action="None" />
+  </Rules>
+  <!-- Rules for native code analysis are not needed for Managed Code-->
+  <Rules AnalyzerId="Microsoft.Analyzers.NativeCodeAnalysis" RuleNamespace="Microsoft.Rules.Native">
```
yugangw-msft: I don't think we need this

yugangw-msft: ah, i took it back,  we just disabled them

AutoRest/Generators/Azure.Common/Azure.Common.Tests/AzureServiceClientNormalizerTests.cs
```diff
@@ -33,7 +33,36 @@ public void ResourceIsFlattenedForSimpleResource()
 
             resource.Name = "resource";
             resource.Extensions[AzureCodeGenerator.ExternalExtension] = true;
-
+            resource.Properties.Add(new Property
```
yugangw-msft: Not familiar in the area, but this will fail the following validation code

AutoRest/Generators/Azure.Common/Azure.Common/AzureCodeGenerator.cs
```diff
@@ -332,12 +332,16 @@ public static bool IsDerivedFromExternalResource(CompositeType compositeType)
...
-            return compositeType.BaseModelType != null &&
-                   (compositeType.BaseModelType.Name.Equals(ResourceType, StringComparison.OrdinalIgnoreCase) ||
-                    compositeType.BaseModelType.Name.Equals(SubResourceType, StringComparison.OrdinalIgnoreCase)) &&
-                   compositeType.BaseModelType.Extensions.ContainsKey(ExternalExtension) &&
-                   compositeType.BaseModelType.Extensions[ExternalExtension] is bool &&
-                   (bool)compositeType.BaseModelType.Extensions[ExternalExtension];
+            if (compositeType.BaseModelType != null &&
+                (compositeType.BaseModelType.Name.Equals(ResourceType, StringComparison.OrdinalIgnoreCase) ||
+                 compositeType.BaseModelType.Name.Equals(SubResourceType, StringComparison.OrdinalIgnoreCase)) &&
+                compositeType.BaseModelType.Extensions.ContainsKey(ExternalExtension))
```
yugangw-msft: maybe this is intentional, but this changes the condition that it doesn't consider the "bool", hence CI failed

markcowl: It actually does - the intention is to make it wpork for

x-ms-external

-or -

x-ms-external: true

AutoRest/Generators/Azure.Common/Azure.Common.Tests/AzureServiceClientNormalizerTests.cs
markcowl: This is now required for external resources

AutoRest/Generators/Azure.Common/Azure.Common/AzureCodeGenerator.cs
markcowl: The intention is to accept:

x-ms-external

-or

x-ms-external: true

but if 

x-ms-external: false is set, then this would be disabled

PR #141, opened 2015-07-01
Azure common lro support for "noheader", subresource, nonresource

yugangw-msft:
The work is to port C# version to node and then add tests (unit test and acceptance) for both C#, Node


yugangw-msft:
CI failure caused by fxcop error from generated code in the test projects. Rather using loads of suppression, I will wait for the #140 merged first which will use minimum rule set on the test project so to emit out less errors

yugangw-msft:
merged, as it went through several pair session and 4 people have taken a look at the code

PR #156, opened 2015-07-02
renamed generated nodejs files with camelcase naming convention

amarzavery:


matt-gibbs:
test slack integration

PR #163, opened 2015-07-03
Fixed cloning of the Method object

nicklebedev37:
Currently when we create a LRO method - we copy it from the existing one. But the copy isn't "deep" enough. So when I update parameter of the LRO method - I'm updating parameter of the method it was cloned too (and vice versa).
This was revealed in the Ruby code generator because I was applying CodeNaming framework on method parameters.

### DIFF
--- a/AutoRest/AutoRest.Core/ClientModel/Method.cs
+++ b/AutoRest/AutoRest.Core/ClientModel/Method.cs
@@ -123,7 +123,7 @@ namespace Microsoft.Rest.Generator.ClientModel
             newMethod.RequestHeaders = new Dictionary<string, string>();
             newMethod.Responses = new Dictionary<HttpStatusCode, IType>();
             this.Extensions.ForEach(e => newMethod.Extensions[e.Key] = e.Value);
-            this.Parameters.ForEach(p => newMethod.Parameters.Add(p));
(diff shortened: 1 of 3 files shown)


ogail:
@nicklebedev37 This is a valid change but can u please explain what is the scenario u have when u need to modify the LRO method but not the original?

nicklebedev37:
@ogail yes, sure,

please take a look at the following line: https://github.com/nicklebedev37/autorest/blob/dev/AutoRest/AutoRest.Core/CodeNamer.cs#L129
(It reflects a version of AutoRest of the state of week ago)

Here we iterate the list of methods. One of the methods here is LRO, it is a clone of another method from this collection. So let's assume that we firstly iterate initial method. Let's assume it has parameter with name "thisIsParameter"; this parameter gets SerializedName - "thisIsParameter" and after it has been applied function GetParameterName it changes its Name to "this_is_parameter". Since the cloning of method wasn't deep enough - the same became applied to the clone of this method. So when in the next iteration we update SerializedName - we don't get "thisIsParameter" but we get "this_is_parameter". Obviously it is not expected.

This situation was fixed in the commit https://github.com/Azure/autorest/commit/c1da6afff121766163ac9cdc8b33bc0a1670c114. But I would rather ask to get it merged anyway to prevent any further similar bugs.

PR #166, opened 2015-07-06
[DO NOT MERGE] Platform agnostic build process

devigned:
This pr is a start toward an agnostic build process. There are a few things that we need to review related to Warnings as errors from unused vars and some class which are not available in Mono (EventAttribute and System.Net.PeerToPeer). Also, /r/n was replaced in expected with /n likely due to platform dependent template generation. Due to the mass of the change and the issues outlined above, I'd like to have a solid review and chat about the design direction.

- Use gulp for build process
- Remove powershell build scripts and replace with gulp tasks (nuspec / nuproj sync logic)
- Make the .proj files xbuild and msbuild compatible and adaptable to current platform
- Ran a build and all sort of files changed :boom:. Ouch. This happened because of **committed code** that is **generated** (platform dependent line endings).
- Deleted multiple AssemblyInfo DefaultImportDLL statement, which should probably be reviewed
- Commented out ETW stuff due to lack of support in Mono
- Added some silly .ToString() on some tests b/c the compiler was choking on unused var warnings b/c we are testing constructers throwing errors on input.


AutoRest/Generators/CSharp/Azure.CSharp/Properties/AssemblyInfo.cs
```diff
@@ -4,7 +4,7 @@
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
-// General Information about an assembly is controlled through the following 
+// General Information about an assembly is controlled through the following
 // set of attributes. Change these attribute values to modify the information
 // associated with an assembly.
 [assembly: AssemblyTitle("AutoRest")]
```
yugangw-msft: we can be more specific to call out the it is "AutoRest C# code generator"?

yugangw-msft: also we can fill in more for several other fields below, such as "AssemblyCompany"

devigned: I do agree with you, though those were not part of my changes and probably should be part of another pull request.

PR #180, opened 2015-07-08
Added automation for testing NuGet packages

yugangw-msft:


AutoRest/Generators/AcceptanceTests/NugetPackageTest/packages.config
```diff
@@ -0,0 +1,15 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Microsoft.Rest.ClientRuntime" version="1.0.2" targetFramework="net45" />
+  <package id="AutoRest" version="0.10.0" targetFramework="net45" />
+  <package id="Microsoft.Bcl" version="1.1.10" targetFramework="net45" />
```
yugangw-msft: i will clean them up before merge. The test project is 4.5 and we do not need them

yugangw-msft:
In the upcoming PR, i will add a new logic of emitting out a better error when packages to test is unavailable which is usually caused by you run "test" w/o running "package" first

PR #183, opened 2015-07-09
add node code-gen assemblies into the package

yugangw-msft:
Notes:
1. I also put all the assemblies in the tools folder
2. The package test to verify the node js code-gen is being written and will get in through a separate PR. That involves more testing file change, for risk control, I submit it later today.

### DIFF
--- a/AutoRest/AutoRest.Core/AutoRest.nuspec
+++ b/AutoRest/AutoRest.Core/AutoRest.nuspec
@@ -16,11 +16,12 @@
     <tags>Microsoft AutoRest</tags>
   </metadata>
   <files>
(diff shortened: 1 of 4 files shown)


yugangw-msft:
Note, with assemblies get moved over to under the tools folder, nuget.exe will report following warnings, which is by-design. But I take a look to see whether we can suppress it.  
   EXEC : warning : 5 issue(s) found with package 'autorest'. [D:\sdk\autorest\build.proj]
   Issue: Assembly outside lib folder.
  Description: The assembly 'tools\AutoRest.Core.dll' is not inside the 'lib' folder and hence it won't be added as reference when the package is installed into a project.
  Solution: Move it into the 'lib' folder if it should be referenced.

PR #208, opened 2015-07-15
Adding AD credentials to Azure runtime

markcowl:
AD Credentials and tests added to azure runtime
Change azure generator to use new credentials classes


markcowl:

PR #212, opened 2015-07-15
Added PageCollection to ClientRuntime

hovsepm:


stankovski:
@hovsepm how about method signature for ListNext()? Should the argument be called "nextPageLink" instead of "nextLink"?

PR #232, opened 2015-07-20
Bumped up the version of Autorest 

### LINKED ISSUE #214: Discussion: 'properties' property is embbeded in the request even if user didn't specify any value
I'm trying to call PATCH on a resource like below.
```C#
                // Workflow is derived from Resource
                workflow = new Workflow()
                    {
                        Tags = new Dictionary<string, string>()
                    };
                workflow.Tags.Add("abc", "def");
                workflow = client.Workflows.Update(this.resourceGroupName, workflowName, workflow);
```

The actual request body is like below, which causes an error because 'properties' was set to an empty object.

```JSON
{
  "tags": {
    "abc": "def"
  },
  "properties": {}
}
```

Expected request body is like below:
```JSON
{
  "tags": {
    "abc": "def"
  }
}
```

Is it possible to have generated SDK not emitting 'properties' in this case? Bottom line, I could define the same model again with a different name that is *not* derived from Resource in order to workaround the problem caused by flattening the properties.

hovsepm:
Fixed #214


azurecla:
Hi __@hovsepm__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (hovsepm). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

PR #238, opened 2015-07-21
Use correct portable version of adal and some other misc minor fixes

yugangw-msft:
otherwise, store application will not work with runtime. 
I don't have phone app, store app to verify the change at runtime for portable configurations, but i did check out the source code of adal that a null "IPlatformParameters" is accepted


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

PR #259, opened 2015-07-29
Authentication glue to azure nodejs client runtime and other nodejs code gen fixes

amarzavery:


azurecla:
Hi __@amarzavery__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (amzavery). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

ClientRuntimes/NodeJS/ClientRuntime.Azure/lib/credentials/userTokenCredentials.js
```diff
@@ -0,0 +1,97 @@
...
+* @param {string} username The user name for the Organization Id account.
+* @param {string} password The password for the Organization Id account.
+* @param {string} clientRedirectUri The Uri where the user will be redirected after authenticating with AD.
+* @param {object} [options] Object representing optional parameters.
+* @param {AzureEnvironment} [options.environment] The azure environment to authenticate with.
+* @param {string} [options.authorizationScheme] The authorization scheme. Default value is 'bearer'.
+* @param {object} [options.tokenCache] The token cache. Default value is null.
+*/
+function UserTokenCredentials(clientId, domain, username, password, clientRedirectUri, options) {
+  if (!Boolean(clientId) || typeof clientId !== 'string') {
```
yugangw-msft: if we want to use this style, then we should use it across, otherwise this code is inconsistent with the rest. My preference is just to drop "Boolean"

AutoRest/Generators/NodeJS/NodeJS/TemplateModels/ModelTemplateModel.cs
```diff
@@ -61,6 +61,11 @@ public IEnumerable<Property> ComposedProperties
             }
         }
 
+        public List<T> ConvertIEnumerableToList<T>(IEnumerable<T> enumerable)
```
yugangw-msft: we can remove this one as this is not used any more

AutoRest/Generators/NodeJS/Azure.NodeJS.Tests/Expected/AcceptanceTests/Lro/models/product.js
```diff
@@ -59,10 +58,14 @@ Product.prototype.validate = function (payload) {
  */
 Product.prototype.deserialize = function (instance) {
   if (instance) {
+
```
yugangw-msft: we might have a code-gen bug

ClientRuntimes/NodeJS/ClientRuntime.Azure/lib/credentials/userTokenCredentials.js
yugangw-msft: We will have to change. if user use "new String()", our code will throw by mistake.
Suggest just simplify by verifying the value is set. We might catch less errors, but better than throw on valid cases.

PR #267, opened 2015-07-30
Reworked MsRest(Azure) to use Faraday

nicklebedev37:
The PR includes:

1. Transition of MsRest to Faraday.
2. Implementation of automatic token refresh.
3. Implementation of AzureApplicationCredentials. It's much simper than CSharp's one since it misses authentication via user id but only allows to authenticate only via clientid/clientsecret/tenantid. Also it doesn't have token provider object model. The token retrieval is encapsulated into AzureApplicationCrendetials object.
4. Various minor stylistic/serialization/namespace fixes.


azurecla:
Hi __@nicklebedev37__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (nick.lebedev@akvelon.com). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

stankovski:
Cortana retest this please

ClientRuntimes/Ruby/ms-rest-azure/lib/ms_rest_azure/azure_application_credentials.rb
```diff
@@ -0,0 +1,93 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+
+module MsRestAzure
+  #
+  # Class which represents an Azure token credentials.
+  #
+  class AzureApplicationCredentials < MsRest::ServiceClientCredentials
+
+    TOKEN_ACQUIRE_URL = 'https://login.windows.net/{tenant_id}/oauth2/token'
```
markcowl: The login authority should be settable, it is different for azure china, azure stack, government cloud

ClientRuntimes/Ruby/ms-rest-azure/lib/ms_rest_azure/azure_application_credentials.rb
```diff
@@ -0,0 +1,93 @@
...
+    REQUEST_BODY_PATTERN = 'resource={resource_uri}&client_id={client_id}&client_secret={client_secret}&grant_type=client_credentials'
+    RESOURCE_URI = 'https://management.core.windows.net/'
```
markcowl: the token audience should also be settable - both graph.windows.net and keyvault require a different audience, as do azure china, azure stack, etc.

ClientRuntimes/Ruby/ms-rest-azure/lib/ms_rest_azure/token_refresh_middleware.rb
```diff
@@ -0,0 +1,57 @@
...
+module MsRestAzure
+  #
+  # Class which handles token renewal.
+  #
+  class TokenRefreshMiddleware < Faraday::Response::Middleware
+    #
+    # Initializes a new instance of the TokenRefreshMiddleware class.
+    #
+    def initialize(app, options = nil)
+      fail ArgumentError, 'options can\'t be nil' if options.nil?
```
markcowl: cannot

ClientRuntimes/Ruby/ms-rest-azure/lib/ms_rest_azure/token_refresh_middleware.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      begin
+        response_body = JSON.load(response.body)
+        error_code = response_body['error']['code']
+        error_message = response_body['error']['message']
+      rescue Exception => e
+        return false
+      end
+
+      return (error_code == 'AuthenticationFailed' && (error_message.start_with?('The access token expiry') || (error_message.start_with?('The access token is missing or invalid'))))
```
markcowl: i'm not sure how robust this is over changes in the autrhentication service. I woulkd suggest checking the token expiration (which is returned with the access token) when receiving a 401, instead of relying on message text.  You *mightr* also consider pre-emptively checking expiration before sending the request, instead of waiting for the 401

devigned:
Would you move the duplicated serialize and deserialize methods into a module in ms-ruby-azure? This is repeated all over the place in the generated code: https://github.com/Azure/azure-sdk-for-ruby/pull/257/files#diff-32406c0ddfae8419333805efbc957fcaR26.

nicklebedev37:
Am I correct that you're suggesting to replace methods like:

```Ruby
deserialized_property = object['computerName']
output_object.computer_name = deserialized_property
```

to

```Ruby
get_property(output_object, object, 'computer_name', 'computerName')
```

and this method - gets property from object and assigns it to output_object?

I was thinking about it. But with your approach:
* It's a little bit more difficult to generate code, since sometimes the code won't be such plain, e.g. when complex serialization is required. And having temporary field here is very useful. E.g. here - 74e6d43acccc473e5d3f172f331ab8cbe3a7e205 
* It didn't look to me that it's a good win to have one line instead of two. Also when one line is a method invocation with 4 parameters.

Please let me know if there is some good approach of elaborating these duplicate property assignments.

nicklebedev37:
Updated PR but not all code review notes are fixed yet.

nicklebedev37:
@markcowl I've reworked expiration check to use preemptive checking. Also I've adjusted credentials/token_provider object model to be closer to CSharp one but it still more simplified and have less functionality (e.g. doesn't have Azure user authentication but only application authentication)

Could you take a look and review it please?

https://github.com/Azure/autorest/pull/267/files#diff-76258d1b8ba309e25a463d03e4f210ceR67

nicklebedev37:
Cortana retest this please

PR #311, opened 2015-08-20
Removing the requirement for the 'get' operation when declare Azure LRO in the swagger doc

yugangw-msft:
"Get" was needed for a while when generated code needed to pass in a generic getResoruce callback, but turned out that is not necessary as the azure-common can get the resource url from the initial request and use that to poll for the resource. C# has done that improvement, but Node.js was not covered till this PR.


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

ljhljh235:
LGTM as long as CI passing

PR #315, opened 2015-08-26
Fixing various setup and build issues

brjohnstmsft:
- Bumped Microsoft.Rest.ClientRuntime package version to 1.3.0 in
  NugetPackageTest to fix test failure.
- Suppressed Code Analysis error that was breaking the build
- Swapped expected and actual in AcceptanceTests.EnsureTestCoverage
  since the output was misleading when it failed.
- Added missing Ruby setup step to README and changed instructions
  for running gulp. 'gulp build test' fails because 'packages' is
  actually a prerequisite of 'test', so just running 'gulp' to invoke
  the default task is safer.
- Added msbuild targetVersion to gulpfile to prevent gulp-msbuild from
  loading an older version of msbuild. This was causing the C# URL tests
  to fail since URL parsing behavior is different between .NET 4.0 and
  .NET 4.5. See the Remarks for System.Uri on MSDN for more details:
  https://msdn.microsoft.com/en-us/library/system.uri(v=vs.110).aspx

### DIFF
--- a/AutoRest/AutoRest.Core/Extensibility/ExtensionsLoader.cs
+++ b/AutoRest/AutoRest.Core/Extensibility/ExtensionsLoader.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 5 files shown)


azurecla:
Hi __@brjohnstmsft__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (brjohnst). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

devigned:
Thank you for the pr @brjohnstmsft!

PR #333, opened 2015-09-03
Fix for issue #293 - segment concatenation in URLs

### LINKED ISSUE #293: Urls Incorrect when BaseUri contains Segments
The generated code builds the url like this:

    var url = new Uri(this.Client.BaseUri, "/myroute/{id}").ToString();

However this creates incorrect urls if ```BaseUri``` contains segmets for example ```http://somesite/segment1``` the resulting url will drop the ```segment1``` and use the base address only.
The resulting url is therefore ```http://somesite/myroute/{id}```.

This is because the beginning of relativeUri begins with a ```/```

It should be:

    var url = new Uri(this.Client.BaseUri, "myroute/{id}").ToString();

This was introduced in 11.0 as it worked perfectly in 10.0

stankovski:
Fixes #293


azurecla:
Hi __@stankovski__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (destanko). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

PR #351, opened 2015-09-21
Fix x-ms-pagable and Produces for "application/json; ..."

xingwu1:
1. Change "x-ms-pageable" extension to use "string" instead of bool literal  

In ODATA, the nextLink string in response is "odata.nextLink", so the extension should be:
"x-ms-pageable" : "odata.nextLink"

2. JSON is part of produces string is good enough

For example: this produces should be good enough to present for JSON request body.

 "produces": [
    "application/json; odata=minimalmetadata"
  ]

### DIFF
--- a/AutoRest/AutoRest.Core.Tests/Resource/RedisResource.json
+++ b/AutoRest/AutoRest.Core.Tests/Resource/RedisResource.json
@@ -12,7 +12,7 @@
     "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Cache/Redis": {
       "get": {
         "description": "",
-        "x-ms-pageable": "true",
+        "x-ms-pageable": "nextLink",
         "operationId": "List",
         "consumes": [],
         "produces": [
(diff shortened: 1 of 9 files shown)


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

devigned:
@xingwu1 
```json
"x-ms-pageable" : "odata.nextLink"
```
Could we change this to be an object value rather than a string. It'll make it more extensible in the future.

Perhaps like this:
```json
"x-ms-pageable" : { "path": "odata.nextLink" }
```
Where path is the location within the payload relative to root of the payload document with keys delimited by '/'. If the next page link would be "paging/next" if the key was within the paging object with the next key containing the value of the next link, which should be a uri.

Thoughts?

brjohnstmsft:
What version of OData is this change targeting? FWIW, in OData V4 it's called `@odata.nextLink` (note the '@'), and the Content Type is `application/json; odata.metadata=minimal`

xingwu1:
I think our server is using OData v3. But the fix is intent to general usage to any possible nextLink string.

devigned:
@xingwu1 any thoughts on the object rather than the string?

xingwu1:
I think the object idea is good. I will work on it. One extra property for this object could be the name of ListNext function name (instead of hard code "Next" at end of the original function name). However, I don't know whether there has real scenario that the "nextLink" location is not under the root.

xingwu1:
Somehow mess the branch. Close this pull request and reopen new open (https://github.com/Azure/autorest/pull/360)

PR #353, opened 2015-09-22
[DO NOT MERGE] Add "client-request-id" support in AutoRest

matthchr:
Existing AutoRest has a hardcoded "x-ms-clientrequestid" but that is not sufficient for all services (as some use "client-request-id" instead.  

This change adds an additional header to all requests.

If somebody has a suggestion for avoiding having 2 client request IDs all the time (which is confusing...) I'm open to it.


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

brjohnstmsft:
Is there a way to let the caller specify client-request-id instead of always using a new GUID?

matthchr:
@brjohnstmsft I think the intent is that the client-request-id (or x-ms-clientrequestid for that matter) should be specified as an optional header in the swagger spec.  That way it is just like any other header and would show up in the method signatures.  If it was specified in the method signature, then we would generate a guid, which would then be overwritten by the client specified header of the same name.

devigned:
@Azure/adx-sdk-team thoughts on always providing two headers here?

matthchr:
We discussed in our meeting and I will investigate a swagger extension for this to handle the return header case too.  I think that means we can hold off on merging this PR...

PR #359, opened 2015-09-24
Fix bug: any one of the produces start with "application/json" is enough.

xingwu1:
Any one of the produces start with "application/json" is enough to prove operation produce JSON. In our ODATA server, the produce has to be "application/json; odata=minimalmetadata", which will fail in original check.


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

PR #368, opened 2015-09-28
Added TypeScript support, generating .d.ts files alongside NodeJS JavaScript

BretJohnson:
Here's the updated pull request for TypeScript support. Now the d.ts files are consolidated, so it'll generate at most 3 d.ts files: .d.ts, models/index.d.t.s, and operations/index.d.ts.

d.ts generation is pretty harmless, but to disable, if that's desired for any reason, can do this:
Change NodeJSCodeGenerator.DisableTypeScriptGeneration to true
Don't include index.d.ts addition in ms-rest npm update (though, again, it's pretty harmless if included)

### DIFF
--- a/AutoRest/Generators/NodeJS/NodeJS/ClientModelExtensions.cs
+++ b/AutoRest/Generators/NodeJS/NodeJS/ClientModelExtensions.cs
@@ -242,6 +242,28 @@ namespace Microsoft.Rest.Generator.NodeJS.TemplateModels
             }
         }
 
+        /// <summary>
+        /// Returns the TypeScript type string for the specified primary type
+        /// </summary>
+        /// <param name="primary">primary type to query</param>
+        /// <returns>The TypeScript type correspoinding to this model primary type</returns>
+        private static string PrimaryTSType(this PrimaryType primary) {
+            if (primary == PrimaryType.Boolean)
+                return "boolean";
+            else if (primary == PrimaryType.Double || primary == PrimaryType.Int || primary == PrimaryType.Long)
+                return "number";
+            else if (primary == PrimaryType.String)
+                return "string";
+            else if (primary == PrimaryType.Date || primary == PrimaryType.DateTime)
+                return "Date";
+            else if (primary == PrimaryType.Object)
+                return "Object";   // TODO: test this
+            else {
+                throw new NotImplementedException(string.Format(CultureInfo.InvariantCulture,
+                    "Type '{0}' not implemented", primary));
+            }
+        }
+
         private static string ValidateEnumType(this EnumType enumType, IScopeProvider scope, string valueReference, bool isRequired)
         {
             if (scope == null)
@@ -424,6 +446,48 @@ namespace Microsoft.Rest.Generator.NodeJS.TemplateModels
 
             return null;
         }
+		
+        /// <summary>
+        /// Return the TypeScript type (as a string) for specified type.
+        /// </summary>
+        public static string TSType(this IType type, bool inModelsModule) {
+            CompositeType composite = type as CompositeType;
+            SequenceType sequence = type as SequenceType;
+            DictionaryType dictionary = type as DictionaryType;
+            PrimaryType primary = type as PrimaryType;
+            EnumType enumType = type as EnumType;
+
+            string tsType;
+            if (primary != null)
+            {
+                tsType = primary.PrimaryTSType();
+            }
+            else if (enumType != null)
+            {
+                tsType = "string";
+            }
+            else if (composite != null)
+            {
+                if (inModelsModule)
+                    tsType = composite.Name;
+                else tsType = "models." + composite.Name;
+            }
+            else if (sequence != null)
+            {
+                tsType = sequence.ElementType.TSType(inModelsModule) + "[]";
+            }
+            else if (dictionary != null)
+            {
+                // TODO: Confirm with Mark exactly what cases for additionalProperties AutoRest intends to handle (what about
+                // additonalProperties combined with explicit properties?) and add support for those if needed to at least match
+                // C# target level of functionality
+                tsType = "{ [propertyName: string]: " + dictionary.ValueType.TSType(inModelsModule) + " }";
+            }
+            else throw new NotImplementedException(string.Format(CultureInfo.InvariantCulture, "Type '{0}' not implemented", type));
+
+            return tsType;
+        }
+		
 
         private static string SerializePrimaryType(this PrimaryType primary, IScopeProvider scope, string objectReference, string valueReference, bool isRequired)
         {
--- a/AutoRest/Generators/NodeJS/NodeJS/ClientModelExtensions.cs
+++ b/AutoRest/Generators/NodeJS/NodeJS/ClientModelExtensions.cs
@@ -257,7 +257,7 @@ namespace Microsoft.Rest.Generator.NodeJS.TemplateModels
             else if (primary == PrimaryType.Date || primary == PrimaryType.DateTime)
                 return "Date";
             else if (primary == PrimaryType.Object)
-                return "Object";   // TODO: test this
+                return "any";   // TODO: test this
             else {
                 throw new NotImplementedException(string.Format(CultureInfo.InvariantCulture,
                     "Type '{0}' not implemented", primary));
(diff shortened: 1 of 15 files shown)


azurecla:
Hi __@BretJohnson__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
__@BretJohnson__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

BretJohnson:
My GitHub account is MS linked, so I'm not sure why it continues to think I need to sign a CLA rather than tagging with cla-not-required.   I double checked GitHub account settings & tried relinking the account in the //opensourcehub portal, but that didn't seem to make a difference.   So I just went ahead & signed the CLA, even though I'm an MS employee.

AutoRest/Generators/NodeJS/NodeJS/TemplateModels/MethodTemplateModel.cs
```diff
@@ -102,6 +132,19 @@ public string MethodParameterDeclarationWithCallback
...
 
         /// <summary>
+        /// Generate the method parameter declarations with callback for a method, using TypeScript method syntax
+        /// </summary>
+        public string MethodParameterDeclarationWithCallbackTS {
+            get
+            {
+                var parameters = MethodParameterDeclarationTS;
+                var returnTypeTSString = ReturnType == null ? "void" : ReturnType.TSType(false);
+                parameters += ", callback: (err: Error, result: HttpOperationResponse<" + returnTypeTSString + ">) => void";
```
amarzavery: the result will never be of type HTTPOperationResponse any more. It will be dynamic, based on the return type defined for an operation.

BretJohnson:
I fixed up the generated code to conform to the new 4 parameter callbacks, per Amar's feedback above.

devigned:
@amarzavery or @yugangw-msft please give this a review.

BretJohnson:
I expect you'll end up tweaking something here, eventually, but it's probably most convenient to do that after pulling in the change (and generating the TypeScript output to see what's it's like).  I will have others review also, after this is merged.  Thanks.

amarzavery:
lgtm

amarzavery:
@BretJohnson - please look at gulp.js in the root directory.
Execute:
- npm install (from the root of your cloned repo)
- npm install gulp -g

Then run gulp -T (it will list all the tasks)
when you run:
gulp regenerate:expected:node   and  
gulp regenerate:expected:nodeazure

You will see the *.d.ts files for all the test projects

We should think about generating *.d.ts files for the test files and then run tests from the ypescript generated code. In this way we shall have coverage for typescript as well.

@devigned , @yugangw-msft - What do you guys think?

BretJohnson:
@amarzavery Much like you suggest, I still would like to convert the JS tests to be implemented in TypeScript instead, which should help test the generated d.ts files by compiling the tests with type checking.   I was planning to do that next, after this change is merged & you guys feel basically comfortable with it.   Changing the test code is a more invasive change though.

AutoRest/Generators/NodeJS/NodeJS/ClientModelExtensions.cs
```diff
@@ -242,6 +242,28 @@ private static string ValidatePrimaryType(this PrimaryType primary, IScopeProvid
...
+            if (primary == PrimaryType.Boolean)
+                return "boolean";
+            else if (primary == PrimaryType.Double || primary == PrimaryType.Int || primary == PrimaryType.Long)
+                return "number";
+            else if (primary == PrimaryType.String)
+                return "string";
+            else if (primary == PrimaryType.Date || primary == PrimaryType.DateTime)
+                return "Date";
+            else if (primary == PrimaryType.Object)
+                return "Object";   // TODO: test this
```
DanielRosenwasser: If `PrimaryType.Object` is supposed to work more like a dictionary, then `{ [key: string]: any }` might be a better type here.

AutoRest/Generators/NodeJS/NodeJS/Templates/MethodGroupIndexTemplateTS.cshtml
```diff
@@ -0,0 +1,19 @@
...
+@using System.Linq
+@inherits Microsoft.Rest.Generator.Template<Microsoft.Rest.Generator.NodeJS.ServiceClientTemplateModel>
+/*
+@Header(" * ")
+*/
+@EmptyLine
+/// <reference path="../typings/tsd.d.ts" />
+import { RequestOptions, WebResource } from "../node_modules/ms-rest/lib";
+@if (Model.ModelTypes.Any()) {
+@:import * as models from "../models";
```
DanielRosenwasser: Not 100% familiar with the output, but do you mean `./models`?

AutoRest/Generators/NodeJS/NodeJS/ClientModelExtensions.cs
DanielRosenwasser: If not, you probably mean `{}`

DanielRosenwasser:
Discussed with @weswigham, and he confirmed the following with me:

1. If you add a `"typings": "./lib/index.d.ts"` to your `package.json`, you'll be able to simply write `import {  RequestOptions, WebResource } from "ms-rest";`.
2. You shouldn't have the `/// <reference path="../typings/tsd.d.ts" />` in either your `index.d.ts` or your generated files. Rather:
    1. You should expect your consumers to include `node.d.ts` themselves at present.
	2. For your own uses, you can include a `tsconfig.json` which would point to your `tsd.d.ts`.

AutoRest/Generators/NodeJS/NodeJS/Templates/MethodGroupIndexTemplateTS.cshtml
BretJohnson: That code is output in the "operations" subdirectory, so "../models" is needed to point to the sibling "models" directory.   So it should be OK.

AutoRest/Generators/NodeJS/NodeJS/ClientModelExtensions.cs
BretJohnson: I chatted with @markcowl about this.   PrimaryType.Object is used when there's no schema type specified.   The question is whether primitive types (e.g. string and number) should be allowed here or just dictionaries.  I think it's a little safer to keep it as a TypeScript "Object", which is the more permissive type (and can't hurt), but after the code is merged I will test this more and chat with Amar about his intention for how the JavaScript code should behave in this scenario.

BretJohnson:
Re the typings & type references, I'll book a call with @weswigham to chat.   I still had some questions (and problems) I'd like to talk thru with him, on the type file reference config and getting good IntelliSense.   That conversation doesn't need to hold up the merge though--it would actually probably help if the pull request is merged first, making it easier for others (like Wesley) to try it out themselves & us to tweak things from there.

AutoRest/Generators/NodeJS/NodeJS/ClientModelExtensions.cs
DanielRosenwasser: `Object` usually isn't what users actually want. If users are unsure of what a value can be, they usually go with `any` which I'd suggest in this case.

BretJohnson: Just for my understanding, is there any difference in the semantics between Object and any as far as the *caller* is concerned?   Object is certainly more restrictive for the callee (which is moot in this case for a d.ts file as the callee isn't type checked), but does it matter to the caller?   I just wanted to make sure I understood the distinction.   Still, any is probably a better choice here for reasons of convention even if the semantics are exactly the same.   So I can change it for that reason.   And thanks for the review.

BretJohnson:
I fixed up the imports, per @weswigham's suggestions.

BretJohnson:
@amarzavery - Can you go ahead and merge this.   Or is there anything else we're waiting for?   I'd like to move on to get some other reviewers looking at it, once it's merged in & they can try it out themselves.  Thanks.

BretJohnson:
Oops - i missed that change.   I just fixed it as well.

weswigham:
Correspondingly, [`ClientRuntimes/NodeJS/ms-rest/package.json`](https://github.com/Azure/autorest/blob/master/ClientRuntimes/NodeJS/ms-rest/package.json#L12) should have a line inserted which reads `"typings": "./lib/index.d.ts",`

BretJohnson:
I fixed up package.json as well.

weswigham:
:+1:

amarzavery:
LGTM :+1:

PR #369, opened 2015-09-29
Add duration support to C#, Java, NodeJS

matthchr:
Also fixed a small issue in visual studio where duplicate files were included which made using visual studio to browse the generated code quite hard.
Ruby still treats duration as string.


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

amarzavery:
@matthchr - Can you please pull from upstream dev and resolve the merge conflicts?

matthchr:
Fixed -- I had to fix some issues which were pulled in which were causing gulp to fail when building the C# code (related to some oacr warnings), so those fixes are in this PR too now.

ClientRuntimes/CSharp/ClientRuntime/Serialization/Iso8601TimeSpanConverter.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <param name="value">The value to serialize.</param>
+        /// <param name="serializer">The JSON serializer.</param> 
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            if (serializer == null)
+            {
+                throw new ArgumentNullException("serializer");
+            }
+
+            //TODO: Will only be called with values of TimeSpan or TimeSpan? (Do we have to handle Nullable<TimeSpan>?)
```
yugangw-msft: @stankovski , could you provide comment on this question?

ClientRuntimes/CSharp/ClientRuntime/Serialization/Iso8601TimeSpanConverter.cs
```diff
@@ -0,0 +1,72 @@
+﻿namespace Microsoft.Rest.Serialization
```
yugangw-msft: license header

AutoRest/TestServer/server/routes/duration.js
```diff
@@ -0,0 +1,54 @@
+var express = require('express');
```
yugangw-msft: add license header

AutoRest/Generators/Ruby/Ruby/ClientModelExtensions.cs
```diff
@@ -86,6 +86,11 @@ private static string PrepareTypeForDocRecursively(IType type)
                 return "Array<Integer>";
             }
 
+            if (type == PrimaryType.TimeSpan)
+            {
+                return "Duration"; //TODO: Is this a real Ruby type...?
```
yugangw-msft: @devigned, please clarify

AutoRest/Generators/Java/Java/JavaCodeNamer.cs
```diff
@@ -348,7 +348,7 @@ public static String ImportedFrom(PrimaryType primaryType)
             else if (primaryType == PrimaryType.TimeSpan ||
                 primaryType.Name == "Period")
             {
-                return "java.time.Period";
+                return "org.joda.time.Period";
```
jianghaolu: Sorry I might have missed some context, but what's the reason you choose `Period` over `Duration` or `Interval`? (Thumbs up for using `org.joda.time` instead!)

AutoRest/Generators/NodeJS/NodeJS/Templates/ModelTemplateTS.cshtml
```diff
@@ -22,7 +22,7 @@ export interface @(Model.Name) {
     var propertyList = new List<Property>(Model.ComposedProperties);
     for (int i = 0; i < propertyList.Count; i++)
     {
-    @:@(Model.PropertyTS(propertyList[i], true));
+    @:@(ModelTemplateModel.PropertyTS(propertyList[i], true));
```
yugangw-msft: any reason you want to change this?

AutoRest/Generators/NodeJS/NodeJS/TemplateModels/ServiceClientTemplateModel.cs
```diff
@@ -105,5 +105,10 @@ public string RequiredConstructorParameters
                 return requiredParams.ToString();
             }
         }
+
+        public bool ContainsTimeSpan
+        {
+            get { return this.Methods.FirstOrDefault(m => m.Parameters.FirstOrDefault(p => p.Type == PrimaryType.TimeSpan) != null) != null; }
```
yugangw-msft: please use new line to maintain reasonable line width

AutoRest/Generators/Java/Java/JavaCodeNamer.cs
matthchr: An ISO8601 duration is described here: https://en.wikipedia.org/wiki/ISO_8601#Durations

Looking at: http://stackoverflow.com/questions/2653567/joda-time-whats-the-difference-between-period-interval-and-duration

> An interval in Joda-Time represents an interval of time from one millisecond instant to another instant. Both instants are fully specified instants in the datetime continuum, complete with time zone. Specific times are defined e.g. this might be the interval between 20:00:00GMT yesterday and 09:00:00GMT this morning.

So it's pretty clear that we don't want an interval, as a "Duration" does not have a start and an end, so interval doesn't apply.

> A duration in Joda-Time represents a duration of time measured in milliseconds. The duration is often obtained from an interval. i.e. we can subtract start from end of an interval to derive a duration

We could choose a joda Duration... 

> A period in Joda-Time represents a period of time defined in terms of fields, for example, 3 years 5 months 2 days and 7 hours. This differs from a duration in that it is inexact in terms of milliseconds. A period can only be resolved to an exact number of milliseconds by specifying the instant (including chronology and time zone) it is relative to. e.g. consider the period of 1 year, if we add this to January 1st we will always arrive at the next January 1st but the duration will depend on whether the intervening year is a leap year or not. Similarly if we add 1 month to the 1st of a month then we will arrive at the 1st of the next month but the duration (in milliseconds) will vary based on the month in question

Period matches most closely with the description of ISO8601 I think, (ISO8601 is Years/Months/Weeks/Days/Hours/Minutes/Seconds as well).  In Joda, a Duration only has these constructors:

> Duration(long duration)
> Duration(long startInstant, long endInstant) 

But a Period has this:

> Period(int years, int months, int weeks, int days, int hours, int minutes, int seconds, int millis)

That lead me to choose Period over Duration.  If you think we should use Duration instead I can change to that.

AutoRest/Generators/NodeJS/NodeJS/Templates/ModelTemplateTS.cshtml
matthchr: I didn't really want to change it -- this happened when I did a pull from "dev" branch and ran "gulp".  Gulp complained that PropertyTS could be a static method, so I had to make it a static method (and thus change this) in order to get gulp to proceed.

AutoRest/Generators/NodeJS/NodeJS/ClientModelExtensions.cs
```diff
@@ -606,6 +623,26 @@ private static string SerializePrimaryType(this PrimaryType primary, IScopeProvi
...
+                {
+                    builder.AppendLine("if(!{0} || !moment.isDuration({0})) {{", objectReference);
+                    builder = ConstructValidationCheck(builder, requiredTypeErrorMessage, objectReference, primary.Name);
+                    builder = ConstructBasePropertyCheck(builder, valueReference);
+                    return builder.AppendLine("{0} = {1}.toISOString();", valueReference, objectReference).ToString();
+                }
+
+                builder.AppendLine("if ({0}) {{", objectReference)
+                         .Indent()
+                         .AppendLine("if !(moment.isDuration({0})) {{",
```
yugangw-msft: Why you don't combine 2 conditions together?

AutoRest/Generators/Java/Java/JavaCodeNamer.cs
jianghaolu: Cool! Thanks for the detailed explanation. I also think `Period` is good at handling daylight saving time and less buggy. (from [this post](http://jnb.ociweb.com/jnb/jnbJul2008.html)) In worst cases, people who don't like `Period` can always call `toDuration`.

AutoRest/Generators/NodeJS/NodeJS.Tests/Expected/AcceptanceTests/BodyDuration/autoRestDurationTestService.js
```diff
@@ -0,0 +1,58 @@
...
+'use strict';
+
+var util = require('util');
+var msRest = require('ms-rest');
+var ServiceClient = msRest.ServiceClient;
+var WebResource = msRest.WebResource;
+
+var models = require('./models');
+var operations = require('./operations');
+var moment = require('moment');
```
yugangw-msft: i don't see we are using the "mement" at service level. I assume this should only be imported at operation/mode related code files

AutoRest/Generators/NodeJS/NodeJS.Tests/AcceptanceTests/acceptanceTests.js
```diff
@@ -599,6 +601,44 @@ describe('nodejs', function () {
...
+        testClient.duration.getNull(function (error, result) {
+          should.not.exist(result);
+          should.not.exist(error);
+          done();
+        });
+      });
+
+      it('should properly handle invalid value for Duration', function (done) {
+        testClient.duration.getInvalid(function (error, result) {
+          //TODO: There should be an error here, except for some reason moment.js allows non-ISO strings and will just construct a duration of length 0
```
yugangw-msft: can you do more debugging to clear out this error?

AutoRest/Generators/NodeJS/NodeJS/ClientModelExtensions.cs
matthchr: I followed the pattern of the other primary types (DateTime, Date, ByteArray, etc) which follow this same pattern.

AutoRest/Generators/NodeJS/NodeJS.Tests/AcceptanceTests/acceptanceTests.js
matthchr: I believe this is just the way the Moment library works (i.e. the behavior observed is expected).  The comment was more aimed at pointing out my original assumption was wrong... I think the comment I added is unclear, I'll remove it.

AutoRest/Generators/NodeJS/NodeJS.Tests/Expected/AcceptanceTests/BodyDuration/autoRestDurationTestService.js
matthchr: @yugangw-msft Good catch -- will fix this.

AutoRest/TestServer/server/routes/duration.js
matthchr: Did this

amarzavery:
@matthchr - The PR from Bret has been merged. You should be able to execute gulp regenerate:expected:node after pulling from upstream dev

matthchr:
@amarzavery Actually, these changes (seen above) are the result of me rebasing on top of the latest dev (with Bret's changes).  His fixes solved all of my issues.

stankovski:
@azuresdkci add to whitelist

PR #371, opened 2015-09-30
Add support for RequestId header name overloads

matthchr:
Added 2 extensions.  
1 is "x-ms-client-request-id" on parameter
1 is "x-ms-request-id" on operation

The requestId on operation one should really probably be on response headers, but AutoRest doesn't really understand/do anything with response headers yet so I couldn't put it there.


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

matthchr:
I'm looking at why the travis ci is failing... will update.

AutoRest/Generators/Azure.Common/Azure.Common/AzureCodeGenerator.cs
```diff
@@ -454,6 +459,65 @@ public static void AddPageableMethod(ServiceClient serviceClient)
...
         }
 
+        public static string GetClientRequestIdString(Method method)
+        {
+            if (method == null)
+            {
+                throw new ArgumentNullException("method");
+            }
+            
+            string clientRequestIdName = null;
```
yugangw-msft: can you set the default value of ""x-ms-client-request-id"" here? This way you don't need the "if" block later

AutoRest/Generators/Azure.Common/Azure.Common/AzureCodeGenerator.cs
```diff
@@ -454,6 +459,65 @@ public static void AddPageableMethod(ServiceClient serviceClient)
...
+        }
+
+        public static string GetRequestIdString(Method method)
+        {
+            if (method == null)
+            {
+                throw new ArgumentNullException("method");
+            }
+
+            string requestIdName = null;
```
yugangw-msft: same comment here.

AutoRest/TestServer/swagger/azure-special-properties.json
```diff
@@ -732,6 +732,43 @@
...
+        "operationId": "header_customNamedRequestId",
+        "description": "Send foo-client-request-id = 9C4D50EE-2D56-4CD3-8152-34347DC9F2B0 in the header of the request",
+        "parameters": [
+          {
+            "name": "foo-client-request-id",
+            "in": "header",
+            "required": true,
+            "type": "string",
+            "description": "The fooRequestId",
+            "x-ms-client-request-id": { "value": "true" },
```
yugangw-msft: forgive me if i am out of context, but any reason we want to use a json objectl rather just a bool value
     "x-ms-client-request-id": true

yugangw-msft:
the rest looks good to me.

AutoRest/TestServer/swagger/azure-special-properties.json
matthchr: David Justice specifically requested that I use an object instead of a boolean for possible future extensibility purposes

matthchr:
Updated in response to the comments

PR #381, opened 2015-10-08
Add Parameter Grouping to AutoRest for Azure.CSharp, Azure.NodeJS

matthchr:
A "Parameter group" is just a nicer way to bundle parameters so that methods don't get horribly long.  It is controlled by an extension to the Swagger specification.

Left out Ruby/Java for now, as Java doesn't have an "Azure.Java" yet and cut Ruby in the name of time (if/when somebody wants the feature in Ruby they should be able to follow the implementation from NodeJS/C#/Java since the metadata is in the client model now)


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

amarzavery:
@matthchr - It is a nice concept. As per the current implementation, the user can group parameters in 1 bucket. What if the user wants to group them into more than 1 bucket? Would it be a better idea if the extension takes a key value pair instead of a Boolean, something like this:
```
"x-ms-parameter-grouping" : {"groupName" : "group1" }
```
In this way the user can put parameters into different buckets if required. It also provides a flexibility to name the parameter group as per their requirement.

What do you think?

matthchr:
@amarzavery Yes, allowing more than one parameter group was something I thought about (if you see in the implementation the "groups" is actually a collection and everywhere when I generate the code I treat it as such), so the only thing currently stopping us from having multiple groups is the extension piece...

With that said are some additional complications with allowing multiple parameter groups which is why I leaned against it initially:
1) You must allow swagger to specify "type" names then (which only apply to some languages?)
2) You must allow swagger to specify "parameter" names (instead of just hardcoding the strong "parameter") - maybe this could be sourced off of the same string as the "type" names issue mentioned above
3) You have to handle type-collisions better then, which I wasn't really sure how to do. Since right now the type of the parameter group class is based on the OperationId (which I believe is unique) we're protected from that naturally.

I agree it would be good to have, but maybe leave it for a V2 implementation?

brjohnstmsft:
Have you considered a more general approach? Rather than indirectly define a model type by grouping parameters, it would be really handy if we could bind some parameter values to others. For example, for PUT operations, it would be really nice to be able to bind the URL parameter for the resource ID to the corresponding property on the resource itself. See this issue for a concrete example: https://github.com/Azure/autorest/issues/382

matthchr:
@brjohnstmsft David Justice had discussed something similar to what you're asking for - he colloquially called it "parameter factories" or something.  My team has a similar request, but instead of a "duplicate" parameter like in your example (where there is Model.Id and id), we want to set certain parameters to be automatically generated instead of including them in the signature.

For example: parameter "string requestId" (required: true) could instead be optional or omitted entirely and automatically generated as a Guid (or in your case automatically generated from an existing model field).

I do see the two things being different features though: this is about how to structure the parameters we **do** want, what you're asking for is a way to omit parameters which are "required" by the REST API but which you **do not** want.

yugangw-msft:
reset this please

yugangw-msft:
add to whitelist

AutoRest/Generators/Azure.Common/Azure.Common/AzureCodeGenerator.cs
```diff
@@ -204,6 +207,84 @@ public static void AddLongRunningOperations(ServiceClient serviceClient)
...
+        public static void AddParameterGroups(ServiceClient serviceClient)
+        {
+            if (serviceClient == null)
+            {
+                throw new ArgumentNullException("serviceClient");
+            }
+
+            foreach (Method method in serviceClient.Methods)
+            {
+                //TODO: Does this group name need to be normalized later, or does that already happen?
```
yugangw-msft: it will be normalized by each code generator, so you are good

AutoRest/Generators/Azure.Common/Azure.Common/AzureCodeGenerator.cs
```diff
@@ -204,6 +207,84 @@ public static void AddLongRunningOperations(ServiceClient serviceClient)
...
+
+                            if (!properties.ContainsKey(parameterGroupName))
+                            {
+                                properties.Add(parameterGroupName, new List<Property>());
+                            }
+
+                            properties[parameterGroupName].Add(new Property()
+                                {
+                                    IsReadOnly = false, //Since these properties are used as parameters they are never read only
+                                    Name = CodeNamer.PascalCase(parameter.Name),
```
yugangw-msft: any reason, you like to do the pascal conversion here? If you don't do it would c# generator produces property name with wrong casing?

AutoRest/Generators/Azure.Common/Azure.Common/AzureCodeGenerator.cs
```diff
@@ -204,6 +207,84 @@ public static void AddLongRunningOperations(ServiceClient serviceClient)
...
+                                    IsRequired = parameter.IsRequired,
+                                    DefaultValue = parameter.DefaultValue,
+                                    //Constraints = parameter.Constraints, TODO do these need to be copied?
```
yugangw-msft: let us do not copy it, unless you like to produce error validation on set.

AutoRest/Generators/CSharp/Azure.CSharp.Tests/AcceptanceTests.cs
```diff
@@ -685,5 +686,64 @@ public void CustomNamedRequestIdTest()
...
             }
         }
+        
+        [Fact]
+        public void ParameterGroupingTests()
+        {
+            const int body = 1234;
+            const string header = "header";
+            const int query = 21;
+            const string path = "path";
```
yugangw-msft: For readibility, suggest rename the variable such as "pathParameter", "bodyParameter", etc

AutoRest/Generators/CSharp/Azure.CSharp/AzureCSharpCodeGenerator.cs
```diff
@@ -139,6 +142,20 @@ public override void NormalizeClientModel(ServiceClient serviceClient)
...
+            foreach (Method method in serviceClient.Methods)
+            {
+                foreach (string parameterGroup in method.ParameterGroups)
+                {
+                    Parameter groupParameter = method.Parameters.First(p => string.Equals(p.Type.Name, parameterGroup, StringComparison.InvariantCultureIgnoreCase));
+                    var parameterGroupTemplate = new ParameterGroupTemplate
+                    {
+                        Model = new ParameterGroupTemplateModel(serviceClient, groupParameter)
+                    };
+                    await Write(parameterGroupTemplate, Path.Combine("Parameters", parameterGroupTemplate.Model.ParameterGroupType + ".cs"));
```
yugangw-msft: Does that mean we will create the group parameter types under \\Model\\Parameters? It is just fine to produce under \\Models? Those are all model classes anyway.

AutoRest/Generators/CSharp/CSharp/Templates/MethodTemplate.cshtml
```diff
@@ -116,7 +126,7 @@ public async Task<@(Model.OperationResponseReturnTypeString)> @(Model.Name)WithH
     {
         <text>
     // Serialize Request
-    string requestContent = JsonConvert.SerializeObject(@(Model.RequestBody.Name), @(Model.GetSerializationSettingsReference(Model.RequestBody.Type)));
+    string requestContent = JsonConvert.SerializeObject(@(Model.RequestBody.ParameterAccessor), @(Model.GetSerializationSettingsReference(Model.RequestBody.Type)));
```
yugangw-msft: Is it necessary for "RequestBody"?

AutoRest/Generators/CSharp/Azure.CSharp/TemplateModels/AzureParameterTemplateModel.cs
```diff
@@ -31,5 +32,20 @@ public override string DeclarationExpression
...
+        {
+            get
+            {
+                if (string.IsNullOrEmpty(this.ParameterGroup))
+                {
+                    return base.ParameterAccessor;
+                }
+                else
+                {
+                    return CodeNamer.CamelCase(this.ParameterGroup) + CodeNamer.PascalCase(base.ParameterAccessor);
```
yugangw-msft: I might misunderstand it, but I would assume we should have a "." between them to make it a property getter expression.

AutoRest/Generators/CSharp/CSharp/Templates/MethodTemplate.cshtml
```diff
@@ -52,9 +62,9 @@ public async Task<@(Model.OperationResponseReturnTypeString)> @(Model.Name)WithH
     {
         invocationId = ServiceClientTracing.NextInvocationId.ToString();
         Dictionary<string, object> tracingParameters = new Dictionary<string, object>();
-    @foreach (var parameter in Model.LocalParameters)
+    @foreach (var parameter in Model.ParameterTemplateModels.Where(p => !p.IsParameterGroup && p.ClientProperty == null))
```
yugangw-msft: you have already updated the "LocalParameters" getter, so filter out one with non empty "ParameterGroup", to me the check of"IsParameterGroup" here is redundant.

AutoRest/Generators/CSharp/CSharp/Templates/MethodGroupTemplate.cshtml
```diff
@@ -8,6 +8,7 @@ namespace @Settings.Namespace
     using System;
     using System.Linq;
     using System.Collections.Generic;
+    using System.Globalization;
```
yugangw-msft: It is unclear why you need to add this import?

AutoRest/Generators/CSharp/CSharp/CSharpCodeGenerator.cs
```diff
@@ -6,6 +6,8 @@
 using System.Threading.Tasks;
 using Microsoft.Rest.Generator.ClientModel;
 using Microsoft.Rest.Generator.CSharp.Templates;
+using System.Collections.Generic;
```
yugangw-msft: again, unclear why we need to add these 2 usings

AutoRest/Generators/CSharp/CSharp/TemplateModels/ParameterGroupTemplateModel.cs
```diff
@@ -0,0 +1,64 @@
...
+            
+            CompositeType compositeType = this.groupParameter.Type as CompositeType;
+
+            this.GroupedParameters = compositeType.Properties;
+            this.Documentation = compositeType.Documentation;
+        }
+
+        public string ParameterGroupName
+        {
+            get { return this.groupParameter.Type.Name; }
```
yugangw-msft: we do mean to use the type name, right?

AutoRest/Generators/CSharp/CSharp/Templates/ParameterGroupTemplate.cshtml
```diff
@@ -0,0 +1,56 @@
...
+    using System.Linq;
+    using System.Collections.Generic;
+    using System.Net;
+    using System.Net.Http;
+    using System.Net.Http.Headers;
+    using System.Text;
+    using System.Text.RegularExpressions;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Rest;
```
yugangw-msft: have you checked whether all those imports are really needed

AutoRest/Generators/NodeJS/Azure.NodeJS/TemplateModels/AzureMethodTemplateModel.cs
```diff
@@ -12,8 +13,18 @@ public class AzureMethodTemplateModel : MethodTemplateModel
         public AzureMethodTemplateModel(Method source, ServiceClient serviceClient)
             : base(source, serviceClient)
         {
+            if (source == null)
+            {
+                throw new ArgumentNullException("source");
+            }
+
```
yugangw-msft: remove one extra empty line

AutoRest/Generators/Ruby/Ruby/TemplateModels/MethodTemplateModel.cs
```diff
@@ -137,9 +137,10 @@ public IEnumerable<ParameterTemplateModel> LocalParameters
         {
             get
             {
+                //Omit parameter group parameters for now since AutoRest-Ruby doesn't support them
                 return
```
yugangw-msft: @devigned, FYI

AutoRest/TestServer/server/routes/azureParameterGrouping.js
```diff
@@ -0,0 +1,50 @@
+var express = require('express');
```
yugangw-msft: license header?

AutoRest/Generators/CSharp/CSharp/Templates/MethodTemplate.cshtml
```diff
@@ -30,17 +31,26 @@
 /// </param>
 public async Task<@(Model.OperationResponseReturnTypeString)> @(Model.Name)WithHttpMessagesAsync(@(Model.GetAsyncMethodParameterDeclaration(true)))
 {
+    @foreach (var groupParameter in Model.ParameterTemplateModels.Where(p => p.IsParameterGroup))
```
yugangw-msft: make sure you adjust the generated comments for group parameters as well, do the same for node.js and java as well.

AutoRest/Generators/Azure.Common/Azure.Common/AzureCodeGenerator.cs
matthchr: @yugangw-msft Updated comment to reflect that fact

AutoRest/Generators/Azure.Common/Azure.Common/AzureCodeGenerator.cs
matthchr: @yugangw-msft Updated this comment then

AutoRest/Generators/CSharp/Azure.CSharp/AzureCSharpCodeGenerator.cs
matthchr: @yugangw-msft Right now the folder structure is like so:
>Root\
>>Models\
>>Parameters\

I can move the Parameters classes into the model folder if you think that is better/more clear.

AutoRest/Generators/Azure.Common/Azure.Common/AzureCodeGenerator.cs
matthchr: @yugangw-msft This is actually not required at all (as you say, normalization is performed later in the process), so I removed this line and just use parameter.Name

AutoRest/Generators/CSharp/Azure.CSharp.Tests/AcceptanceTests.cs
matthchr: @yugangw-msft Done

AutoRest/Generators/CSharp/CSharp/Templates/MethodTemplate.cshtml
matthchr: @yugangw-msft Yes, because we allow you to group requestBody into the parameterGroup as well, which means that instead of referring to it by "body" it'll be "parameters.Body" (or something like that).  Take a look at the generated code `ParameterGroupingOperations.cs` and `ParameterGroupingPostRequiredParameters.cs` to see what I mean.

AutoRest/Generators/CSharp/Azure.CSharp/TemplateModels/AzureParameterTemplateModel.cs
matthchr: At the top of the method, we "flatten" the group object to avoid having to repeat null checks, etc.  So this signature:
```C#
public async Task<AzureOperationResponse> PostRequiredWithHttpMessagesAsync(ParameterGroupingPostRequiredParameters parameterGroupingPostRequiredParameters, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken))
```
has this at the top to extract the parameters:
```C#
var parameterGroupingPostRequiredParametersBody = (parameterGroupingPostRequiredParameters == null ? default(int?) : parameterGroupingPostRequiredParameters.Body);
var parameterGroupingPostRequiredParametersHeader = (parameterGroupingPostRequiredParameters == null ? default(string) : parameterGroupingPostRequiredParameters.Header);
var parameterGroupingPostRequiredParametersQuery = (parameterGroupingPostRequiredParameters == null ? default(int?) : parameterGroupingPostRequiredParameters.Query);
var parameterGroupingPostRequiredParametersPath = (parameterGroupingPostRequiredParameters == null ? default(string) : parameterGroupingPostRequiredParameters.Path);
```

These parameters are then used later like so:
```C#
tracingParameters.Add("body", parameterGroupingPostRequiredParametersBody);
tracingParameters.Add("header", parameterGroupingPostRequiredParametersHeader);
tracingParameters.Add("query", parameterGroupingPostRequiredParametersQuery);
tracingParameters.Add("path", parameterGroupingPostRequiredParametersPath);
```

This is mostly just to avoid having to repeat null checks on the "group" object over and over (since groups can be required or not required, so sometimes a null-group is allowed)

AutoRest/Generators/CSharp/CSharp/Templates/MethodTemplate.cshtml
matthchr: @yugangw-msft I updated the getter to return all parameters *without* a parameter group.  That set would *include* the ParameterGroup parameter itself.
```C#
ParameterTemplateModels.Where(
                        p => p != null && p.ClientProperty == null && string.IsNullOrEmpty(p.ParameterGroup) && !string.IsNullOrWhiteSpace(p.Name))
                        .OrderBy(item => !item.IsRequired);
```

Here, I am getting all the parameters (even if they were grouped) and excluding just the "fake" parameters (the ParameterGroup parameter itself)
```C#
Model.ParameterTemplateModels.Where(p => !p.IsParameterGroup && p.ClientProperty == null)
```

So it's really two different cases -- this change isn't redundant.

AutoRest/Generators/CSharp/CSharp/Templates/MethodGroupTemplate.cshtml
matthchr: @yugangw-msft You're right -- this isn't needed.  Removed it.

AutoRest/Generators/CSharp/CSharp/CSharpCodeGenerator.cs
matthchr: @yugangw-msft From a previous iteration of the code -- I missed removing them.  Removed them now.

AutoRest/Generators/CSharp/CSharp/TemplateModels/ParameterGroupTemplateModel.cs
matthchr: @yugangw-msft This field is actually never used anywhere, since the "name" of the parameterGroup is the name of the parameter on the method.  I've removed this property and just left the other property `ParameterGroupType`

AutoRest/Generators/CSharp/CSharp/Templates/ParameterGroupTemplate.cshtml
matthchr: @yugangw-msft Good catch.  I had copied this from elsewhere -- they in fact aren't all required.  I have removed all of them except for `using System`.

AutoRest/Generators/NodeJS/Azure.NodeJS/TemplateModels/AzureMethodTemplateModel.cs
matthchr: @yugangw-msft Done

AutoRest/TestServer/server/routes/azureParameterGrouping.js
matthchr: @yugangw-msft None of these testserver .js files seem to have the license header

AutoRest/Generators/CSharp/CSharp/Templates/MethodTemplate.cshtml
matthchr: @yugangw-msft Yes, that's already done in node and C# -- for Java, ParameterGrouping isn't fully supported yet since there isn't an Azure.Java yet, but for C# and node I've included examples:
```C#
/// <summary>
/// Post a bunch of required parameters grouped
/// </summary>
/// <param name='operations'>
/// The operations group for this extension method.
/// </param>
/// <param name='parameterGroupingPostRequiredParameters'>
/// Additional parameters for the operation
/// </param>
public static void PostRequired(this IParameterGroupingOperations operations, ParameterGroupingPostRequiredParameters parameterGroupingPostRequiredParameters)
```
I am not a node comment expert so it'd be good if you can confirm that this comment looks correct.

```node
/**
 * Post a bunch of required parameters grouped
 *
 * @param {object} parameterGroupingPostRequiredParameters Additional
 * parameters for the operation
 * 
 * @param {number} [parameterGroupingPostRequiredParameters.body]
 * 
 * @param {string} [parameterGroupingPostRequiredParameters.header]
 * 
 * @param {number} [parameterGroupingPostRequiredParameters.query] Query
 * parameter with default
 * 
 * @param {string} [parameterGroupingPostRequiredParameters.path] Path
 * parameter
 * 
 * @param {object} [options]
 *
 * @param {object} [options.customHeaders] headers that will be added to
 * request
 *
 * @param {function} callback
 *
 * @returns {function} callback(err, result, request, response)
 *
 *                      {Error}  err        - The Error object if an error occurred, null otherwise.
 *
 *                      {null} [result]   - The deserialized result object.
 *
 *                      {object} [request]  - The HTTP Request object if an error did not occur.
 *
 *                      {stream} [response] - The HTTP Response stream if an error did not occur.
 */
ParameterGrouping.prototype.postRequired = function (parameterGroupingPostRequiredParameters, options, callback) {
```

AutoRest/Generators/CSharp/Azure.CSharp/AzureCSharpCodeGenerator.cs
yugangw-msft: Please move the parameters classes into the model folder, I prefer flat better than nested for files with similar roles

AutoRest/Generators/CSharp/CSharp/Templates/MethodTemplate.cshtml
yugangw-msft: yep, looks good to me

matthchr:
Okay, I moved the Parameters objects into the models folder and namespace.

matthchr:
@yugangw-msft I am not sure why the "default" check failed in jenkins, can somebody help me take a look at that?  Travis pass seems okay...

yugangw-msft:
@matthchr please merge with dev. The CI failure was caused by the java change made to the dev branch this afternoon

AutoRest/Generators/CSharp/CSharp/Templates/MethodTemplate.cshtml
yugangw-msft: @devigned @stankovski 
Please do a quick review. With this change a client model method's parameters list gets a little complicated that it no long simply maps to the argument list to emit out for the method declarations, so more condition checks were introduced to filter based on the context. 
This works, but if you have better suggestion, please comment here before it gets merged.
For "ungrouping", we probably won't do it because other than storage team's ask, there is less benefits; rather the author can tweak the spec easily.

matthchr:
@yugangw-msft Merge with dev done

matthchr:
@devigned @stankovski Should we use these comments to discuss proposed changes to this implementation?

One thing that @xingwu1 pointed out to me is that this PR is pretty critical to our adoption, so I don't really want to have to re-design/re-do everything (especially if it's significantly more complicated to do an alternate way) because it will delay our swagger adoption (after this PR, we can actually generate a real client).

If we can align on:
1. The Swagger representation (tentatively, I think what I have now is fine).
2. The generated code structure (I think what I have now is fine).
3. A model structure which is generic *enough* for now.

If we change the internals of AutoRest (the model structure) in the future to be more generic or cater to more scenarios, that change is fine because it's hidden from clients generated code and their swagger specifications.

**I'll start with my suggestion:**
1. Swagger extension format stays as it is in this PR (a piece of metadata on the `Parameter` object in the swagger spec).
2. Client model is updated to have some sort of "mapped" parameter notion.  Basically, there can be parameters in the `ParameterList`, and there will be a dictionary mapping `ParameterList` parameters to "real" parameters, and there will be some generic methods which will transform the parameters in the list to parameters in the internals of the method.  At the end of the day the generated method code will still look almost exactly like I have it in the PR today (top of the method will extract stuff out of the object into a set of named local variables, other pieces of the method will consume the local variables).

I think that the part which *might* be reusable is the stuff for the client model, but even that is maybe not totally code-reuse but rather just concept re-use... what I mean is this:

- For input parameters, there is this notion of a ParameterList, and a dictionary or something which describes the mapping of Parameters -> UnwrappedParameters.
- For output parameters, there is the *reverse* notion - you have a set of parameters and a mapping collapsing them into a new type, so it's UnwrappedParameters -> OutputParameters.

Let me know what you think.

AutoRest/AutoRest.Core/ClientModel/Method.cs
```diff
@@ -59,6 +59,27 @@ public Method()
         public List<Parameter> Parameters { get; private set; }
 
         /// <summary>
+        /// Gets the parameter groups.
+        /// </summary>
+        public IEnumerable<string> ParameterGroups
```
stankovski: This will be replaced with two "Mapping" properties - one for input and one for output.

AutoRest/Generators/CSharp/Azure.CSharp.Tests/Expected/AcceptanceTests/AzureParameterGrouping/ParameterGroupingOperations.cs
```diff
@@ -0,0 +1,463 @@
...
+        /// Additional parameters for the operation/// </param>
+        /// <param name='customHeaders'>
+        /// Headers that will be added to request.
+        /// </param>
+        /// <param name='cancellationToken'>
+        /// The cancellation token.
+        /// </param>
+        public async Task<AzureOperationResponse> PostRequiredWithHttpMessagesAsync(ParameterGroupingPostRequiredParameters parameterGroupingPostRequiredParameters, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var parameterGroupingPostRequiredParametersBody = (parameterGroupingPostRequiredParameters == null ? default(int?) : parameterGroupingPostRequiredParameters.Body);
```
stankovski: The name of the variable should probably be the same as in swagger (unless it's not unique)

stankovski:
@matthchr I agree with the approach. Let's leave swagger extension as-is. I also think that the generated code will stay almost the same. An output block of similar structure will be needed and you may need to convert terniary expression to an `if` block for mapping from primitives to a class.

As far as mapping, I would leave current ParameterList as is and add a new data structure (either just a dictionanry or a class to work with output mapping) that matches an existing parameter or parameter property to a new parameter or parameter property.

matthchr:
@stankovski I've created an issue #395 to discuss the structure of the returned object for the return header -> object path.  Can we delay the implementation of that until we've decided on what the output object should look like (is it a union of the returned headers and the body, a <T, K> double-generic (in C#) or something else?).

Until then, I think I won't be able to write any test cases or samples for the "re-grouping" case (since there will be no place to put the returned header seeing as the body is already populated by the JSON deserialization), so I will hold off on writing it.  When we get there, we can follow the exact same pattern as the un-grouping case in this PR.

Does that sound okay?

matthchr:
@stankovski This has the changes to ParameterGrouping to move to the "Mapping" idea -- the CI is failing due to an issue in dev branch -- once that is fixed I'll merge and update.

AutoRest/AutoRest.Core/ClientModel/Method.cs
```diff
@@ -59,6 +60,22 @@ public Method()
         public List<Parameter> Parameters { get; private set; }
 
         /// <summary>
+        /// Gets the list of Parameter Expansions
+        /// </summary>
+        public Dictionary<string, Dictionary<Property, Parameter>> ParameterExpansions { get; private set; }
```
stankovski: To make it more generic and support parameter groups as well primitive to object mapping can we do the following:
1. Rename this to ParameterMappings
2. Add new type to ClientModel called `ParameterMapping` with the following properties
   - `Parameter InputParameter {get; set;}`
   - `string InputParameterProperty {get; set;}` (either null, which will map the parameter itself, or dot separated string to match properties)
   - `Parameter OutputParameter {get; set;}`
   - `string OutputParameterProperty {get; set;}` (either null, which will map the parameter itself, or dot separated string to match properties)

stankovski: We should also add a `public Parameter Body {get;}` to get easier access to the body parameter

PR #383, opened 2015-10-09
 Page class should be auto-generated instead of hardcoded in client runtime

xingwu1:
In current client runtime, the Page class can only de-serialize "nextLink" and "@odata.nextLink" two kind of nextLink string. However, it should be dynamic. So change the "x-ms-pageable" extension to:

     "x-ms-pageable": { "nextLinkName": "nextLink", "itemName" : "value" }

So both next link and items JSON name can be dynamic.


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

yugangw-msft:
reset this please

yugangw-msft:
add to whitelist

AutoRest/Generators/CSharp/Azure.CSharp/AzureCSharpCodeNamer.cs
```diff
@@ -24,15 +24,31 @@ public class AzureCSharpCodeNamer : CSharpCodeNamer
...
                 return null;
             }
 
-            return (string)ext["path"];
+            nextLinkName = (string)ext["nextLinkName"] ?? "nextLink";
+            string itemName = (string)ext["itemName"] ?? "value";
+            var keypair = new KeyValuePair<string, string>(nextLinkName, itemName);
+            if (!serviceClient.PageClasses.ContainsKey(keypair))
+            {
+                if (serviceClient.PageClasses.Count > 0)
```
yugangw-msft: I think we will only have a one page class per each service client, right? if yes, then we should throw when we have more than one.

yugangw-msft:
The rest looks good to me

AutoRest/Generators/CSharp/Azure.CSharp/AzureCSharpCodeNamer.cs
xingwu1: Actually, the code will generate multiple page classes (named: Page, Page1, Page2, ...) if there are multiple x-page extensions with different "nextLinkName" or "itemName".

xingwu1: Do you think I shouldn't go that far? I can change to only have a single page class.

yugangw-msft: We can keep multiple page class, but are you happy with the default naming of "Page1", "Page2", etc? Rather better name by either getting from new filed from the extension in swagger, or derived from the 'itemname'

AutoRest/Generators/CSharp/Azure.CSharp.Tests/PageJsonTest.cs
```diff
@@ -73,39 +72,11 @@ public void TestNextLinkDeSerialization()
                 ContractResolver = new ReadOnlyJsonContractResolver()
             };
             deserializeSettings.Converters.Add(new ResourceJsonConverter());
-            deserializeSettings.Converters.Add(new PolymorphicDeserializeJsonConverter<Product>("dType"));
-            var deserializedProduct = JsonConvert.DeserializeObject<Page<Product>>(responseBody, deserializeSettings);
+            deserializeSettings.Converters.Add(new PolymorphicDeserializeJsonConverter<TestProduct>("dType"));
+            var deserializedProduct = JsonConvert.DeserializeObject<Fixtures.Azure.AcceptanceTestsPaging.Models.Page<TestProduct>>(responseBody, deserializeSettings);
```
devigned: Could you change this to be IPage rather than the concrete implementation of the generated class. I'm thinking that IPage should live in the client runtime and also implement IEnumerable. IPage would consist of the enumerator and the NextLink, but not be the concrete implementation. Leaving it to the generator to implement the interface.

Thoughts?

devigned: @xingwu1 just one item I'm hoping we can change a bit.

AutoRest/AutoRest.Core/ClientModel/ServiceClient.cs
```diff
@@ -93,6 +94,11 @@ public IEnumerable<string> MethodGroups
         public Dictionary<string, object> Extensions { get; private set; }
 
         /// <summary>
+        /// Gets page extensions dictionary.
+        /// </summary>
+        public Dictionary<KeyValuePair<string, string>, string> PageClasses { get; private set; }
```
stankovski: We should either not have this property on ServiceClient or make x-ms-pageable not Azure specific.

stankovski:
Please also increment a ClientRuntime version since you've removed a Page class.

xingwu1:
Update code in following change:
1. Create IPage interface in ClientRuntime.
2. Generated Page class inherits from IPage.
3. Generated function will return IPage instead of concrete Page class.
4. There is extra field in extension call "className", to let user specify the Page class name.

stankovski:
@xingwu1 please pull from upstream to resolve merge conflict

PR #387, opened 2015-10-13
Fix bug: Uri drop the base uri last segment if it is not trail with slash

xingwu1:
If the baseUri (with some path) is not end with "/", the Uri constructor will drop the last segment from baseUri. The fix is automatic append "/" if the baseUri is not end with slash.

### DIFF
--- a/AutoRest/Generators/CSharp/CSharp/Templates/MethodTemplate.cshtml
+++ b/AutoRest/Generators/CSharp/CSharp/Templates/MethodTemplate.cshtml
@@ -67,7 +67,8 @@ public async Task<@(Model.OperationResponseReturnTypeString)> @(Model.Name)WithH
     }
     else
     {
-    @:var url = new Uri(@(Model.ClientReference).BaseUri, "@(Model.Url.Trim('/'))").ToString();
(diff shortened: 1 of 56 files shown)


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

yugangw-msft:
reset this please

AutoRest/Generators/CSharp/CSharp/Templates/MethodTemplate.cshtml
```diff
@@ -67,7 +67,8 @@ public async Task<@(Model.OperationResponseReturnTypeString)> @(Model.Name)WithH
     }
     else
     {
-    @:var url = new Uri(@(Model.ClientReference).BaseUri, "@(Model.Url.Trim('/'))").ToString();
+    @:var baseUrl = @(Model.ClientReference).BaseUri.AbsoluteUri;
+    @:var url = new Uri(new Uri(baseUrl + (baseUrl.EndsWith("/") ? "" : "/")), "@(Model.Url.Trim('/'))").ToString();
```
yugangw-msft: It looks little odd, but I guess it is the best we can do here :)

yugangw-msft:
add to whitelist

PR #388, opened 2015-10-14
fix a flakiness in autorest nuget package testing

yugangw-msft:
This flakiness has surfaced more frequently, so we need to do something to mitigate 
The CI box could cache an older version of client run time package, say if it ran jobs for Maml or Posh. This causes a build break on the package test which is supposed to use the local package just built out under the binaries folder. So add a step to clean the cached one.


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

PR #393, opened 2015-10-15
[#105719632] Added x-ms-enum object to use for string constants class generation.

hovsepm:


azurecla:
Hi __@hovsepm__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (hovsepm). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

AutoRest/Modelers/Swagger.Tests/Swagger/swagger-data-types.json
```diff
@@ -368,7 +368,7 @@
           "type": "string"
         },
         "color": {
-          "x-ms-enum": "Colors",
+          "x-ms-enum": { "name": "Colors", "expandable": false },
```
brjohnstmsft: The name "expandable" isn't really self-explanatory. Maybe something more like "modelAsString"?

AutoRest/Modelers/Swagger/ObjectBuilder.cs
```diff
@@ -45,14 +46,24 @@ public virtual IType BuildServiceType(string serviceTypeName)
...
                 {
-                    enumType.IsExpandable = false;
-                    enumType.Name = SwaggerObject.Extensions["x-ms-enum"] as string;
+                    var enumObject = SwaggerObject.Extensions[CodeGenerator.EnumObject] as Newtonsoft.Json.Linq.JContainer;
+                    if (enumObject != null)
+                    {
+                        enumType.Name= enumObject["name"].ToString();
+                        if (enumObject["expandable"] != null)
+                        {
+                            enumType.IsExpandable = bool.Parse(enumObject["expandable"].ToString());
```
brjohnstmsft: If I'm reading this right, `IsExpandable` will be `false` by default. This is a breaking change for everyone currently using `x-ms-enum` that expects it to generate enums rather than string constants. The default should be the current behavior.

hovsepm: the default is `false`. in that case the type will be generated as Enum. If it is set to true then it will be generated as string constants. I'm not sure where do we brake existing behavior.

brjohnstmsft: My mistake. I misread BodyComplex as BodyString in the expected code above.

aws-na-ss-usplu:
Novice: Our very small two cents **_;-)_**

We currently develop solutions in two worlds… LoopBack/Node.js & WebApi/Asp.net

 We would love to have a solution that complements each others interpretation of the concepts around

- Generic Extensions for Enum
- Implement of Enum
- Solution path to transition from Swagger Extension to first class member of Swagger 2.x schema

We currently use swagger to generate both client/server code for LoopBack & .NET, and client code in Java and PHP.  When it comes to **Enum** and **Additional Properties** we have to deal with exceptions because like HTML specs each community is interpreting the specifications a little differently rather than each community discussing the discrepancy to improve the understanding and interpreting of the Swagger Specs.

Here is StrongLooop / Loopback discussion around Enums 

- [New type: ENUM #1321](https://github.com/strongloop/loopback/issues/1321#issuecomment-216868272)

We still need to investigate where the StrongLoops/ IBM thoughts on implementing Enum, and we plan to review shortly.

We just propose that all participants in providing a Swagger Code Generators to deal with Enum to converge on a solution rather than have a lot of silo implementations of swagger generic extensions for Enum.

We have investigated the following code generators to date.  Non have a consist implementation to deal with **Additional Properties** or **Enum**

- [Azure AutoRest](https://github.com/Azure/autorest)
- [NSwag ](http://nswag.org)
- [Swagger-API Online Editor](http://editor.swagger.io)
- [Swagger-API CodeGen](https://github.com/swagger-api/swagger-codegen )
- [Loopback Swagger Generator](https://docs.strongloop.com/display/public/LB/Swagger+generator ) 

Appreciate feedback, concerns and issues with our thoughts…

PR #447, opened 2015-11-03
Reorder custom client credentials and body generation

matthchr:
- Only done for C# since other languages don't seem to support it.


matthchr:
Is this feature not fully supported in other languages?  I didn't see MethodTemplate.cshmtl for other languages having this same structure.

PR #449, opened 2015-11-04
Implemented paging for NodeJS

amarzavery:


AutoRest/Generators/Java/Azure.Java.Tests/src/main/java/fixtures/paging/models/ProductResult.java
```diff
@@ -1,66 +0,0 @@
-/**
```
amarzavery: @jianghaolu - When I executed gulp regenerate:expected, this Paging file got deleted. It seems right to me. However, it would be nice if you can confirm that this is fine.

jianghaolu: Yes, it's fine!

Sorry it's my my bad not removing that file - looks like regenerate:expected:javaazure does not remove before regeneration.

amarzavery: Oh no that is completely fine. It happens with me so many times.. Thanks for confirming.

PR #463, opened 2015-11-06
[#107070538] Updated ctors to be based on Credentials property

stankovski:


markcowl:
LGTM once the build passes.  Nice simplification

PR #464, opened 2015-11-06
Parameter groups should support "postfix" for naming

matthchr:


NuGet.Config
```diff
@@ -6,4 +6,7 @@
   <packageSources>
     <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
   </packageSources>
+  <disabledPackageSources>
+    <add key="nuget.org" value="true" />
```
yugangw-msft: expected change?

yugangw-msft:
Could you update test scenario or at least regenerate the C# code to ensure your change does become effective. The code change itself looks good.

NuGet.Config
matthchr: No.

matthchr:
@yugangw-msft Yes, I will update this PR.

matthchr:
I will look at the build failure and see why it failed -- looks like I need to merge and update...

matthchr:
@yugangw-msft This is good to merge now provided you're okay with the changes and the jenkins build passes.

brjohnstmsft:
Included with this PR was a fix for this issue: https://github.com/Azure/autorest/issues/479

PR #483, opened 2015-11-11
Add customize exception supports.

xingwu1:
Instead of general HttpOperationException, the new csharp generated code will have xxxException.cs which xxx is the default response type name.


xingwu1:
Add new extension "x-ms-exception", which has a single element "name", if the error type has this extension, the generated Exception class name will use the extension value instead of type plus "Exception".

For example: one operation has the default response:

          "default": {
            "schema": {
              "$ref": "#/definitions/A"
            }
          }

The type A is defined as:

    "A": {
      "x-ms-exception": {  "name": "MyException"   },
      "properties": {
        "statusCode": {
          "type": "string"
        }
      }

The generated Exception will be named as "MyException" instead of "AException".

Also, C# generated Exception will inherit from RestException instead of HttpOperationException.

AutoRest/AutoRest.Core/ClientModel/ServiceClient.cs
```diff
@@ -24,6 +24,7 @@ public ServiceClient()
             Methods = new List<Method>();
             ModelTypes = new List<CompositeType>();
             EnumTypes = new List<EnumType>();
+            Exceptions = new HashSet<CompositeType>();
```
stankovski: Can we rename this to ErrorTypes?

stankovski: Also add a TODO to change ModelTypes and EnumTypes to ISet.

AutoRest/AutoRest.Core/CodeNamer.cs
```diff
@@ -109,6 +109,10 @@ public virtual void NormalizeClientModel(ServiceClient client)
             foreach (var method in client.Methods)
             {
                 NormalizeMethod(method);
+                if (method.DefaultResponse != null && method.DefaultResponse is CompositeType)
```
stankovski: This belongs to SwaggerModeler.cs `public override ServiceClient Build()`

AutoRest/TestServer/swagger/httpInfrastructure.json
```diff
@@ -2795,6 +2795,9 @@
       }
     },
     "A": {
+      "x-ms-exception": {
```
stankovski: Why not name `A` as `MyException`? Do you have a strong use case for custom exception names?

xingwu1: In batch scenario, we have error type named: BatchError, but we would like to make the exception named: BatchException instead of BatchErrorException. This extension is optional.

xingwu1: Same as ARM. ARM has error named CloudError, but the exception is CloudException.

AutoRest/AutoRest.Core/ClientModel/ServiceClient.cs
xingwu1: Updated.

AutoRest/AutoRest.Core/CodeNamer.cs
xingwu1: Fixed.

AutoRest/TestServer/swagger/httpInfrastructure.json
stankovski: In that case can you make it a general purpose extension: `x-ms-client-name: name` which can be applied to any model (not just errors). The less extensions the better.

brjohnstmsft: :+1: If `x-ms-client-name` could be applied to model properties too, that would be even better.

xingwu1: Change to "x-ms-client-name".

Issue #478, opened 2015-11-11
AutoRest generates duplicate constructors for model with no properties


brjohnstmsft:
Given this Swagger:

    "DataChangeDetectionPolicy": {
      "discriminator": "@odata.type",
      "properties": {
        "@odata.type": {
          "type": "string"
        }
      },
      "required": [
        "@odata.type"
      ]
    },
    "SqlIntegratedChangeTrackingPolicy": {
      "allOf": [
        {
          "$ref": "#/definitions/DataChangeDetectionPolicy"
        }
      ]
    },

AutoRest generates two default constructors in C# for both `DataChangeDetectionPolicy` and `SqlIntegratedChangeTrackingPolicy`, which results in an error when compiling the generated code.

AutoRest should omit the second constructor when the model has no properties.

brjohnstmsft:
I'm going to try to fix this.

Issue #477, opened 2015-11-11
Derived class constructors are missing parameters for base class properties


brjohnstmsft:
Given this Swagger:

    "ScoringFunction": {
      "discriminator": "type",
      "properties": {
        "type": {
          "type": "string"
        },
        "fieldName": {
          "type": "string"
        },
        "boost": {
          "type": "number",
          "format": "double"
        },
        "interpolation": {
          "$ref": "#/definitions/ScoringFunctionInterpolation"
        }
      },
      "required": [
        "type"
      ]
    },
    "DistanceScoringFunction": {
      "allOf": [
        {
          "$ref": "#/definitions/ScoringFunction"
        },
        {
          "properties": {
            "distance": {
              "$ref": "#/definitions/DistanceScoringParameters"
            }
          }
        }
      ]
    },

AutoRest generates the following non-default constructors in C#.

For `ScoringFunction`:

        public ScoringFunction(string fieldName = default(string), double? boost = default(double?), ScoringFunctionInterpolation? interpolation = default(ScoringFunctionInterpolation?))
        {
            FieldName = fieldName;
            Boost = boost;
            Interpolation = interpolation;
        }

For `DistanceScoringFunction`:

        public DistanceScoringFunction(DistanceScoringParameters distance = default(DistanceScoringParameters))
        {
            Distance = distance;
        }

The `DistanceScoringFunction` constructor ought to include the parameters from its base class counterpart in `ScoringFunction`.

brjohnstmsft:
Further to this, if you mark the derived class properties as `required` in the Swagger spec, AutoRest does not emit non-defaulted parameters for those properties in the constructor. It does for required base class properties.

brjohnstmsft:
I'm going to try to fix this.

PR #497, opened 2015-11-17
Fixing issues with constructor generation in C#

brjohnstmsft:
When a derived class has no properties, AutoRest would generate a second
default constructor, resulting in an error when compiling the generated code.
See related issue: https://github.com/Azure/autorest/issues/478


brjohnstmsft:
Adding another related fix to this PR, for this issue: https://github.com/Azure/autorest/issues/477

stankovski:
Thanks!

PR #499, opened 2015-11-18
 [do not merge] fixed method naming for long running operations that are reserved words in nodejs

amarzavery:


AutoRest/Generators/Extensions/Azure.Extensions/AzureExtensions.cs
```diff
@@ -214,7 +221,15 @@ public static void AddLongRunningOperations(ServiceClient serviceClient)
                     if (isLongRunning is bool && (bool)isLongRunning)
                     {
                         serviceClient.Methods.Insert(i, (Method) method.Clone());
-                        method.Name = "Begin" + method.Name.ToPascalCase();
+                        if (codeNamer != null)
+                        {
+                            method.Name = "Begin" + codeNamer.GetMethodName(method.Name).ToPascalCase();
```
amarzavery: delete is a reserved word in nodejs. Hence methods named as "delete" are renamed as "deleteMethod" in generic nodejs. Cloning of delete to beginDelete in "Azure.NodeJS" happens before renaming of delete to deleteMethod in generic. beginDelete is not a reserved word hence does not get converted to beginDeleteMethod. Hence we make this change so that beginDelete is accurately coverted to beginDeleteMethod

AutoRest/Generators/Extensions/Azure.Extensions/AzureExtensions.cs
```diff
@@ -77,7 +78,12 @@ public static void NormalizeAzureClientModel(ServiceClient serviceClient, Settin
             FlattenResourceProperties(serviceClient);
             FlattenRequestPayload(serviceClient, settings);
             AddPageableMethod(serviceClient);
-            AddLongRunningOperations(serviceClient);
+            CodeNamer codeNamer = default(CodeNamer);
+            if (settings != null && settings.CodeGenerator == "Azure.NodeJS")
```
stankovski: It's a good idea to have a dependency on code generator from extensions. Dependency should go the other way.

PR #505, opened 2015-11-23
Relax Json dependency

### LINKED ISSUE #476: Update JSON gem dependency, since 1.8.3 requires native code compilation
The [JSON gem](https://github.com/flori/json) carries with it native C extensions, that require a compiler toolchain be present on any machine that installs the latest `1.8.3` gem (since there are no precompiled binaries available for this gem).

Ruby carries the following JSON gem versions with the distribution by default:

* JSON 1.4.2 with 1.9.2 - https://github.com/ruby/ruby/blob/ruby_1_9_2/ext/json/lib/json/version.rb
* JSON 1.5.5 with 1.9.3 - https://github.com/ruby/ruby/blob/ruby_1_9_3/ext/json/lib/json/version.rb
* JSON 1.7.7 with 2.0.0 series - https://github.com/ruby/ruby/blob/ruby_2_0_0/ext/json/lib/json/version.rb
* JSON 1.8.1 with 2.1.0 - through 2.1.7 - https://github.com/ruby/ruby/blob/v2_1_7/ext/json/lib/json/version.rb
* JSON 1.8.1 with 2.2.0 - through 2.2.3 - https://github.com/ruby/ruby/blob/v2_2_3/ext/json/lib/json/version.rb

Requiring `1.8.3` unfortunately prevents Windows Puppet agents (latest, based on a vendored Ruby 2.1.7) from being to able to consume the Azure Ruby SDK (*) and use the Puppet Azure module (under development).  Please consider changing the minimum required gem version to something older based on the earliest version of Ruby the Azure SDK supports - see the changelog at https://github.com/flori/json/blob/master/CHANGES for more details on what was fixed in each version. 

A couple of solutions are:

* Lowering the version requirement from `~> 1.8.3` to `>= 1.8.1` or `>= 1.7.7` in https://github.com/Azure/autorest/blob/master/ClientRuntimes/Ruby/ms-rest-azure/ms_rest_azure.gemspec#L31
* Completely removing the version associated with the JSON gem
* Removing the reference to the Gem outright, since Ruby distributions ship with it

(*) Technically, users may install a compiler toolchain in production with the DevKit, but not only is this never really done in practice, we actively discourage it for security reasons.

Thanks in advance!

PS - I also see that the `json` reference in the `gemspec` at https://github.com/Azure/autorest/blob/master/ClientRuntimes/Ruby/ms-rest/ms_rest.gemspec can be removed entirely given that code makes no calls to JSON libraries.

UPDATE: Original request updated based on new information - sorry for any confusion the initial errant request might have caused.

devigned:
resolves #476


devigned:
wrong target branch... closing

PR #509, opened 2015-11-30
Optional parameters grouped into options object for nodejs methods

amarzavery:
- User need not explicitly specify null for optional parameters in the method call


amarzavery:
@BretJohnson - Please review this PR. I did not make much changes in the MethodTemplateTS.cshtml and any TS methods in MethodTemplateModel.cs as I am not sure what would be the right thing to do over there. 
For javascript, it is clumsy for the end user to specify null for the optional parameter.
For example
```
VirtualMachineExtensionImages.prototype.listVersions = 
function (location, publisherName, type, filter, top, orderBy, options, callback) {
```
where,
location -> required
publisherName -> required
type -> required
filter-> optional
top -> optional
orderBy -> optional
options -> optional
callback ->  required

so previously the user would say 
`listVersions("West US", "Microsoft", "SomeType", null, null, null, callback)` //This is clumsy

With this PR , all the optional parameters are shoved into the options parameter (one before callback)
so, `listVersions("West US", "Microsoft", "SomeType", callback)` would work seamlessly.

I don't know what should be done in TS to achieve this. 

At the same time can you let me know what should I do in the .ts files under `AutoRest\AutoRest\Generators\NodeJS\NodeJS.Tests\AcceptanceTests` so that my tests pass.

I made the necessary changes in the equivalent .js test files and all the tests pass locally there after.

Can you please document the steps that one needs to do before running the TS tests.

PR #524, opened 2015-12-05
Support typed response headers for all operations

matthchr:
- Also update tests to check that deserialization works
    for all header response objects.


matthchr:
@stankovski I don't really love the way that there is an AzureOperationResponse and an AzureOperationHeaderResponse -- but I can't think of a better way and as long as we have the IAzureOperationResponse interface I think nothing is really lost.

Let me know what you think.  

Supporting this for all operations is (IMO) really good though - it actually means you could get rid of the AzureOperationResponse type entirely (since all that has is RequestId and you could just make people put that in their swagger spec instead of magically pulling it out).

PR #531, opened 2015-12-08
Bumped up version of Client Runtime related to the addition of header response

stankovski:


matthchr:
@stankovski Do you want to hold off on doing this for a bit just to make sure I don't run into any other changes?  I'll finish integrating stuff into our client early tomorrow probably and should be able to see then if I missed any features?

stankovski:
@matthchr I'm not planning to publish the client runtime yet. We just try to always be in a state where changed packages have a higher version then ones in nuget/npm. This way we can publish at any time without needing to go through version update process.

stankovski:

PR #535, opened 2015-12-08
Updated C# code gen to read both content and message headers when generating a Header object

stankovski:


ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/HttpExtensions.cs
```diff
@@ -164,5 +164,46 @@ public static JObject ToJson(this HttpResponseHeaders headers)
...
+        /// <returns>Json string</returns>
+        public static JObject GetHeadersAsJson(this HttpResponseMessage message)
+        {
+            if (message == null)
+            {
+                return new JObject();
+            }
+
+            var jObject = new JObject();
+            foreach (var httpResponseHeader in message.Headers)
```
matthchr: Can message.Headers be null?

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime.Tests/JsonSerializationTests.cs
```diff
@@ -262,5 +263,21 @@ public void HeaderGetsSerializedToJson()
                 Environment.NewLine);
             Assert.Equal(expectedJsonString, json);
         }
+
+        [Fact]
+        public void HeaderKnownValuesGetSerializedToJson()
+        {
+            var message = new HttpResponseMessage();
+            message.Content = new StringContent("Test");
```
matthchr: Can you please get Content-Length or Content-Type (a "well-named" header)?

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/HttpExtensions.cs
stankovski: It's not really documented anywhere but based on the source code it should never be null - https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/src/System/Net/Http/HttpResponseMessage.cs#L99

matthchr:
@stankovski This looks good to me

Issue #372, opened 2015-09-30
Generated C# code should avoid using JsonConvert


brjohnstmsft:
See related issue in the pre-AutoRest Search SDK: https://github.com/Azure/azure-sdk-for-net/issues/1438

`JsonConvert` has a static `DefaultSettings` property. Unfortunately, even if you pass your own `JsonSerializerSettings` to `SerializeObject`/`DeserializeObject`, they do not completely replace the default settings. Some settings like `JsonConverters` are merged rather than replaced.

What this means is that code using an AutoRest-generated client library might modify `JsonConvert.DefaultSettings` in ways that disrupt the serialization/deserialization of payloads by the client library. Instead, the generated code should call `JsonSerializer` directly. It is simple to write some wrappers that mimic `JsonConvert` (see this PR for an example: https://github.com/Azure/azure-sdk-for-net/pull/1451).

stankovski:
Tracking via https://www.pivotaltracker.com/story/show/107069918

brjohnstmsft:
I'm going to try to fix this.

PR #538, opened 2015-12-09
Ensuring that generated C# code does not depend on JsonConvert

brjohnstmsft:
This is a fix for this issue: https://github.com/Azure/autorest/issues/372
The solution adds a new SafeJsonConvert class to ClientRuntime and references it from all generated code.


devigned:
LGTM. @stankovski thoughts?

Issue #278, opened 2015-08-03
IFooManagementClient should implement IDisposable


tonytang-microsoft-com:
This will allow caller to consume IFooManagementClient directly instead of consuming FooManagementClient implementation class.

devigned:
:+1:

devigned:
I smell a pull request!

PR #548, opened 2015-12-11
Changed service interface in C# to implement IDisposable (#278)

stankovski:


stankovski:
Implements https://github.com/Azure/autorest/issues/278

PR #556, opened 2015-12-14
Bumped up version of AutoRest and ClientRuntime

stankovski:


amarzavery:
LGTM. Is this for the next release in jan or is there some bug fix that we want to publish right now?

PR #562, opened 2015-12-17
Implemented form-data support and file upload in NodeJS and C#

stankovski:


AutoRest/Generators/CSharp/CSharp/Templates/MethodTemplate.cshtml
```diff
@@ -105,12 +105,58 @@ public async Task<@(Model.OperationResponseReturnTypeString)> @(Model.Name)WithH
 
     @if (Model.RequestBody != null)
     {
+        if (Model.RequestBody.Type == PrimaryType.Stream)
+        {
+        <text>
+    // Serialize Request    
+    StreamContent _fileStreamContent = new StreamContent(@(Model.RequestBody.Name));
+    _httpRequest.Content = _fileStreamContent;
+    _httpRequest.Content.Headers.ContentType = MediaTypeHeaderValue.Parse("@(Model.RequestContentType)");
```
begoldsm: Will this also support octet-stream?

stankovski: What should be incorporated into RequestContentType?

begoldsm: application/octet-stream. Is that somewhere that it should go? I was under the impression that RequestContentType controlled what the request content type was. If so, with this change, is "application\octet-stream" a valid content type that will be returned by this parse?

PR #566, opened 2015-12-18
Use 6.0.8 of newton.json when build for 4.5

yugangw-msft:
Otherwise, conflicts will happen once a client application, like azure powershell, pulls in older version of azure-sdk-for-net library which still uses 6.0.* version


ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/project.json
```diff
@@ -24,6 +24,9 @@
         "System.Net.Http": "",
         "System.Net.Http.WebRequest": "",
         "System.Xml": ""
+      },
+      "dependencies": {
+        "Newtonsoft.Json": "6.0.8"
```
brjohnstmsft: This seems like a step backwards. What's the plan for eventually bumping this to 7.x? We can't stay on this version forever.

yugangw-msft: IMO, the dependency version should be the minimum, we can move to a higher one but should come with a reason that the higher one provide what we need. This way, the library can have the most audience. For autorest, we have to use 7.0 because of the dnx support, out of dnx scenario, the 6.0.x is enough.

brjohnstmsft: JSON.NET does fix bugs from time to time. I reported an issue that was fixed in 7.0.1: https://github.com/JamesNK/Newtonsoft.Json/issues/672  It doesn't address a common scenario, but it has affected some of our customers in the past.

More generally, I don't think it's good policy to freeze dependency versions for all time. The question shouldn't really be *if*, but *when* should it be updated.

stankovski: We need to do this for now, unfortunately, but we will upgrade everyone to 7.0.1 after Jan powershell release.

PR #571, opened 2015-12-23
Added OperationName property to x-ms-pageable

stankovski:


AutoRest/Generators/Extensions/Azure.Extensions/Model/PageableExtension.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+
```
hovsepm: Remove this line please

hovsepm:
LGTM

AutoRest/Generators/Java/Java.Tests/src/main/java/fixtures/validation/AutoRestValidationTest.java
```diff
@@ -84,10 +84,10 @@
      * used by Retrofit to perform actually REST calls.
      */
     interface AutoRestValidationTestService {
-        @GET("/fakepath/{subscriptionId}/{resourceGroupName}/{id}?api-version={apiVersion}")
+        @GET("/fakepath/{subscriptionId}/{resourceGroupName}/{id}")
```
brjohnstmsft: Why is api-version now removed in the generated code?

PR #574, opened 2015-12-28
Added documentation on Swagger extensions

stankovski:


Documentation/swagger-extensions.md
```diff
@@ -0,0 +1,180 @@
...
+The following documents describes AutoRest specific vendor extensions for [Swagger 2.0](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md) schema. Some of the extensions are only applicable to Microsoft Azure and as such are only available in Azure code generators (e.g. Azure.CSharp, Azure.NodeJS, etc.).
+
+## Generic Extensions
+* [x-ms-skip-url-encoding](#x-ms-skip-url-encoding) - skips URL encoding for path and query parameters
+* [x-ms-enum](#x-ms-enum) - additional metadata for enums
+* [x-ms-parameter-grouping](#x-ms-parameter-grouping) - groups method parameters in generated clients
+* [x-ms-paths](#x-ms-paths) - alternative to [Paths Object](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#pathsObject) that allows [Path Item Object](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#pathItemObject) to have query parameters for non pure REST APIs
+* x-ms-client-name - *not currently implemented*
+* [x-ms-external](#x-ms-external) - allows specific [Definition Objects](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#definitionsObject) to not be excluded from code generation
```
brjohnstmsft: "to not be excluded" -> "to be excluded"

Documentation/swagger-extensions.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Introduction
+The following documents describes AutoRest specific vendor extensions for [Swagger 2.0](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md) schema. Some of the extensions are only applicable to Microsoft Azure and as such are only available in Azure code generators (e.g. Azure.CSharp, Azure.NodeJS, etc.).
+
+## Generic Extensions
+* [x-ms-skip-url-encoding](#x-ms-skip-url-encoding) - skips URL encoding for path and query parameters
+* [x-ms-enum](#x-ms-enum) - additional metadata for enums
+* [x-ms-parameter-grouping](#x-ms-parameter-grouping) - groups method parameters in generated clients
+* [x-ms-paths](#x-ms-paths) - alternative to [Paths Object](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#pathsObject) that allows [Path Item Object](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#pathItemObject) to have query parameters for non pure REST APIs
+* x-ms-client-name - *not currently implemented*
```
brjohnstmsft: :+1:

Documentation/swagger-extensions.md
```diff
@@ -0,0 +1,180 @@
...
+* [x-ms-external](#x-ms-external) - allows specific [Definition Objects](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#definitionsObject) to not be excluded from code generation
+
+## Microsoft Azure Extensions
+* x-ms-odata
+* x-ms-pageable
+* x-ms-long-running-operation
+* x-ms-azure-resource
+* x-ms-discriminator-value
```
brjohnstmsft: Custom polymorphic discriminators have value outside Azure-specific scenarios. I'm pretty sure I implemented this extension in the base (non-Azure) code generators.

Documentation/swagger-extensions.md
```diff
@@ -0,0 +1,180 @@
...
+    "x-ms-skip-url-encoding": true
+  }
+]
+```
+
+## x-ms-enum
+Enum definitions in Swagger indicate that only a particular set of values may be used for a property or parameter. When the property is represented on the wire as a string, it would be a natural choice to represent the property type in C# as an enum. However, not all enumeration values should necessarily be represented as C# enums - there are additional considerations, such as how often expected values might change, since adding a new value to a C# enum is a breaking change requiring an updated API version. Additionally, there is some metadata that is required to create a useful C# enum, such as a descriptive name, which is not represented in swagger. For this reason, enums are not automatically turned into enum types in C# - instead they are rendered in the documentation comments for the property or parameter to indcate allowed values. To indicate that an enum will rarely change and that C# enum semantics are desired, use the `x-ms-enum` exension.
+
+In C#, an enum type is generated and is declared as the type of the related request/response object. The enum is serialized as the string expected by the REST API.
```
brjohnstmsft: Do generators for other languages currently ignore x-ms-enum? Either way, we should call it out here.

stankovski:
@brjohnstmsft - thanks for the review

PR #575, opened 2015-12-29
Python generator version 0.9

xingwu1:
The Python generator for autorest. It will generate both generic python and azure ARM desired python. The generated code should be compatible with both python 2.7 and python 3.5. The check in include: client runtime for generic python and azure python, generator for generic python and azure python, and the test cases for client runtime and generator.


AutoRest/AutoRest.Core/Settings.cs
```diff
@@ -318,8 +318,7 @@ public void Validate()
             {
                 foreach (var unmatchedSetting in CustomSettings.Keys)
                 {
-                    Logger.LogError(new ArgumentException(unmatchedSetting),
-                        Resources.ParameterIsNotValid, unmatchedSetting);
```
yugangw-msft: this is a breaking change, any specific reason you want to do this?

AutoRest/AutoRest.Core/CodeNamer.cs
```diff
@@ -29,6 +30,46 @@ protected CodeNamer()
         /// </summary>
         public HashSet<string> ReservedWords { get; private set; }
 
+        // "joined_lower" for functions, methods, attributes
+        // do not invoke this for a class name which should be "StudlyCaps"
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Globalization", "CA1308:NormalizeStringsToUppercase")]
+        public static string PythonCase(string name)
```
yugangw-msft: is it feasible to move this logic to python's own codenamer?

AutoRest/AutoRest.Core/Utilities/Extensions.cs
```diff
@@ -195,6 +195,16 @@ private static PropertyInfo GetBaseProperty(this Type type, string propertyName)
         }
 
         /// <summary>
+        /// Converts the specified string to a python style string.
+        /// </summary>
+        /// <param name="value">The string to convert.</param>
+        /// <returns>The python style string.</returns>
+        public static string ToPythonCase(this string value)
```
yugangw-msft: can this be moved to python's own code-gen project?

PR #582, opened 2015-12-31
Dispose requests and responses

iouri-s:
Note that when we throw an exception, we will now be sending a disposed request and response. I am not sure how useful these objects are in disposed state.


iouri-s:
Sorry, this is not the right fix as response and request are returned to the caller. A better fix is to make AzureOperationResponse disposable.

PR #585, opened 2016-01-06
[DNM] Support reusing grouping types for paging next operation

jianghaolu:
If a grouped parameter only consists header operations, for which case the Next() operation will reuse the parameter type, instead of generating one for its own.

Original bug:
When a paging method
`GetMultiplePagesWithHttpMessagesAsync()`
generates
`GetMultiplePagesWithHttpMessagesNextAsync()`
method, the parameter schema definition is also duplicated. For example,
```csharp
GetMultiplePagesWithHttpMessagesAsync(string clientRequestId = default(string), PagingGetMultiplePagesOptions pagingGetMultiplePagesOptions = default(PagingGetMultiplePagesOptions),  <- This is parameter group class
GetMultiplePagesWithHttpMessagesAsync(string clientRequestId = default(string), PagingGetMultiplePagesNextOptions pagingGetMultiplePagesNextOptions = default(PagingGetMultiplePagesNextOptions),

```
The second definition `BatchListNextOption` should not be generated.


jianghaolu:

jianghaolu:

jianghaolu:
pending this afternoon's discussion.

Issue #570, opened 2015-12-22
Generate Client constructors are protected (csharp)


dotdot2:
With the 0.13 version the client class has protected constructors. I remember earlier it was public.

Why the change? Should I just modify the protected to public?

stankovski:
Protected ctors are not meant to be used by end users as they would put the client in the non-usable state (e.g. if no credentials are passed). We recommend that only public ctors be exposed.

dotdot2:
I understand that. But in this case I did not see even a single public constructor. Is that expected?

Btw, can we keep this issue open before a clear resolution is reached?

stankovski:
Can you please provide a sample swagger for repro?

dotdot2:
[projectconfig.zip](https://github.com/Azure/autorest/files/70528/projectconfig.zip)

maxkeller:
I've got the same issue. Commit was https://github.com/Azure/autorest/commit/85b8055087ebc3842acaefc8347df3c25f459617

Looking at the templates for it, it seems like this option just went untested ;) I can provide a proper constructor for this if you want me to - whether it be the empty constructor that I use now or one that takes Credentials as an argument. I might need some help setting up that build for testing it, though.

stankovski:
Thanks for the sample. The issue should be fixed in #592.

dotdot2:
Awesone, thanks

PR #592, opened 2016-01-08
Fixed constructor when -AddCredentials is used and there are no other client properties (issue #570)

stankovski:
Fixes issue #570


ingAlfano:
Hi, will we get a new nuget release anytime soon?

PR #610, opened 2016-01-13
Exposing ClientRequestId in Azure client

stankovski:


stankovski:
@brjohnstmsft take a look at the PR again. We've changed client level ClientRequestId to bool GenerateClientRequestId (I know - nullable bool is not optimal, but we don't have an option right now)

PR #615, opened 2016-01-14
Clarify LongRunningOperationRetryTimeout

iouri-s:


AutoRest/Generators/Extensions/Azure.Extensions/AzureExtensions.cs
```diff
@@ -295,7 +295,7 @@ public static void AddAzureProperties(ServiceClient serviceClient)
                 Name = "LongRunningOperationRetryTimeout",
                 SerializedName = "longRunningOperationRetryTimeout",
                 Type = PrimaryType.Int,
-                Documentation = "The retry timeout for Long Running Operations."
+                Documentation = "The retry timeout for Long Running Operations in milliseconds."
```
amarzavery: This is in seconds with default of 30 seconds. We just fixed this in another PR over [here](https://github.com/Azure/autorest/pull/610/files#diff-7df3f3e4d5191fab3fd05d888ce0e9d3). Hence, closing this one.

Issue #619, opened 2016-01-15
Add ability to specify code gen settings via swagger spec


stankovski:

brjohnstmsft:
:+1:

Issue #618, opened 2016-01-14
Add setting for C# gen to specify constructor protection level


stankovski:
If set constructors should be generated as internal

brjohnstmsft:
What's the scenario for this?

stankovski:
Allow to generate clients with "internal" ctor so that the client can not be initialized directly. Customization layer would be used to wrap the client and customize it as needed. This is necessary for scenarios where clients need to be forcefully pre-configured (e.g. custom delegating halndlers, custom retry policy, etc.).

brjohnstmsft:
Cool, that's something we're interested in for the Search SDK as well. But why stop at the constructor? Why not make the classes themselves internal?

stankovski:
@brjohnstmsft keeping class public allows developers to re-use the entire client and just hide the ctor. For example you can have a custom ctor that takes in an account name and appends it to base Uri as is the case in DataLake

PR #621, opened 2016-01-15
Add ability to specify code gen settings via swagger spec and internal ctors in C#

stankovski:
Add ability to specify code gen settings via swagger spec #619
Add setting for C# gen to specify constructor protection level #618


stankovski:
@azuresdkci test this please

Issue #624, opened 2016-01-19
Wrap request and response in a new object to avoid being disposed


stankovski:
Currently when we throw an exception, we are now sending a disposed request and response (C#). Need to wrap the data in a new object to preserve it for the end user.

PR #625, opened 2016-01-19
Wrapped C# request and response in a new object to avoid being disposed #624 [Delivers #110445462]

stankovski:
#624


markcowl:
@stankovski As dicussed, we should always dispose of the response message if we are not returning it (so generated code should dispose of it).

We should consider having OperationResponse and AzureOperationResponse simply contain the request and response, and just disposing the response from the OperationResponse  in extension methods.

We should probably dispose the response when throwing an exception after the request is made.

markcowl:
LGTM, modulo the change we discussed.

PR #626, opened 2016-01-19
Android interactive login flow

jianghaolu:
The sample that uses the this is in https://github.com/Azure/azure-sdk-for-java/pull/579


jianghaolu:

PR #631, opened 2016-01-21
initialize requestContent to be an empty string

amarzavery:


amarzavery:
closing this as it does not make sense ....

PR #661, opened 2016-01-29
Generate comments for all python classes.

xingwu1:
1. Get rid useless namespace and functions
2. Generate comments to all python classes.
3. Change Template:GetIntentation() from function to property due to change access modifier to protected. (A new python template is forking WrapComment function to ParameterWrapComment to add extra whitespace in front of the next following lines.)


stankovski:
@xingwu1 please generate tests

xingwu1:
@stankovski  Please review it again.

Issue #513, opened 2015-12-01
(C#, Java) We need a way to model non-nullable readOnly properties


brjohnstmsft:
Sometimes read-only properties (i.e. -- properties that are only ever in a response body) are guaranteed by the REST API to never be null. When these properties map to value types in .NET or primitive types in Java, they should be modeled as non-nullable types. Currently all AutoRest-generated properties are nullable.

This [issue](https://github.com/Azure/autorest/issues/375) is related. Assuming we fix that issue, we'll be able to ensure that constructor parameters and model properties in C# and Java are non-nullable if they are marked as `required` in the Swagger. This would also be very useful for some `readOnly` properties, but there is no way to express this in Swagger today.

I see two short-term options that don't involve changes to Swagger itself:
1. Extend AutoRest to allow properties to be both `required` and `readOnly` at the same time.
2. Add a new extension, like `x-ms-required-in-response` (naming suggestions welcome) that would behave like `required` with respect to nullability, but for `readOnly` properties.

Either way, we'd also need to fix [the related issue](https://github.com/Azure/autorest/issues/375).

The downside of option 1 is that it would break the rules for `readOnly` in the [Swagger specification](http://swagger.io/specification/#schemaObject):

> Relevant only for Schema "properties" definitions. Declares the property as "read only". This means that it MAY be sent as part of a response but MUST NOT be sent as part of the request. Properties marked as readOnly being true SHOULD NOT be in the required list of the defined schema. Default value is false.

The semantics of `readOnly` + `required` do seem muddy, and I buy that they ought to be mutually exclusive. For this reason I tend to favor option 2.

@devigned @Azure/adx-sdk-team Thoughts?

brjohnstmsft:
This issue also affects responses that only contain a primitive value.

stankovski:
Related #375

NiklasGustafsson:
Also see: https://github.com/Azure/autorest/issues/816. Duplicate issue, with a third option proposed.

fearthecowboy:
C# supports `x-nullable` now. 
Java may be a bit yet.

olydis:
Just tested this on Java - works for me! Example:
```YAML
  someModel:
    type: object
    properties:
      prop:
        type: integer
        format: int32
        readOnly: true
        x-nullable: false # toggle me :-)
```
Will therefore close this issue, please reopen if something is missing.

Issue #492, opened 2015-11-13
C# method generated for HTTP HEAD operation should return bool, not bool?


brjohnstmsft:
Returning bool? is inconvenient for the caller. HEAD will always tell you precisely whether a resource exists or not, so there is no reason for the return type of the method to be nullable.

stankovski:
Relates to #375

Issue #375, opened 2015-10-02
Required value types should not be nullable if they are required


devigned:
Required query parameter still convert to nullable type:
```json
{
  "name": "timeout",
  "in": "query",
  "required": true,
  "type": "integer",
  "format": "int32",
},
```
The generated C# code:
```csharp
int? timeout,
```
 
caused by CodeNamer.NormalizeType function.

brjohnstmsft:
It would be great to see this fixed. Let me know if I can help.

devigned:
@xingwu1 and @stankovski where did we land on this in our last conversation?

stankovski:
All parameters are validated during the runtime, so the issue here is about compile time validation. Since some data types in C# (such as strings or classes) can not be made non-nullable we will always end up with a mix of compile time and runtime validation. Our decision to have all parameters as nullable has simplified the code generation logic significatly, as such unless there is a strong case for adding non-nullable primitives back I would recommend against it.

stankovski:
@devigned please re-open if you want to persue this change

brjohnstmsft:
@devigned @stankovski Just to clarify, are we only talking about parameters here, or properties too? I would love for my customers to not have to worry about nullability of model properties in situations where it doesn't make sense.

As a matter of principle, we should make decisions based on what's best for the developer experience and not ease of implementation.

xingwu1:
Totally agree @brjohnstmsft, actually compare to validation of assignment of required primitive type, I think the accurate of model properties type is more important.

brjohnstmsft:
@xingwu1 Exactly -- I don't actually need the generated code to validate model properties, since the REST API already does this.

PR #701, opened 2016-02-12
Converted required primitive parameters and properties to be non-nullable

stankovski:
Implements #492, ~~#513~~, and #375


stankovski:
@brjohnstmsft can you take a look at this PR to make sure it addresses #492, #513, and #375?

brjohnstmsft:
Looking...

brjohnstmsft:
The PR is too big to leave inline comments, so I'll just comment here. In `CSharpCodeNamer.cs`:

```csharp
private IType NormalizeSequenceType(SequenceType sequenceType)
{
    sequenceType.ElementType = NormalizeTypeReference(sequenceType.ElementType);
    if (sequenceType.ElementType.IsValueType())
    {
        sequenceType.NameFormat = "IList<{0}?>";
    }
    else
    {
        sequenceType.NameFormat = "IList<{0}>";
    }
    return sequenceType;
}

private IType NormalizeDictionaryType(DictionaryType dictionaryType)
{
    dictionaryType.ValueType = NormalizeTypeReference(dictionaryType.ValueType);
    if (dictionaryType.ValueType.IsValueType())
    {
       dictionaryType.NameFormat = "IDictionary<string, {0}?>";
    }
    else
    {
        dictionaryType.NameFormat = "IDictionary<string, {0}>";
    }
    return dictionaryType;
}
```
Does this mean that collection elements will always be nullable, even if the collections themselves are required?

brjohnstmsft:
In `AcceptanceTests.cs`:

```csharp
// NO LONGER VALID - Birthday is not nullable as such validation is removed
// var missingRequired =
//    Assert.Throws<ValidationException>(() => client.Polymorphism.PutValidMissingRequired(badRequest));
// Assert.Equal("Birthday", missingRequired.Target);
```

Why not just delete the commented-out code?

brjohnstmsft:
Looks good overall. Some general comments:
- #492 looks resolved
- #375 looks resolved
- #513 does not appear to be addressed since it applies specifically to `readOnly` properties, not `required` properties. There is still an open question of how to model nullable versus non-nullable `readOnly` properties (Swagger currently has no way to express this without a vendor-specific extension).
- Do you plan to apply this change to Java as well? It has the same issues, only in Java's case it's about the usage of "boxed" primitives like `java.lang.Integer` instead of `int`, etc.

stankovski:
@brjohnstmsft as far as nullability of collections, I didn't find anywhere in the swagger spec a way to say that the collection should not contain null values (http://swagger.io/specification/#itemsObject).

stankovski:
@brjohnstmsft you are correct. I didn't read #513 in details. This particular change will leave readOnly fields as nullable since they are by definition not required. 
@jianghaolu I believe has `Integer` vs `int` implemented in java code already (not for readOnly but for required fields).

Issue #672, opened 2016-02-02
Support parameterized host names using an extension


markcowl:
Add an extension:

x-ms-parameterized-host to the base swagger element in the swagger doc

```json
XMSParameterizedHostSchema: {
  "type": "object",
  "required": ["hostTemplate", "parameters"],
  "properties": {
        "hostTemplate": "string",
        "parameters": {
            "type": "array",
            "items": {
               "type":"object",
                "required": ["name", "description"],
                "properties": {
                     "name": "string",
                     "description": "string"
                 }
            }
      }
  }
}
```
for example:
```json
x-ms-parameterized-host: {
  "hostTemplate": "{accountName}.contoso.com",
   "parameters": [{"name":"accountName", "description":"The account id for your contoso account."}]
```

This extension would add a parameterized baseUri to each method, with the specified parameters added as string parameters, with the specified description documentation. The parameters would be substituted in the parameterized host template by name, and the resulting string would be used as the host portion of the URI for the request

brjohnstmsft:
:+1: This is a must-have for multi-tenant services.

One thing that would make this more useful is the ability to take host name as a constructor parameter of the client instead of making it a parameter of every operation. In the case of Azure Search for example, one tenant == one customer == one hostname, so that customer's application will always have the same value for hostname. It would be pointless to pass it around everywhere when it can just be part of the client.

For examples of how Azure Search handles this scenario, see our custom code for [SearchIndexClient](https://github.com/Azure/azure-sdk-for-net/blob/AutoRest/src/Search/Microsoft.Azure.Search/Customizations/SearchIndexClient.Customization.cs) and [SearchServiceClient](https://github.com/Azure/azure-sdk-for-net/blob/AutoRest/src/Search/Microsoft.Azure.Search/Customizations/SearchServiceClient.Customization.cs).

stankovski:
- [ ] Update swagger modeler to handle the new extension
- [ ] Update client model if necessary
- [ ] Implement changes in C# and add tests
- [ ] Implement changes in NodeJS and add tests
- [ ] Implement changes in Java and add tests
- [ ] Implement changes in Python and add tests
- [ ] Implement changes in Ruby and add tests
- [ ] Update [schema](https://github.com/Azure/autorest/blob/master/schema/swagger-extensions.json)
- [ ] Update [documentation](https://github.com/Azure/autorest/blob/master/Documentation/swagger-extensions.md)

PR #709, opened 2016-02-17
Adding extension: x-ms-parameterized-host

begoldsm:
I, along with A LOT of help from Denis, Amar and Jianghao (thanks so much guys!), have added support for the extension: x-ms-parameterized-host which is detailed in the open issue: https://github.com/Azure/autorest/issues/672

We have added support and tests for generic and azure generation for the modeler, C#, Java and NodeJs. At this point it is my understanding that Python and Ruby are not yet ready. Once development for these generators begins, this extension will also need to be added to them.


stankovski:
@azuresdkci add to whitelist

stankovski:
@begoldsm please file a tracking issue for Python and Ruby

begoldsm:
@stankovski 
The following issues are now open:
https://github.com/Azure/autorest/issues/722
https://github.com/Azure/autorest/issues/721

AutoRest/Generators/CSharp/Azure.CSharp/AzureCSharpCodeGenerator.cs
```diff
@@ -60,6 +60,7 @@ public override string ImplementationFileExtension
         /// <param name="serviceClient"></param>
         public override void NormalizeClientModel(ServiceClient serviceClient)
         {
+            Extensions.NormalizeClientModel(serviceClient, Settings);
```
stankovski: You shouldn't call Extensions.NormalizeClientModel here, since AzureExtensions.NormalizeAzureClientModel contains generic extensions as well.

AutoRest/Generators/Extensions/Extensions/Extensions.cs
```diff
@@ -30,6 +35,67 @@ public static void NormalizeClientModel(ServiceClient serviceClient, Settings se
         {
             FlattenRequestPayload(serviceClient, settings);
             AddParameterGroups(serviceClient);
+            ProcessParameterizedHost(serviceClient, settings);
```
stankovski: Add this line to AzureExtenions.NormalizeClientModel

AutoRest/Generators/CSharp/CSharp/ClientModelExtensions.cs
```diff
@@ -13,6 +13,22 @@ namespace Microsoft.Rest.Generator.CSharp
     public static class ClientModelExtensions
     {
         /// <summary>
+        /// Determine whether URL encoding should be skipped for this parameter
+        /// </summary>
+        /// <param name="parameter">The parameter to check</param>
+        /// <returns>true if url encoding should be skipped for the parameter, otherwise false</returns>
+        public static bool SkipUrlEncoding(this Parameter parameter)
```
stankovski: This is a good extension method and it should exist in the extensions. Please move it there and use it whenever we check for `Parameter.Extensions.ContainsKey(AzureExtensions.SkipUrlEncodingExtension)`.

PR #710, opened 2016-02-18
Update python codegen based feedback from bug bash

xingwu1:
1. Generate credentials.py/exceptions.py to reduce the classes imported at _init_.py
2. Change the way to init model class
3. Update AAD auth client lib.


AutoRest/Generators/Python/Azure.Python/TemplateModels/AzureServiceClientTemplateModel.cs
```diff
@@ -100,5 +102,13 @@ public override string SetupRequires
                 return "\"msrest>=0.0.1\", \"msrestazure>=0.0.1\"";
             }
         }
+
+        public override bool SupportsAzure
```
stankovski: What is this for? It looks odd given that this is an Azure.Python code gen.

xingwu1: Change to NeedsExtraImport to explicit reflect the usage of the property.

jianghaolu:

PR #712, opened 2016-02-19
wrap properties with single quotes if they have a dot in their name

amarzavery:


amarzavery:
@xingwu1 - This should fix the odata.nextLink problem.

PR #723, opened 2016-02-23
Added build target to produce zip package

stankovski:


build.proj
```diff
@@ -267,7 +268,14 @@
...
 
-  <Target Name="SignAssembliesInDnxPackages">
+  <Target Name="GetZip">
+    <Message Text="Ensure 7zip is available" />
+    <Exec
+      Command="$(LibraryToolsFolder)\AzCopy\AzCopy.exe /Source:https://azuresdktools.blob.core.windows.net/7-zip  /S /Dest:$(ZipExeFolder) /Y"
+      Condition="!Exists('$(ZipExe)')" />
+  </Target>
+  
+  <Target Name="SignAssembliesInDnxPackages" DependsOnTargets="GetZip">
```
yugangw-msft: this might not be enough, because i saw you have non codesign related tasks, such as "PackageAutoRestZip", also needs the zip executable

PR #728, opened 2016-02-24
Implemented Ignoring readOnly property in HTTP put.

haocs:


jianghaolu:
LGTM

@amarzavery this is the PR for the read-only feature. JFYI, the swagger path is here: https://github.com/Azure/autorest/pull/728/files#diff-08343c7d7dac04e0e8edf87a09725cecR1064.

Issue #746, opened 2016-02-26
CSharp: HttpOperationException response content is null


jhancock93:
When handling an unexpected status code, the current CSharp generator generators an HttpOperationException with null response content. If there was an HttpError in the response content, the error is lost to the client. In addition, because AutoRest currently handles error response codes like 400 as a success, this provides no good way to return an error.

The current code looks like:
            **string _responseContent = null;**
            if ((int)_statusCode != 200)
            {
                var ex = new HttpOperationException(string.Format("Operation returned an invalid status code '{0}'", _statusCode));
                ex.Request = new HttpRequestMessageWrapper(_httpRequest, _requestContent);
                ex.Response = new HttpResponseMessageWrapper(_httpResponse, **_responseContent**);
                if (_shouldTrace)
                {
                    ServiceClientTracing.Error(_invocationId, ex);
                }
                _httpRequest.Dispose();
                if (_httpResponse != null)
                {
                    _httpResponse.Dispose();
                }
                throw ex;
            }

As can be seen, the responseContent is null. The responseContent string is only grabbed from the message in case of a success.

stankovski:
@jhancock93 this should only happen if the swagger spec does not indicate that the operation returns a model (e.g. DELETE operation with void response). If you have a model in "default" response and your operation has "produces" with JSON you should get details in the error. Alternatively if you have any response coming back from the operation (e.i. we understand that there is a payload) the code gen will read it and add it to the exception even if there is no error model. This is similar to #647.

jhancock93:
This is a change from AutoRest v0.13 when the response was contained in the HttpOperationException regardless. This is a pretty simple code fix in the method template to address the issue and I will generate a pull request shortly. The suggestion in #647 seems like a workaround instead of a fix.

stankovski:
@jhancock93 thanks for the PR. I left a comment regarding your approach.

jhancock93:
Issue has been addressed with merged PR.

PR #749, opened 2016-02-26
Fixes problem with null response content in CSharp generator for non-…

jhancock93:
…documented codes

The main fix is in MethodTemplate.cshtml
The remainder of the changes are auto-generated files.

This addresses issue #746 by grabbing the response content string as already done in the success case.

### DIFF
--- a/AutoRest/Generators/CSharp/CSharp/Templates/MethodTemplate.cshtml
+++ b/AutoRest/Generators/CSharp/CSharp/Templates/MethodTemplate.cshtml
@@ -258,6 +258,12 @@ else
         @:    // Ignore the exception
         @:}
     }
+    else
+    {
+        //If not defined by default model, read content as string
+        @:_responseContent = await _httpResponse.Content.ReadAsStringAsync().ConfigureAwait(false);
+    }
+            
         ex.Request = new HttpRequestMessageWrapper(_httpRequest, _requestContent);
         ex.Response = new HttpResponseMessageWrapper(_httpResponse, _responseContent);
         @(Model.InitializeException)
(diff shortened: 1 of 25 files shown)


azurecla:
Hi __@jhancock93__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!

TTYL, AZPRBOT;

stankovski:
@azuresdkci test this please

AutoRest/Generators/CSharp/CSharp/Templates/MethodTemplate.cshtml
```diff
@@ -228,7 +228,7 @@ else
 
     HttpStatusCode _statusCode = _httpResponse.StatusCode;
     cancellationToken.ThrowIfCancellationRequested();
-    string _responseContent = null;
+    string _responseContent = await _httpResponse.Content.ReadAsStringAsync().ConfigureAwait(false);
```
stankovski: What if the payload is a stream? You will end up with OutOfMemory exception for APIs that pipe back files or blobs.

jhancock93: Good point. This was the same code generated for the 200 case, but I guess that was for a valid model instead of a stream. I'll investigate an improved fix.

jhancock93:
I've modified the original change to only read non-modeled error responses as a string in the CSharp client. So, if the default response specifies a stream model, that will work as before. I think it would be unusual for an error to return a stream, but at least this continues to handle that as before. I have also added acceptance tests to ensure that I can get the returned response content.

jhancock93:
Because of the merge conflict, I can open a new pull request with code rebased on the master branch if that is preferable.

amarzavery:
@jhancock93 - pulling from upstream master to resolve conflicts and then pushing the changes should be fine.

jhancock93:
Rebased against master and squashed commits to clean up history.

amarzavery:

Issue #719, opened 2016-02-22
Create *nix build for AutoRest and Docker container


stankovski:
- [x] Add bash script to check for mono and run AutoRest.exe
- [x] Create an automation for windows and unix builds
- [x] Create Docker file for AutoRest

PR #760, opened 2016-02-29
Added dockerfile for AutoRest execution

stankovski:
Implements #719


devigned:
LGTM. Though, it would be nice to have a uri that points to the latests autorest.

PR #780, opened 2016-03-02
CustomURI Python test reformat

annatisch:


stankovski:
@azuresdkci add to whitelist

stankovski:
@annatisch can you run `gulp regenerate:expected` to update generated tests with the change

annatisch:
There are no changes to the generated code - only the test case itself.

annatisch:
I take it back - it looks like they weren't regenerated at some point

PR #792, opened 2016-03-03
fix long running operations in ruby

devigned:
refactor / fix the long running operations


devigned:
@stankovski when these pass, we should be good to go

PR #795, opened 2016-03-04
provide the gem-version information when generating

devigned:
Just provide some guidance on the gem version dependency when gen'ing code.


stankovski:
@devigned please pull from upstream and update the PR

devigned:
@stankovski should be good after a re-run of tests

PR #797, opened 2016-03-04
Escape dots in serialized name if present in the property name

stankovski:


amarzavery:
@shuagarw - This should fix the "odata.nextLink" in UserListResult.

PR #801, opened 2016-03-04
x-ms-client-name implementation

NiklasGustafsson:
This PR implements x-ms-client-name for parameters and model properties.

It also includes a fix to preserve leading underscores, which is not representing a use of underscores as word separators.

I added a boolean property to test whether ClientProperty is non-null, a test that is made in many places. I did not, however, convert any locations to use it.

Added tests


azurecla:
Hi __@NiklasGustafsson__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (niklasg). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

stankovski:
@azuresdkci add to whitelist

AutoRest/AutoRest.Core/ClientModel/Parameter.cs
```diff
@@ -28,6 +28,20 @@ public Parameter()
         public string Name { get; set; }
 
         /// <summary>
+        /// Gets or sets the parameter client (explicitly defined code generation) name.
+        /// </summary>
+        public string ClientName {
```
stankovski: While conceptually this change is OK, it's not in line with our design philosophy (which is not documented :)). We try to not change Core client model for these extensions as much as possible. Instead we implement extensions inside Extensions project. In this case I would have a method inside extensions that would change the parameter.Name and property.Name if x-ms-client-name was specified.

stankovski: With this design, each language will be able to opt-in the extensions if appropriate.

NiklasGustafsson: OK, I'll adjust it, but I'd appreciate your advice.

There's a central location to change the name of parameters in Autorest.Core.CodeNamer, which is why I placed ClientName on the model itself. If it should be in Extensions, we would essentially have to lift the logic to determine parameter names out of CodeNamer.cs, where it depends on the virtual method GetParameterName(). We would need to add a virtual method for the parameter naming in CodeNamer, and then pass the language-specific code namer in.

What if the CodeNamer code called the virtual parameter normalizing method? The choice would be language-specific, but not involve Extensions, which seems like a detour.

NiklasGustafsson: Do you mean the 'Extensions.cs' file in the Autorest.Generator.Extensions project, or the various ClientModelExtension.cs files among the code generators?

stankovski:
Once you finish implementation, please add tests and run `gulp regenerate:expected` to update generated code.

amarzavery:
Please update the documentation for this extension over [here](https://github.com/Azure/autorest/blob/master/Documentation/swagger-extensions.md) as a part of this PR.

NiklasGustafsson:
Is there any more information you need from me on this one?

Documentation/swagger-extensions.md
```diff
@@ -1,487 +1,535 @@
...
+
+## Generic Extensions
+* [x-ms-code-generation-settings](#x-ms-code-generation-settings) - enables passing code generation settings via swagger document
+* [x-ms-skip-url-encoding](#x-ms-skip-url-encoding) - skips URL encoding for path and query parameters
+* [x-ms-enum](#x-ms-enum) - additional metadata for enums
+* [x-ms-parameter-grouping](#x-ms-parameter-grouping) - groups method parameters in generated clients
+* [x-ms-paths](#x-ms-paths) - alternative to [Paths Object](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#pathsObject) that allows [Path Item Object](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#pathItemObject) to have query parameters for non pure REST APIs
+* [x-ms-client-name](#x-ms-external) - allows control over identifier names used in client-side code generation for parameters and schema properties.
```
amarzavery: (#x-ms-client-name)

NiklasGustafsson:
Amar,

Thanks for catching that -- it's been fixed.

amarzavery:
Can you add this extension on one of the swagger files over [here](https://github.com/Azure/autorest/tree/master/AutoRest/TestServer/swagger) or add a new swagger file with the "x-ms-client-name" extension for global/local parameters and a property of the model schema. 
It would be even better if this is used in conjunction with "x-ms-client-flatten" extension to make sure that the generated code serializes/deserializes correctly in all the languages.
A simple test in every language
and the corresponding scenario on the test server over [here](https://github.com/Azure/autorest/tree/master/AutoRest/TestServer/server). 

This would help us if we regress on this in future.

NiklasGustafsson:
Amar and I met and discussed. The new implementation tests client name in conjunction with flattening, both for parameters and models.

PR #828, opened 2016-03-11
optional single enum will not be treated as a constant

amarzavery:


AutoRest/Generators/Extensions/Extensions.Tests/ExtensionsTests.cs
```diff
@@ -216,8 +216,8 @@ public void TestClientNameJavaNormalization()
 
             var type = clientModel.ModelTypes.First();
 
-            Assert.Equal("code", type.Properties[0].Name);
-            Assert.Equal("message", type.Properties[1].Name);
+            Assert.Equal("errorCode", type.Properties[0].Name);
+            Assert.Equal("errorMessage", type.Properties[1].Name);
             Assert.Equal("parentError", type.Properties[2].Name);
```
amarzavery: @NiklasGustafsson - I had to make these changes for the x-ms-client-name extension tests for java and python. The tests were expecting wire format instead of the client name. Hence modified them appropriately. Let me know what you think ?

Issue #825, opened 2016-03-09
useDateTimeOffset only works with regular CSharp Generation, not AzureCSharp Generation


begoldsm:
After resolution of issue: https://github.com/Azure/autorest/issues/500 we can properly generate clients in C# to use DateTimeOffset, however the same is not true for AzureCSharperGenerator. Based on my  investigation the problem is that the property is set properly only for regular CSharpCodeGenerator, not the AzureCSharpCodeGenerator. It looks like the reason for this is that the CSharp settings are not also being propagated to the AzureCSharp settings, so the _namer in AzureCSharp will always have the value set to false.

I am not sure how to resolve this issue @stankovski can you take a look?

devigned:
Thank you, @begoldsm. @stankovski might be delayed in responding. I'll see if we can get some other eyes on this.

begoldsm:
@devigned thanks! Let me know who is looking at it and I can help them out, since I know where in the code the problem is, I am just not quite sure how to fix it/get the settings propagated :)

begoldsm:
@devigned I think I have a fix for this now, but I would like it looked at to make sure it fits within the AutoRest design.

devigned:
@sphibbs would you direct this to the right folks?

begoldsm:
This issue is now resolved by the merged change above. I am closing this out now as all new builds of AutoRest nightly will have this change.

PR #841, opened 2016-03-14
Change to the Azure code generator to pass settings

begoldsm:
This enables UseDateTimeOffset to work properly for Azure generated C#
clients which is part of issue: https://github.com/Azure/autorest/issues/825

### DIFF
--- a/AutoRest/Generators/CSharp/Azure.CSharp/AzureCSharpCodeGenerator.cs
+++ b/AutoRest/Generators/CSharp/Azure.CSharp/AzureCSharpCodeGenerator.cs
@@ -25,7 +25,8 @@ namespace Microsoft.Rest.Generator.CSharp.Azure
 
         public AzureCSharpCodeGenerator(Settings settings) : base(settings)
         {
-            _namer = new AzureCSharpCodeNamer();
+
+            _namer = new AzureCSharpCodeNamer(settings);
             IsSingleFileGenerationSupported = true;
             pageClasses = new Dictionary<KeyValuePair<string, string>, string>();
         }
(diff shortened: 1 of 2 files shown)


AutoRest/Generators/CSharp/Azure.CSharp/AzureCSharpCodeGenerator.cs
```diff
@@ -25,7 +25,8 @@ public class AzureCSharpCodeGenerator : CSharpCodeGenerator
 
         public AzureCSharpCodeGenerator(Settings settings) : base(settings)
         {
-            _namer = new AzureCSharpCodeNamer();
+
```
amarzavery: Why an empty line

begoldsm: Fixed!

AutoRest/Generators/CSharp/Azure.CSharp/AzureCSharpCodeNamer.cs
```diff
@@ -24,6 +24,28 @@ public class AzureCSharpCodeNamer : CSharpCodeNamer
             // Do nothing
         }
 
+        public AzureCSharpCodeNamer(Settings settings)
+        {
+            if (settings != null && settings.CustomSettings != null)
```
brodyberg: Is this an error condition for your object? Is it interesting to log about whether or not AzureCSharpCodeNamer got a valid settings object and/or took action on the CustomSetting?

begoldsm: This is to fix the bug where the base class' property: UseDateTimeOffset is not being set when the AzureCSharpCodeNamer is instantiated. We don't care if settings are passed in, so this constructor now behaves exactly like the default constructor if settings is null, however if settings is not null we will attempt to properly set this value.

jianghaolu:
Hey @azuresdkci retest this please

PR #844, opened 2016-03-14
Support const values and add overloads without optional parameters

jianghaolu:
Addresses issue https://github.com/Azure/azure-sdk-for-java/issues/614.


begoldsm:
Looks good to me, can we get this merged? I am very interested in consuming it :)

Issue #847, opened 2016-03-15
Parameter flattening doesn't correctly handle "required" parameters


matthchr:
A quick summary of the issue is that AutoRest treats flattened `required: true` parameters the same as it treats flattened `required: false` parameters, but it shouldn't because the expected behavior is different. A `required: true` parameter should _never_ be omitted from the API call.

For example if I have a parameter like so: 
```
          {
            "name": "taskUpdateParameter",
            "in": "body",
            "required": true,
            "schema": {
              "$ref": "#/definitions/TaskUpdateParameter"
            },
            "description": "The parameters for the request."
          },
```

where `taskUpdateParameter` looks like this:
```
"TaskUpdateParameter": {
      "properties": {
        "constraints": {
          "$ref": "#/definitions/TaskConstraints",
          "description": "Sets constraints that apply to this task. If omitted, the task is given the default constraints."
        }
      },
      "description": "Parameters for a CloudTaskOperations.Update request."
    },
```

And I use `-ft 1` I get a method with the following signature like so: 
```
public async Task<AzureOperationHeaderResponse<TaskUpdateHeaders>> UpdateWithHttpMessagesAsync(string jobId, string taskId, TaskConstraints constraints = default(TaskConstraints), TaskUpdateOptions taskUpdateOptions = default(TaskUpdateOptions), Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken))
```

The problem is, `TaskUpdateParameter` is `required`, and the C# code that generates this looks like so:
```
TaskUpdateParameter taskUpdateParameter = default(TaskUpdateParameter);
if (constraints != null)
{
    taskUpdateParameter = new TaskUpdateParameter();
    taskUpdateParameter.Constraints = constraints;
}
```

But what can happen is that the `constraints` passed in is `null`, which means that `taskUpdateParameter` is `null`, which means that a required field was omitted from the request.

The C# code should be:
```
TaskUpdateParameter taskUpdateParameter = new TaskUpdateParameter();
taskUpdateParameter.Constraints = constraints;
```

This impacts at least C# and Java, it may impact other languages as well I am not sure.

amarzavery:
IMO, constraints should be a required property of TaskUpdateParameter. It seems silly, that the parent object with 1 property is required and that property is optional.

matthchr:
@amarzavery The issue is that this is for an "Update" call.  To clear the constraints, you must send a body with no constraints (i.e. `{ }`)

[...]

PR #849, opened 2016-03-16
Better Python handling for issue #847

annatisch:
Better handling for optional body parameters that have been flattened.


annatisch:
Oops, bad merge....

PR #851, opened 2016-03-16
Lazy initialized paging results

jianghaolu:


AutoRest/Generators/Java/Azure.Java.Tests/src/main/java/fixtures/paging/PagingOperationsImpl.java
```diff
@@ -433,14 +439,15 @@ public void onResponse(Call<ResponseBody> call, Response<ResponseBody> response)
...
-            PagingGetMultiplePagesWithOffsetNextOptions pagingGetMultiplePagesWithOffsetNextOptions = new PagingGetMultiplePagesWithOffsetNextOptions();
-            pagingGetMultiplePagesWithOffsetNextOptions.setMaxresults(pagingGetMultiplePagesWithOffsetOptions.getMaxresults());
-            pagingGetMultiplePagesWithOffsetNextOptions.setTimeout(pagingGetMultiplePagesWithOffsetOptions.getTimeout());
-            response = getMultiplePagesWithOffsetNext(response.getBody().getNextPageLink(), clientRequestId, pagingGetMultiplePagesWithOffsetNextOptions);
-            result.addAll(response.getBody().getItems());
-        }
+        PagedList<Product> result = new PagedList<Product>(response.getBody()) {
+            @Override
+            public Page<Product> loadPage(String nextPageLink) throws CloudException, IOException {
+                PagingGetMultiplePagesWithOffsetNextOptions pagingGetMultiplePagesWithOffsetNextOptions = new PagingGetMultiplePagesWithOffsetNextOptions();
```
jianghaolu: These 4 lines are called redundantly every time but for now there is no good workaround.
Moving this block out would require `pagingGetMultiplePagesWithOffsetNextOptions` to be declared final but when it's optional we would have something like
```java
final PagingGetMultiplePagesWithOffsetNextOptions pagingGetMultiplePagesWithOffsetNextOptions = null;
if (maxresults != null) {
    pagingGetMultiplePagesWithOffsetNextOptions = new PagingGetMultiplePagesWithOffsetNextOptions();
}
```
which won't compile.

ClientRuntimes/Java/azure-client-runtime/src/main/java/com/microsoft/azure/PagedList.java
```diff
@@ -0,0 +1,277 @@
...
+    /**
+     * Keep loading the next page from the next page link until all items are loaded.
+     */
+    public void loadAll() {
+        while (nextPageLink != null) {
+            try {
+                Page<E> nextPage = loadPage(nextPageLink);
+                nextPageLink = nextPage.getNextPageLink();
+                addAll(nextPage.getItems());
+            } catch (CloudException e) {
```
jianghaolu: A [RuntimeException](https://docs.oracle.com/javase/7/docs/api/java/lang/RuntimeException.html) is required here so I have to wrap them up.

ClientRuntimes/Java/azure-client-runtime/src/main/java/com/microsoft/azure/PagedList.java
```diff
@@ -0,0 +1,245 @@
...
+ */
+public abstract class PagedList<E> implements List<E> {
+    private List<E> items;
+    private int pageCount;
+    private String nextPageLink;
+
+    public PagedList(Page<E> page) {
+        items = page.getItems();
+        nextPageLink = page.getNextPageLink();
+        pageCount = 1;
```
anuchandy: Are we planning to use 'pageCount' member to implement some functionalities ?

ClientRuntimes/Java/azure-client-runtime/src/main/java/com/microsoft/azure/PagedList.java
```diff
@@ -0,0 +1,245 @@
...
+    @Override
+    public boolean retainAll(Collection<?> c) {
+        return items.retainAll(c);
+    }
+
+    @Override
+    public void clear() {
+        items.clear();
+    }
+
```
anuchandy: All the below methods that access "items" expect that user already invoked "loadAll", right? 

for example, if there is only first page fetched which has 10 items and if user invoke get(12) then he will receive indexOutOfBounds exception.

In that case may be we want to document this behaviour or mark these methods as not implemented.

ClientRuntimes/Java/azure-client-runtime/src/main/java/com/microsoft/azure/PagedList.java
```diff
@@ -0,0 +1,245 @@
...
+    public <T> T[] toArray(T[] a) {
+        loadAll();
+        return items.toArray(a);
+    }
+
+    @Override
+    public boolean add(E e) {
+        return items.add(e);
+    }
+
```
anuchandy: All the below methods that access "items" expect that user already invoked "loadAll", right?

for example, if there is only first page fetched which has 10 items and if user invoke get(12) then he will receive indexOutOfBounds exception.

In that case may be we want to document this behaviour or mark these methods as not implemented.

ClientRuntimes/Java/azure-client-runtime/src/main/java/com/microsoft/azure/PagedList.java
jianghaolu: That is removed.

ClientRuntimes/Java/azure-client-runtime/src/main/java/com/microsoft/azure/PagedList.java
jianghaolu: get(12) will load to where exactly it's needed.

add() doesn't. This is debatable. I can't think of a scenario users want to modify the returned list() result.

jianghaolu: See test here: https://github.com/Azure/autorest/pull/851/files#diff-b4935f05ef8ab8fbfdff4fd53cad6eb4.

anuchandy: Fine, thanks for confirming get behavior. - you are right - modifying returned list is not a common scenario for us. 

LGTM

ClientRuntimes/Java/azure-client-runtime/src/main/java/com/microsoft/azure/PagedList.java
anuchandy: There are couple of places (e.g. get, loadAll, next) we use the below pattern:

```
Page<E> nextPage = loadPage(nextPageLink);
nextPageLink = nextPage.getNextPageLink();
addAll(nextPage.getItems());
```

This is likely to grow.

What do you think about making loadPage responsible for setting nextPageLink and adding the result to items (may be function renamed to loadNextPage())?

In addition to that defining a getter for nextPageLink for example : hasMorePage(). 

With these changes we don't have to pass nextPageLink to loadNextPage(). RW access to nextPageLink is wrapped in loadNextPage and hasMorePage()

With these changes our get looks like:

```
while (index >= items.Size() && hasMorePage()) {
  loadNextPage()
}  catch (CloudException e) {
 ...
} catch (IOException e) {
 ....
}

```

The loadAll will become:

```
public void loadAll() {
  while (hasMorePage()) {
    loadNextPage()
  }  catch (CloudException e) {
 ...
  } catch (IOException e) {
   ....
  }
}
```

Similarly next().

ClientRuntimes/Java/azure-client-runtime/src/main/java/com/microsoft/azure/PagedList.java
```diff
@@ -0,0 +1,291 @@
...
+    @Override
+    public boolean isEmpty() {
+        return items.isEmpty() && nextPageLink == null;
+    }
+
+    @Override
+    public boolean contains(Object o) {
+        loadAll();
+        return items.contains(o);
```
anuchandy: Do we need to invoke loadAll() here. We could load page by page - check for the item in each page and return as soon as we find it.

ClientRuntimes/Java/azure-client-runtime/src/main/java/com/microsoft/azure/PagedList.java
jianghaolu: > What do you think about making loadPage responsible for setting nextPageLink and adding the result to items?

This will make our code clean but whoever is extending `PagedList` won't have much idea of that's what they need to do. With the current model it's clearer the implementation class needs to know how to get a `Page` from a `nextPageLink` simply by looking at the method signature of `loadPage()`.

Your `hasNextPage()` reminds of iterators, maybe something like this?
```java
// user override
public abstract Page<E> nextPage(Sting nextPageLink) throws CloudException, IOException;

public Page<E> loadNextPage() {
    try {
        return loadPage(this.nextPageLink);
    } catch (CloudException) {
    } catch (IOException) {
    }
}

public boolean hasNextPage() {
    // check if there is next page
}
```

ClientRuntimes/Java/azure-client-runtime/src/main/java/com/microsoft/azure/PagedList.java
jianghaolu: Good stuff.

ClientRuntimes/Java/azure-client-runtime/src/main/java/com/microsoft/azure/PagedList.java
anuchandy: May be we could have a private method "loadNextPage" in PagedList, that wraps the user's implementation of abstract method "nextPage" along with setting nextPageLink and appending result to items

I mean something like below:

```
public abstract class PagedList<E> implements List<E> {
  ....
  // user override
   public abstract Page<E> nextPage(Sting nextPageLink) throws CloudException, IOException;

   private bool hasNextPage() {
     return nextPageLink != null
   }

   private void loadNextPage() {
     Page<E> nextPage = loadPage(nextPageLink);
     nextPageLink = nextPage.getNextPageLink();
     addAll(nextPage.getItems());
   }

   public E get(int index) {
     while (index >= items.Size() && hasMorePage()) {
       try {
         loadNextPage()
       }  catch (CloudException e) {
       ...
       } catch (IOException e) {
       ...
       }
     }  
     return items.get(index)
  }
   
  public void loadAll() {
    while (hasMorePage()) {
      try { 
        loadNextPage()
      }  catch (CloudException e) {
      ...
      } catch (IOException e) {
       ....
      }
    }
  }
  ...
}

PR #852, opened 2016-03-17
Ruby CG -- passing parameter constants

NiklasGustafsson:
This implements support for passing constants as operation parameters and suppressing them in the method signature corresponding to the operation. Several tests turned out to already be using constant parameters, so the feature is being tested and some tests had to be rewritten to no longer pass in values since there is no need.

Also uncovered a bug in the handling of how byte arrays were converted to strings, which was being masked by a test before the change was made.


NiklasGustafsson:
There were tests that needed to be re-written that I missed. I will reopen a new PR after everything is OK again.

devigned:
LGTM!

PR #858, opened 2016-03-18
Ruby CG and Swagger JSON schema

NiklasGustafsson:
1) Implementing synchronous and HTTP-response-free operation methods for Ruby. This will lead to silent breaking changes in any generated libraries as the type returned will have changed for all non-LRO operations. Discussed with David Justice and we concluded that it was better to take the breaking change now than later.
2) Modified tests to use the '_async' methods, except for LRO tests.
3) Added 'x-ms-client-name' to Swagger extensions schema.

### DIFF
--- a/schema/swagger-extensions.json
+++ b/schema/swagger-extensions.json
@@ -538,6 +538,10 @@
         "x-ms-parameter-grouping": {
           "$ref": "#/definitions/xmsParameterGrouping"
         },
+        "x-ms-client-name": {
+          "type": "string",
+          "description": "A code-generator-friendly name for the parameter."
+        },
         "schema": {
           "$ref": "#/definitions/schema"
         }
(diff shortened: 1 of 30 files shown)


NiklasGustafsson:
If someone could code review this, I'll merge it when the checks complete. The schema file changes, which there are no tests for, are especially of interest in terms of CR.

schema/swagger-extensions.json
```diff
@@ -1122,6 +1142,9 @@
         "example": { },
         "x-ms-client-flatten": {
           "$ref": "#/definitions/xmsClientFlatten"
+        },
+        "x-ms-client-name": {
+          "type": "string"
```
amarzavery: overall the addition of this extension definition looks good to me. Just curious to know, if there was a special reason to not have description here v/s having it in other places?

NiklasGustafsson: I took it out because referring to 'parameter' was wrong and I didn't think of the right term to use here -- 'element,' 'property,' etc.

devigned: I think we need to be careful when describing extensions with a root of anything other than Object. When we define extensions as primitives (bool, string, number), then we lock ourselves into that structure. The soft nature of objects and keys will allow for extensibility and adaptation later as the extension matures.

Thoughts?

PR #860, opened 2016-03-18
fixes #782 for node.js and JSON.stringify of null value during serialization

### LINKED ISSUE #782: Change constant behavior to not initialize parent of child is a constant
Currently in the below example if `zoo` is a constant `bar` will be initialized in `foo` object. It should not be. `bar` should only be initialized if `bar` is required.

``` 
   foo
       bar
          zoo
```

amarzavery:
-   fixes #782 for node.js and JSON.stringify of null value during serialization

### DIFF
--- a/AutoRest/Generators/NodeJS/NodeJS/Templates/MethodTemplate.cshtml
+++ b/AutoRest/Generators/NodeJS/NodeJS/Templates/MethodTemplate.cshtml
@@ -149,8 +149,8 @@
     if (@(Model.RequestBody.Name) !== null && @(Model.RequestBody.Name) !== undefined) {
       @(Model.ConstructRequestBodyMapper)
       requestModel = client.serialize(requestModelMapper, @(Model.RequestBody.Name), '@(Model.RequestBody.Name)');
+      requestContent = JSON.stringify(requestModel);
     }
-    requestContent = JSON.stringify(requestModel);
   } catch (error) {
     var serializationError = new Error(util.format('Error "%s" occurred in serializing the ' + 
         'payload - "%s"', error.message, util.inspect(@(Model.RequestBody.Name), {depth: null})));
--- a/AutoRest/TestServer/server/routes/string.js
+++ b/AutoRest/TestServer/server/routes/string.js
@@ -4,68 +4,68 @@ var util = require('util');
 var constants = require('../util/constants');
 var utils = require('../util/utils')
 
-var string = function(coverage) {
-    router.put('/:scenario', function(req, res, next) {
-        if (req.params.scenario === 'null') {
-            if (req.body !== null) {
-                utils.send400(res, next, "Did not like null req '" + util.inspect(req) + "'");
-            } else {
-                coverage['putStringNull']++;
-                res.status(200).end();
-            }
-        } else if (req.params.scenario === 'empty') {
-            if (req.body !== '') {
-                utils.send400(res, next, "Did not like empty req '" + util.inspect(req.body) + "'");
-            } else {
-                coverage['putStringEmpty']++;
-                res.status(200).end();
-            }
-        } else if (req.params.scenario === 'mbcs') {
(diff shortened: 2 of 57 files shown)


AutoRest/Generators/NodeJS/NodeJS/Templates/MethodTemplate.cshtml
```diff
@@ -149,8 +149,8 @@
     if (@(Model.RequestBody.Name) !== null && @(Model.RequestBody.Name) !== undefined) {
       @(Model.ConstructRequestBodyMapper)
       requestModel = client.serialize(requestModelMapper, @(Model.RequestBody.Name), '@(Model.RequestBody.Name)');
+      requestContent = JSON.stringify(requestModel);
```
amarzavery: @xingwu1, @begoldsm - This should fix the stringification of null value.

ClientRuntimes/NodeJS/ms-rest/lib/serialization.js
```diff
@@ -266,8 +266,8 @@ function serializeCompositeType(mapper, object, objectName) {
           continue;
         }
         //serialize the property if it is present in the provided object instance
-        if ((modelProps[key].defaultValue !== null && modelProps[key].defaultValue !== undefined) || 
-          (object[key] !== null && object[key] !== undefined)) {
+        if (((parentObject !== null && parentObject !== undefined) && (modelProps[key].defaultValue !== null && modelProps[key].defaultValue !== undefined)) || 
+            (object[key] !== null && object[key] !== undefined)) {
           var propertyObjectName = objectName + '.' + modelProps[key].serializedName;
```
amarzavery: @yaxia - This should fix issue #782

amarzavery: @stankovski - Can you verify this node.js fix for issue #782?
So, if the parentObject's mapper states that it is optional and if it is not provided in the actual object to be serialized then the parentObject is undefined. In that case we will not serialize the parentObject. Previously we used to satisfy the if condition and then the serialization would fail on line 274 where it tries to assign the serialized constant value to its **undefined** parentObject. Now, this will not happen.

Just trying to give you enough context so you can make sure, I don't miss any edge case.

ClientRuntimes/NodeJS/ms-rest/lib/serialization.js
```diff
@@ -60,7 +60,7 @@ exports.serialize = function (mapper, object, objectName) {
   if (mapperType.match(/^Sequence$/ig) !== null) payload = [];
   //Throw if required and object is null or undefined
   if (mapper.required && (object === null || object === undefined) && !mapper.isConstant) {
-    throw new Error(util.format('\'%s\' cannot be null or undefined.', objectName));
+    throw new Error(util.format('\'%s\' cannot be null or undefined.'), objectName);
```
yaxia: I guess this is a merge issue which overwrites my PR https://github.com/Azure/autorest/pull/853.

amarzavery: Thanks for bringing this to notice.

AutoRest/TestServer/server/routes/string.js
```diff
@@ -4,68 +4,68 @@ var util = require('util');
...
-            coverage['getStringWithLeadingAndTrailingWhitespace']++;
-            res.status(200).end('\"    Now is the time for all good men to come to the aid of their country    \"');
-        } else {
-            res.status(400).end('Request path must contain null or empty or mbcs or whitespace');
-        }
-
+var string = function (coverage) {
+  router.put('/:scenario', function (req, res, next) {
+    if (req.params.scenario === 'null') {
+      if (!req.body || (req.body && Object.keys(req.body).length === 0 && req.headers['content-length'] === '0')) {
```
amarzavery: This is the fix on the test server. It converts null value to an empty object. Hence, most of the teams implemented stringified version of null.

PR #864, opened 2016-03-20
Change file upload type to java.io.File

jianghaolu:
So that OkHttp will stream directly from file stream to socket stream.

There seems to be a significant performance downturn when the file hits > 512k. Running against some other online services confirmed it's an issue for our Express server. Will need run against Azure to see how well it really performs.


jianghaolu:
Hmm... pulled from master but still failing on python tests...

jianghaolu:
False alarm.

PR #868, opened 2016-03-21
[node.js] Add responseBody as one of the parameters in the callback while processing stream response

amarzavery:
- [node.js] Add responseBody as one of the parameters in the callback while processing stream response


AutoRest/Generators/NodeJS/NodeJS.Tests/Expected/AcceptanceTests/BodyFile/operations/files.js
```diff
@@ -82,7 +82,7 @@ Files.prototype.getFile = function (options, callback) {
   httpRequest.body = null;
   // Send Request
   httpRequest.streamedResponse = true;
-  return client.pipeline(httpRequest, function (err, response) {
+  return client.pipeline(httpRequest, function (err, response, responseBody) {
```
amarzavery: the third parameter will mostly be undefined for streaming scenario, unless an erroneous response is sent (4xx, 5xx) by the server.

PR #869, opened 2016-03-21
Fix out of memory exception streaming large streams

matthchr:
- Added tests in C#, Java, Python, and NodeJS to test support for streaming
    large streams.
  - Fixed the bug in C# that caused an out of memory exception when a large
    stream was streamed.


AutoRest/Generators/CSharp/CSharp.Tests/AcceptanceTests.cs
```diff
@@ -301,21 +301,32 @@ public void FileTests()
                 SwaggerPath("body-file.json"), ExpectedPath("BodyFile"));
             using (var client = new AutoRestSwaggerBATFileService(Fixture.Uri))
             {
-                var stream = client.Files.GetFile();
-                Assert.NotEqual(0, stream.Length);
-                byte[] buffer = new byte[16 * 1024];
+                using (var stream = client.Files.GetFile())
```
matthchr: Please make note of this change... there are a few things:
1) Clients should always using() this stream (they should've before too, although these tests weren't doing it)
2) The stream is now _not_ a `MemoryStream` it is a `NetworkStream`, so it is not seekable, and you cannot query its length.  This is a behavior change from before when you could do both of those things.  With that said, I think it's the right change (clients should be forced to opt in to doing what they want with the stream -- and if all they want is the data they won't even notice).

amarzavery:
@matthchr - I just merged #868. Can you please pull from upstream master, build and regenerate. There was a bug in the streaming response scenario for node.js that i fixed in #868.

matthchr:
@amarzavery Done

amarzavery:
LGTM

PR #873, opened 2016-03-22
Changes to Python Azure LROs

annatisch:


annatisch:
Added an additional header to the Python client runtime to resolve issue #867

PR #874, opened 2016-03-22
Fixes for types in parameter mappings

jianghaolu:
- DateTimeRfc1123
- InputStream
- Fix `PagedList` exception type


jianghaolu:
@xingwu1 @begoldsm 
These get the issue fixed but kinda ugly. Once I get the Java typing logic implemented, this should get more stable.

PR #875, opened 2016-03-22
[do not merge] fix async patch and post run

csmengwan:


azurecla:
Hi __@csmengwan__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (qinwa). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

amarzavery:
@yugangw-msft , @markcowl - Can you review the changes for LRO in the client runtime?

ClientRuntimes/NodeJS/ms-rest-azure/lib/azureServiceClient.js
```diff
@@ -233,36 +252,25 @@ AzureServiceClient.prototype._updateStateFromAzureAsyncOperationHeader = functio
...
-      if (result.body.properties && result.body.properties.provisioningState) {
-        pollingState.status = result.body.properties.provisioningState;
-      }
-      else {
-        pollingState.status = LroStates.Succeeded;
-      }
-      
+    } else if (statusCode === 200 ||
+      statusCode === 201 ||
+      statusCode === 204) {
```
yugangw-msft: 204 is an error case for put, sine you are merging "put" and "delete" logic into one method, then this is behavior change.

ClientRuntimes/NodeJS/ms-rest-azure/lib/azureServiceClient.js
```diff
@@ -62,68 +62,73 @@ util.inherits(AzureServiceClient, msRest.ServiceClient);
...
+
+  if (!resultOfInitialRequest.response) {
+    return callback(new Error('Missing resultOfInitialRequest.response'));
+  }
+
   if (resultOfInitialRequest.response.statusCode !== 200 &&
-      resultOfInitialRequest.response.statusCode !== 201 &&
-      resultOfInitialRequest.response.statusCode !== 202) {
-    return callback(new Error(util.format('Unexpected polling status code from long running operation \'%s\'', 
+    resultOfInitialRequest.response.statusCode !== 201 &&
```
yugangw-msft: i know you like to merge 2 methods into one, but doing this we allow status code which would be invalid per operation types, say for "putOrPatch", 204 now becomes a valid value.

yugangw-msft:
I would suggest to first submit a more limited change which address the scenario you are blocked on. Please also undo the pretty format so that we have less trivial change to look at. Yes, the LRO is not very clean, which is caused by too many special cases that services choose not to handle.

ClientRuntimes/NodeJS/ms-rest-azure/lib/azureServiceClient.js
csmengwan: I think 204 check here could just be removed, since 204 should only be seen in the initial request response to indicate no polling needed and should never be expected in update state get call responses. And so as 201. I will make the change.

ClientRuntimes/NodeJS/ms-rest-azure/lib/azureServiceClient.js
csmengwan: I don't think it is good practice to break those into more methods as if that is the case then 201 should also not be an acceptable case for patch. Another example would be in post-delete, that 204 is also not an acceptable statusCode for post either. I suggest adding more checking logic here and if it is a too many lines of code and we should factor the statusCode request method pair check into a helper function.

csmengwan:
I think keep them in separate functions is casing the async patch and post not working issue to be not exposed. So I think we should figure out if merge into one function handles all the cause without breaking existing service so that we can apply this to other languages to avoid future complexity. For the formatting issue you mentioned above, can you send me or link me a JS coding standard to me?

ClientRuntimes/NodeJS/ms-rest-azure/lib/azureServiceClient.js
```diff
@@ -57,73 +57,76 @@ function AzureServiceClient(credentials, options) {
 util.inherits(AzureServiceClient, msRest.ServiceClient);
 
 /**
- * Poll Azure long running PUT operation.
- * @param {object} [resultOfInitialRequest] - Response of the initial request for the long running operation.
+ * Poll Azure long running GET, PATCH, POST or DELETE operations.
```
markcowl: I think you mean PUT here

ClientRuntimes/NodeJS/ms-rest-azure/lib/azureServiceClient.js
```diff
@@ -57,73 +57,76 @@ function AzureServiceClient(credentials, options) {
...
+          return callback(new Error('Location header is missing from long running operation.'));
         }
       }, pollingState.getTimeout());
     },
     //when done
     function (err) {
       if (pollingState.status === LroStates.Succeeded) {
-        if (!pollingState.resource) {
-          self._updateStateFromGetResourceOperation(resourceUrl, pollingState, function (err) {
+        if ((pollingState.azureAsyncOperationHeaderLink || !pollingState.resource) && resultOfInitialRequest.request.method !== 'DELETE' && resultOfInitialRequest.request.method !== 'POST') {
```
markcowl: break line for readability, please

ClientRuntimes/NodeJS/ms-rest-azure/lib/azureServiceClient.js
```diff
@@ -135,78 +138,19 @@ AzureServiceClient.prototype.getPutOrPatchOperationResult = function (resultOfIn
...
- * @param {object} [options]
- * @param {object} [options.customHeaders] headers that will be added to request
- */
-AzureServiceClient.prototype.getPostOrDeleteOperationResult = function (resultOfInitialRequest, options, callback) {
-  var self = this;
-  
-  if (!callback && typeof options === 'function') {
-    callback = options;
-    options = null;
+AzureServiceClient.prototype._checkInitialRequestResponseStatusCodeFailed = function (initialRequest) {
```
markcowl: Consider a simpler name, like checkResponseStatusCodeFailed

ClientRuntimes/NodeJS/ms-rest-azure/lib/azureServiceClient.js
```diff
@@ -57,73 +57,76 @@ function AzureServiceClient(credentials, options) {
...
   
-  if (resultOfInitialRequest.response.statusCode !== 200 &&
-      resultOfInitialRequest.response.statusCode !== 201 &&
-      resultOfInitialRequest.response.statusCode !== 202) {
-    return callback(new Error(util.format('Unexpected polling status code from long running operation \'%s\'', 
-      resultOfInitialRequest.response.statusCode)));
+  if (!resultOfInitialRequest.response) {
+    return callback(new Error('Missing resultOfInitialRequest.response'));
+  }
+
```
markcowl: @csmengwan Inside the "_checkInitialRequestResponseStatusCodeFailed" method, we are using resultOfInitialResponse.request.  We should check that this value is not null before using it, by adding the check here.  Note that if resultOfInitialRequest.request were null before, it would not have caused an error in the old code.

ClientRuntimes/NodeJS/ms-rest-azure/lib/azureServiceClient.js
```diff
@@ -57,73 +57,76 @@ function AzureServiceClient(credentials, options) {
...
+  if (this._checkInitialRequestResponseStatusCodeFailed(resultOfInitialRequest)) {
+    return callback(new Error(util.format('Unexpected polling status code from long running operation \'%s\' for method \'%s\'',
```
markcowl: @csmengwan I see a scenario that worked in the old implementation that looks like it doesn't work here.  What if the initial response to a PUT operation is a status code 200 with ProvisioningState == Failed.  In the old case, this would return an error with provisioning failed.  In the new case this will return 'unexpected status code'.  This looks like it's a result of the way the code was restructured

ClientRuntimes/NodeJS/ms-rest-azure/lib/azureServiceClient.js
markcowl: @csmengwan This would be easier to read if we did two things (1) assign resultOfInitialRequest.request.method to a local variable (called initialRequestMethod, and (2) make these checks positive (initialRequestMethod === 'PUT' || initialRequestMethod === 'PATCH'.  I would aslo suggest just making this request for a PUT or PATCH regardless of which polling method we use - the worst that can happen is an extra Get request.  Additionally, this seems a little unsafe when using the Location header for polling.

csmengwan: (pollingState.azureAsyncOperationHeaderLink || !pollingState.resource) is more loosen than (!pollingState.resource). Do you mean remove (!pollingState.resource) completely?

amarzavery:
@csmengwan - Can you pull from upstream and resolve conflicts.

amarzavery:

Issue #805, opened 2016-03-05
Base64Url is not an AutoRest supported type


pomortaz:
Key Vault REST API for data-plane has data type of Base64Url encoding
e.g. https://msdn.microsoft.com/en-us/library/azure/dn878060.aspx

However, this type is not supported by Swagger. Is it possible to add a Microsoft Azure Extension to allow for supporting Base64Url type? Otherwise, Key Vault data-plane SDK cannot be generated using AutoRest.

stankovski:
It's a bug. Swagger supports base64 formatted string via "byte" format (http://swagger.io/specification)

pomortaz:
Base64Url is very similar to Base64, except that the value encoding for characters 62 and 63 is different as it is explained in https://tools.ietf.org/html/rfc4648#page-7.

Previously, Key Vault data plane SDK, was wrapping the hydra generated codes to enable data type conversion of byte to Base64URL. However, since AutoRest supports Base64, having an additional data type of Base64Url seems possible.
This is how Key Vault is currently doing the [byte to Base64Url conversion](https://github.com/Azure/azure-sdk-for-net/blob/master/src/KeyVault/Microsoft.Azure.KeyVault/WebKey/Json/Base64UrlConverter.cs).

NiklasGustafsson:
Maybe this is best addressed by adopting the convention that if the format is 'byte,' i.e. base64-encoded, and the parameter is a query or path parameter, then it's base64URL instead of standard base64?

markcowl:
Is this always going to occur in the url of the method, or can properties in the body include such an encoding?

pomortaz:
Properties of type Base64Url encoding are in the [body of the requests](https://github.com/pomortaz/azure-sdk-for-net/blob/master/src/KeyVault/Microsoft.Azure.KeyVault/Models/Messages.cs).

PR #878, opened 2016-03-23
Implementing base64url format in C#

tbombach:
This PR addresses issue #805. It adds the new format/primary type, updates the code namers for each language to handle the new primary type, implements the encoder/decoder in C#, and adds tests for C#.


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

AutoRest/Generators/CSharp/CSharp.Tests/Expected/Mirror.Primitives/Models/Product.cs
```diff
@@ -168,6 +176,11 @@ public Product(int? integer = default(int?), int? intProperty = default(int?), l
 
         /// <summary>
         /// </summary>
+        [JsonProperty(PropertyName = "base64urlArray")]
```
stankovski: Why isn't there a `[JsonConverter(typeof(Base64UrlJsonConverter))]` attribute?

AutoRest/Generators/CSharp/CSharp.Tests/Swagger/swagger-mirror-primitives.json
```diff
@@ -185,6 +185,10 @@
                 "string": {
                     "type": "string"
                 },
+                "base64url": {
```
stankovski: It would be nice to have base64url in `path` example in generated code and not just swagger test to make sure the logic for non-body serialization works as well

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/Serialization/Base64UrlJsonConverter.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                return input;
+            }
+
+            return input + new string('=', count);
+        }
+
+        public override bool CanConvert(Type objectType)
+        {
+            if (objectType == typeof(byte[]))
```
stankovski: How about if the type is `IList<byte[]>`?

markcowl: Sequence and dictionary types should both be supported, I think

tbombach: Yep, JsonConverter.DeserializeObject handles IList and dictionary cases correctly - it will just use the ReadJson function on the values of the inner type, not the collection as a whole. I've added two tests for both array and dictionary cases to make verify this is the case in my [latest commit](https://github.com/Azure/autorest/pull/878/commits/3e4df712343c7f3f0614d66133910efb65d319e0) . Thanks!

AutoRest/Generators/CSharp/CSharp.Tests/Expected/Mirror.Primitives/Models/Product.cs
tbombach: I don't think Autorest currently adds the attribute when a property is an array of the type. In the same file, Date property has the [JsonConverter(typeof(DateJsonConverter))] attribute, but DateArray property does not.

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/Serialization/Base64UrlJsonConverter.cs
stankovski: Great! Then we are all set.

stankovski:
@azuresdkci add to whitelist

AutoRest/Generators/CSharp/CSharp.Tests/Expected/Mirror.Primitives/Models/Product.cs
stankovski: Sorry for not responding earlier. That's very odd since there is no other way for json.net to know to use this converter unless it's globally declared (which is not something you want to do in this case)

stankovski:
One more thing. Since there is a change in the runtime, please bump up runtime package version and run the gulp syncDependencies task.

Issue #600, opened 2016-01-11
Web Api Batch support


suddenelfilio:
Hello 

any plans on supporting ASP.NET Multipart content Batch processing?

regards,

Filip

stankovski:
@suddenelfilio please provide an example of the swagger spec

suddenelfilio:
I don't think you should look for this in the swagger spec but rather provide the functionality in the generated client. It is in fact nothing more than aggregating multiple HttpRequestMessages (yes this is C#, but that is my main concern here) into 1 request as explained on the microsoft ASP.NET website: https://aspnetwebstack.codeplex.com/wikipage?title=Web+API+Request+Batching

Or this too specific for this project as it is too tight to the technology stack?

sharadagarwal:
I also would like to see multipart batch support in Autorest. It's not clear to me yet how my Swagger spec would be different, other than perhaps having another API that receives batched requests - the Autorest client would need some way to stuff multiple API calls into the single batch API call using multipart.

stankovski:
@suddenelfilio we do have support for [multipart form data](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameter-object), but I'm not sure if it will work with ASP.NET batching. If you want to propose a design for generated C# clients (as well as a way to determine it from swagger) please add it to the thread and we will consider adding it.

sharadagarwal:
What I'd need is a way to get at the system.net.http.httprequestmessage that is being constructed when I make an API call. For every API call that AutoRest generates a method for, perhaps it can generate another method where it doesn't actually make the API call but rather returns an HttpRequestMessage that is properly filled out. Then I would collect a list of HttpRequestMessages and issue those as a batch request. On receiving a batch response, then I'd need to properly deserialize them, so some support for that would also be useful.

sharadagarwal:
I've issued a pull request for this : #914

fearthecowboy:
Howdy!

In our planning for driving towards a stable '1.0' release, I'm marking this issue as 'deferred' 💤 and we're going to review it during the post-1.0 planning cycle.

It's not to say that we're not going to work on it, or that this isn't not important, but at the moment, we're picking and choosing the stuff we must do before 1.0. 🏇 🏇 🏇

We'll make sure we pick this back up at that point. 🎉

Issue #837, opened 2016-03-14
proposal to support batch requests in AutoRest


sharadagarwal:
## Batch:
Typically, users of REST services issue API calls individually - i.e., each API call is a separate HTTP request. However, some REST services provide a batch interface where users can bundle multiple, separate API calls into a _single_ HTTP call. It is very easy to enable this in your Web API service by following these [instructions](https://blogs.msdn.microsoft.com/webdev/2013/11/01/introducing-batch-support-in-web-api-and-web-api-odata/).

Under this model, your client stuffs multiple API calls into a single HTTP request by making the body of the HTTP request a multipart content, and adding each HttpRequestMessage for each API call as one part of that content, and issuing the HTTP request to the batch URL. Instructions are [here](https://blogs.msdn.microsoft.com/webdev/2013/11/01/introducing-batch-support-in-web-api-and-web-api-odata/).

## Current status in AutoRest:
Clients generated by AutoRest today do not have explicit support for batching. Requests have been made to add batch support: #600

## Proposal:
I have built a DelegatingHandler that provides batch support in AutoRest. When attempting to do batch requests with an AutoRest client, you need to:
- instantiate the BatchHttpMessageHandler class
- add it to an array of DelegatingHandler and pass that into the constructor for your AutoRest client
- call any of your async API calls inside your AutoRest client, just as if you were doing regular non-batch calls
- do not await them yet; once you are done setting up all of your async API calls for this batch, call IssueBatch on your instance of BatchHttpMessageHandler. That will issue all those calls as a single batch, and properly return the appropriate response to all the tasks that were created in the previous step.
- now you can await the tasks from your individual API calls.
- if you wish to issue another batch call, call Reset() on your instance of BatchHttpMessageHandler

There are three tricky parts to building this handler correctly:

1. The handler has to return tasks for each API call that will properly complete only after the batch request is later successfully issued to and returned by the web service.
2. The handler is handling multiple requests and multiple responses. It has to provide the appropriate response to the appropriate request task upon completion.
3. The handler should not get into an async/await deadlock.

I have implemented this in C# and tested it. It works.

## Factors to consider:
- I have this working in C#, but not for any of the other languages supported by AutoRest. I have no idea if the same model would work in the other languages. I have no plans to investigate support for other languages.
[...]

PR #914, opened 2016-04-04
support for Batch API call in C# client

sharadagarwal:
Background information is in issue #837 .

This pull request includes:

- BatchDelegatingHandler : C# clients can use this to pack multiple API calls into a single batch call
- Documentation on how to use it
- Test that verifies it is working, including changes to existing test classes to support this test

Please note that while I've used AutoRest extensively to create clients, I am new to the AutoRest code base. Thanks @devigned and @stankovski for discussions that led up to this pull request.


sharadagarwal:
I did a "git clean -fdx" and built AutoRest.sln, ClientRuntime.sln, and CodeGenerator.sln. All of that worked. It seems that NuGet timed out in the CI build.

amarzavery:
@azuresdkci  Add to whitelist

markcowl:
@sharadagarwal Thanks for the awesome functionality contribution.  I'm curious about the use of the WebAPI client and how / if this limits the target operating systems we support (client versus fullvs. portable skus of .Net)?

Any plans /ideas for how we might support this functionality in other languages?

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+                batchContent.Add(new HttpMessageContent(message));
+            }
+
+            // send the batch request to the inner handler
+            HttpResponseMessage batchResponse = await base.SendAsync(batchRequest, cancellationToken);
+
+            // throw exceptions if the response is unexpected
+            if (batchResponse == null)
+            {
+                throw new Exception("Batch request got back a null response.");
```
markcowl: Exception feels like the wrong exception type - seems like this should be some kind of Http or Web Exception

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+            }
+
+            batchResponse.EnsureSuccessStatusCode();
+
+            if (batchResponse.Content == null)
+            {
+                throw new Exception("Batch request got back a response with null content.");
```
markcowl: Same comment on exception types

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+            }
+
+            if (!batchResponse.Content.IsMimeMultipartContent("mixed"))
+            {
+                throw new Exception("Batch response returned unexpected content type.");
```
markcowl: From here on, you may want to consider aggressively disposing the response before throwing the exception - the response content will hol;d onto a connection with the host until it is read.

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+            }
+
+            // pull out the responses
+            MultipartMemoryStreamProvider responseContents = await batchResponse.Content.ReadAsMultipartAsync();
+            if (responseContents == null)
+            {
+                throw new Exception("Failed to convert response content into multipart components.");
```
markcowl: same comments

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+            }
+
+            if (responseContents.Contents.Count != this.requests.Count)
+            {
+                throw new Exception("Batch response returned " + responseContents.Contents.Count + " number of responses instead of the expected " + this.requests.Count + ".");
```
markcowl: same comments

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        /// <summary>
+        /// Wait for the batch call to finish and then return the requested response.
+        /// </summary>
+        /// <param name="index">Which response to return</param>
+        /// <returns>The HTTP response message inside the batch at the given index</returns>
+        private HttpResponseMessage GetResponse(int index)
+        {
+            this.batchCompleted.WaitOne();
```
markcowl: Should there be a configurable timeout here?

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+    {
+        /// <summary>
+        /// HTTP method (e.g. POST or PUT) to use in issuing a batch call.
+        /// </summary>
+        private readonly HttpMethod batchMethod;
+
+        /// <summary>
+        /// URL to issue the batch call to.
+        /// </summary>
+        private readonly Uri batchURL;
```
markcowl: What are batchUri and batchMethod actually used for?  Are these supposed to be used to filter the incoming requests?  I don't see this happening in the implementation, so curious what the purpose is.

sharadagarwal:
@markcowl Thank you for the code review. 

I believe there is support for portable class libraries in the Microsoft.AspNet.WebApi.Client package: https://www.nuget.org/packages/microsoft.aspnet.webapi.client/ . However, I'm fairly new to AutoRest and mixed SKU projects, so I might need some guidance on what to do here.

It may be possible to build something similar to this for Java: https://developers.google.com/api-client-library/java/google-api-java-client/reference/1.20.0/com/google/api/client/googleapis/batch/BatchRequest#setBatchUrl(com.google.api.client.http.GenericUrl)  but I don't know how that would work with retrofit. 

I don't know enough about the other runtimes in AutoRest. There should be no reason why such functionality couldn't be provided in almost any language. However, providing it in a way that has minimal changes to how the customer calls the generated client is the real challenge. The DelegatingHandler model makes it very easy in C#.

I currently don't have plans to work on support for other runtimes. It's possible that I may need such support in Java in the future for the product that I'm building, but that's not something I can prioritize at this time. Let me know if lack of support for the other runtimes is a blocker for this pull request. If it isn't, then I'll proceed with making the code changes for the rest of your comments.

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+namespace Microsoft.Rest
+{
+    /// <summary>
+    /// HTTP batch handler.
+    /// It expects the caller to dispose HttpRequestMessage and HttpResponseMessage.
+    /// It assumes the web service is supporting batch requests as per this MSDN article:
+    /// https://blogs.msdn.microsoft.com/webdev/2013/11/01/introducing-batch-support-in-web-api-and-web-api-odata/
+    /// i.e. requests are batched into a single request inside the multipart content.
+    /// </summary>
+    public class BatchDelegatingHandler : DelegatingHandler
```
tbombach: I would suggest calling this ODataBatchDelegatingHandler. That way it's clear that this handler does not work when used with a service that implements their batching with a different standard. Additionally, someone else could write a different delegating handler for that standard

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+            }
+
+            // Sanity check. There must not be any queued responses from the server.
+            if (this.responses != null)
+            {
+                throw new Exception("This batch handler instance has been reused without calling reset first.");
+            }
+
+            // Create a request to the batch URL
+            HttpRequestMessage batchRequest = new HttpRequestMessage(this.batchMethod, this.batchURL);
```
tbombach: After discussing with @markcowl, the solution that he came up with would be to determine the batchURL based on the this.requests property (e.g. take the host from the URI of the first request in the collection and append "$batch"). This would accomplish two things:
1. The consumer doesn't need to worry about the batchURL, since it automatically comes from the ServiceClient
2. Make sure the consumer can't specify a batchURL with a host that is different from the host URIs of the individual requests

I don't think it's possible for the requests to have different hosts, simply because of how the requests are created (from the client), but just to be safe, it would be nice to validate that all the requests are being made to that same host.

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+        /// Event that tells all tasks that they can proceed.
+        /// </summary>
+        private ManualResetEvent batchCompleted;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="BatchDelegatingHandler"/> class.
+        /// </summary>
+        /// <param name="method">the type of method that the Batch operation is</param>
+        /// <param name="url">the URL that the Batch operation needs to hit</param>
+        public BatchDelegatingHandler(HttpMethod method, Uri url)
```
tbombach: Since the OData batch standard specifies that the request must be POST and it must be to the "[root]/$batch" endpoint, I think it would be best to make it impossible for users to do something different. The POST verb is easy enough to remove from the constructor arguments, but the batchURL is a little trickier (please see my [comment](https://github.com/Azure/autorest/pull/914/files#r63265593) in `IssueBatch()` in this file)

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+        /// </summary>
+        /// <param name="request">HTTP request message to send to the server</param>
+        /// <param name="cancellationToken">Not used</param>
+        /// <returns>Task that when awaited will provide the HTTP response from the server</returns>
+        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            // Sanity check. There must not be any queued responses from the server.
+            if (this.responses != null)
+            {
+                throw new Exception("This batch handler instance has been reused without calling reset first.");
```
tbombach: I think this should be an InvalidOperationException

tbombach:
For background: [OData 3.0 Batch specification](http://www.odata.org/documentation/odata-version-3-0/batch-processing/)

From a quick glance, 4.0 doesn't look like it changed anything: [OData 4.0 specification](http://docs.oasis-open.org/odata/odata/v4.0/errata02/os/complete/part1-protocol/odata-v4.0-errata02-os-part1-protocol-complete.html#_Toc406398359)

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
```diff
@@ -0,0 +1,185 @@
...
+
+        /// <summary>
+        /// Wait for the batch call to finish and then return the requested response.
+        /// </summary>
+        /// <param name="index">Which response to return</param>
+        /// <returns>The HTTP response message inside the batch at the given index</returns>
+        private HttpResponseMessage GetResponse(int index)
+        {
+            this.batchCompleted.WaitOne();
+            return this.responses[index].ReadAsHttpResponseMessageAsync().Result;
```
tbombach: What happens if the user:

1. Creates some requests (request set A)
2. Calls `IssueBatch()`
3. Calls `Reset()` before awaiting the responses
4. Creates more requests (request set B)
5. Tries to retrieve the response for a request in set A

From what I can tell, this will:
- Retrieve a response for the request in set B that matches the index of the response from set A
- Throw an IndexOutOfRange exception if you try to retrieve the response from a request from set B that has a higher index that the `count()` of set A responses

I'm not sure that this is the behavior, but from looking at the code, this is what I believe will happen.

I don't know of an easy solution for this issue off the top of my head. My instincts would be to create a private class like `BatchContext` that stores the responses and requests each time `IssueBatch()` is called. That way the `requests` collection can be guaranteed to be the same when the responses are read, while allowing more requests to be created.

Documentation/clients-batch.md
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// process response #2
+HttpOperationResponse<SupplierListResponse> suppliersResponse = await getSupplierListTask;
+if (suppliersResponse.IsSuccessStatusCode)
+{
+	Console.WriteLine("Got " + suppliersResponse.Body.Count + " suppliers");
+}
+```
+
```
tbombach: Could you add example code for `Reset()`, that shows the user how to keep creating requests after calling `IssueBatch()`?

tbombach:
@sharadagarwal Thanks for the awesome implementation of request batching! I really like how this is designed - no impact on generated code and it's very easy for the consumer to plug it in to their code.

Since this is based on OData specification, we'd be happy to take this PR as-is with C# support. Especially considering the fact that each language would have a very different implementation, support can be added at a later date for those languages.

Thanks!

sharadagarwal:
@tbombach thanks! I'll start working on all the comments. Thanks for the detailed code review!

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
sharadagarwal: ok. I'll use a HttpRequestException, same type as what EnsureSuccessStatusCode throws.

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
sharadagarwal: ok. I'll use a HttpRequestException, same type as what EnsureSuccessStatusCode throws.

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
sharadagarwal: Thanks, will do.

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
sharadagarwal: ok. I'll use a HttpRequestException, same type as what EnsureSuccessStatusCode throws.

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
sharadagarwal: ok. I'll use a HttpRequestException, same type as what EnsureSuccessStatusCode throws.

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
sharadagarwal: Having a timeout is a good idea. If you want the timeout on individual SendAsync calls, then the developer would have to factor in (1) the time between making individual SendAsync calls and the IssueBatch call; (2) time between the IssueBatch call and the response coming back & being parsed. Is that still useful, or would it make more sense to have the timeout on the IssueBatch call? In the latter case, I can add it as an optional parameter to the IssueBatch call.

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
sharadagarwal: For a service to support incoming batch requests, it has to provide an endpoint for receiving them. That endpoint could be at an arbitrary route on the server. E.g. "/batch" or "/batchfoo". The service would also have to decide what operation issues the batch - should it be a GET or a POST to that endpoint? The batchUri and batchMethod provide both values to the BatchHandler.

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
sharadagarwal: That's an excellent point. Will do.

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
sharadagarwal: Yes, this is a good idea. This works when the service root URI is the same as the hostname - e.g. an individual request is to http://api.contoso.com/Customers , then the batch would be to http://api.contoso.com/$batch . However, when the service root URI contains a path, then this doesn't work : e.g. an individual request is to http://api.contoso.com/service/Customers , then plucking out the hostname and constructing http://api.contoso.com/$batch would be incorrect - it should be http://api.contoso.com/service/$batch . For now, I'll allow the caller to optionally specify the URI for batch, but if you have an idea for doing this more elegantly, please let me know.

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
sharadagarwal: I'll remove the ability to set the HTTP method and make it a POST. However, with the URI, my read of the OData spec indicates that it can include the service root, and getting that out of a requestUri is not trivial. How should we address that?

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
sharadagarwal: thanks, I'll fix this.

ClientRuntimes/CSharp/Microsoft.Rest.ClientRuntime/BatchDelegatingHandler.cs
sharadagarwal: You are correct - that is the behavior that will happen.

One challenge with the private context is that it has to be referred to in SendAsync (the task that is returned has to refer to it). So it would be challenging to create it in IssueBatch and still have the awaited response be able to access it.

Another option would be to remove Reset() and make each instance be one-time-use only. There is very little performance advantage to reusing instances (there's very little state here).

What would you prefer?

Documentation/clients-batch.md
sharadagarwal: I'll add it to clients-batch.md and to the BatchHandlerWorks test method.

sharadagarwal:
@tbombach @markcowl : I've addressed a bunch of your comments. I believe the following items remain:
- configurable timeout on GetResponse
- removing URI passed into the constructor
- preventing second use of instance from trying to access first use's data
Can you please look at my replies to your comments and let me know how you'd like me to proceed?

The "cla-not-required" tag hasn't been applied to this pull request. Is that needed, and if so, how do I get that?

hovsepm:

fearthecowboy:
Is this supposed to fix #600 ?

fearthecowboy:
I'm going to have to close this; this is already a patch against a much older revision; we're about to do another refactoring that makes this even more unlikely to fix. If you still want to pursue this, we'll have to see how it fits into the new model

Issue #776, opened 2016-03-02
long-running-operation without a response body doesn't compile


emgerner-msft:
Long running operations (like below) result in output code that doesn't compile. 

        "responses": {
          "202": {
            "description": ""
          }
        },
        "x-ms-long-running-operation": true

The output error is:

The type arguments for method 'AzureClientExtensions.GetPutOrPatchOperationResultAsync<TBody>(IAzureClient, AzureOperationResponse<TBody>, Dictionary<string, List<string>>, CancellationToken)' cannot be inferred from the usage. Try specifying the type arguments explicitly.

parvezah:
i am facing a similar issue with my spec. is there any work around for this?

amarzavery:
It is in the backlog. We will take this up soon.

devigned:
@parvezah, please open a pull request if it is blocking you. We're happy to review it :).

parvezah:
[https://github.com/Azure/autorest/pull/917](url) Raised a PR. Can you have a look.

PR #917, opened 2016-04-05
A fix for the issue https://github.com/Azure/autorest/issues/776

parvezah:
Adding a method in the AzureClientExtensions to handle the PutOrPatch async calls with 202 responses which have no body and no specific headers


devigned:
@tbombach please give this a review when have a sec. I believe this is blocking @parvezah.

tbombach:
PR looks good. Thanks @parvezah!

PR #922, opened 2016-04-06
Automatically detect MSBuild tools version in Windows platform

daschult:
Any gulp command inside a Windows command prompt fails when only Visual Studio 2015 (VS 14.0) is installed. This is because currently the gulpfile.js assumes that the MSBuild version is 12.0.
This change keeps the default MSBuild tools version to 12.0, but if the user is on Windows, it will default to the highest float-valued folder it can find in C:/Program Files (x86)/MSBuild/. If it can't find any float-valued folders, then it will default back to 12.0.


azurecla:
Hi __@daschult__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (daschult). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

Issue #923, opened 2016-04-06
Allow for defaulted global path parameters


begoldsm:
I have the following ask for all of our clients to support (currently this is already supported in C# and Java from what I can tell).

Given this swagger snippet (note the **bolded **part):

"parameters": {
    "ApiVersionParameter": {
      "name": "api-version",
      "in": "query",
      "required": true,
      "type": "string",
      "description": "Client Api Version."
    },
    **"adlaCatalogDnsSuffixInPath": {
      "name": "adlaCatalogDnsSuffix",
      "in": "path",
      "required": true,
      "type": "string",
      "default": "azuredatalakeanalytics.net",
      "x-ms-skip-url-encoding": true,
      "description": "Gets the DNS suffix used as the base for all Azure Data Lake Analytics Catalog service requests."**
    }
  }

The generated code should not require the user to supply the adlaCatalogDnsSuffix value, since it has a default value, BUT it should check that it is never null/empty when it is being used.

This appears to be a problem specifically in NodeJS right now, and I imagine it will also be a problem in Python, where this value MUST be passed in by the user always, even if they want to use the default value. See the generated snippet below:

function DataLakeAnalyticsJobManagementClient(credentials, **adlaJobDnsSuffix**, options) {
  this.apiVersion = '2016-03-20-preview';
**  this.adlaJobDnsSuffix = 'azuredatalakeanalytics.net';**
  this.acceptLanguage = 'en-US';
  this.longRunningOperationRetryTimeout = 30;
  this.generateClientRequestId = true;
  if (credentials === null || credentials === undefined) {
    throw new Error('\'credentials\' cannot be null.');
  }
 ** if (adlaJobDnsSuffix === null || adlaJobDnsSuffix === undefined) {
    throw new Error('\'adlaJobDnsSuffix\' cannot be null.');
  }**
…

My ask is this: can we remove these hard requirements in client generation for required parameters that have default values?

PR #924, opened 2016-04-06
Fix: https://github.com/Azure/autorest/issues/923

begoldsm:
This should fix issue: https://github.com/Azure/autorest/issues/923


Samples/azure-storage/Azure.Python/setup.py
```diff
@@ -15,7 +15,7 @@
 # prerequisite: setuptools
 # http://pypi.python.org/pypi/setuptools
 
-REQUIRES = ["msrest>=0.1.0", "msrestazure>=0.1.0"]
```
begoldsm: It appears as though whoever checked in python changes didn't re-run regenerate:expected. I am modifying these files as a result of running that after a sync and build.

amarzavery: @lmazuel - Can you take a look at this ?

amarzavery: @annatisch - Can you take a look at this?

annatisch: Yeah I think you're right - I probably just ran regenerate:expected:python/azurepython.
I will take care in future to regenerate the samples as well. Thanks!

PR #929, opened 2016-04-08
Enhancement to "x-ms-parameterized-host" extension

amarzavery:
- Added two more optional properties to "x-ms-parameterized-host" extension
    - useSchemePrefix - It tells AutoRest whether to use the host template as is or to prepend the default scheme while constructing the baseuri. Default value  - true
    - positionInOperation - It tells AutoRest whether to prepend or append the list of parameters specified in the host template to every method. Default value - last. Valid values: first/last.

Important changes are in the following files:  
- AutoRest/Generators/Extensions/Extensions/Extensions.cs
- Documentation/swagger-extensions.md
- schema/swagger-extensions.json
- AutoRest/TestServer/swagger/custom-baseUrl-more-options.json(Test swagger that has a mix of local/global/path/query parameters to ensure that the generated client does the right thing).


NiklasGustafsson:
With positionInOperation -- can I still make the host parameters "global," i.e. occur on the client?

NiklasGustafsson:
Isn't it typically a client parameter, anyway, and when it's not for some reason, why not just make the parameter appear explicitly under each operation? It's repetitive, but less "magic."

AutoRest/TestServer/swagger/custom-baseUrl-more-options.json
```diff
@@ -0,0 +1,109 @@
...
+      {
+        "name": "secret",
+        "description": "Secret value.",
+        "required": true,
+        "type": "string",
+        "in": "path",
+        "x-ms-skip-url-encoding": true
+      },
+      {
+        "$ref": "#/parameters/dnsSuffix"
```
amarzavery: @NiklasGustafsson - This is an example of a reference to global parameter that will be a property on the client.

NiklasGustafsson: You mean the dnsSuffix, right?

amarzavery:
For data plane apis of key vault or datalake, the vault name or the data lake account name is a part of the base uri. "https://{myvault}.vault.azure.net". It is very common for customers to be using multiple vaults and data lake accounts at the same time. It is inconvenient for customers to keep on setting/flipping the  account name if they want to access different accounts. Hence, making them a part of the local method is convenient.

Yup we could make them declare them individually in each operation. However, if you want to make a small change to the definition of the parameter then it has to be made every where and if the parameter is not supposed to be a client property then it needs to be explicitly defined. Having an extension `"x-ms-client-property": true/false` would really help in this case.

AutoRest/TestServer/swagger/custom-baseUrl-more-options.json
amarzavery: yup.

NiklasGustafsson:
I'm wondering about the 'last' value -- isn't there a risk that it interferes with some language's placement of parameters with default values? I'm thinking that the natural thing would be to always have them first in the list, unless they are explicitly included among the parameters of an operation.

AutoRest/Generators/CSharp/CSharp.Tests/Expected/AcceptanceTests/CustomBaseUriMoreOptions/IAutoRestParameterizedCustomHostTestClient.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// <summary>
+        /// The subscription id with value 'test12'.
+        /// </summary>
+        string SubscriptionId { get; set; }
+
+        /// <summary>
+        /// A string value that is used as a global part of the parameterized
+        /// host. Default value 'host'.
+        /// </summary>
+        string DnsSuffix { get; set; }
```
amarzavery: @NiklasGustafsson - As seen here.

AutoRest/Generators/Extensions/Extensions/Extensions.cs
```diff
@@ -89,16 +115,33 @@ public static void ProcessParameterizedHost(ServiceClient serviceClient, Setting
...
+                            hostParamList.Add(parameter);
+                        }
 
-                            foreach (var method in serviceClient.Methods)
+                        foreach (var method in serviceClient.Methods)
+                        {
+                            if (position.Equals("first", StringComparison.OrdinalIgnoreCase))
                             {
-                                method.Parameters.Add(parameter);
+                                method.Parameters.InsertRange(0, hostParamList);
```
amarzavery: @NiklasGustafsson This step happens in the extensions which is before the parameters are sorted into two buckets (required, optional). We preserve the order within these two buckets. So, if the positionInOperation is first, then all the parameters in the host template that will be local parameters in the method, will be added in the beginning of all the required parameters and then we will have all the optional parameters. The x-ms-parameterized-host extension can only have path parameters.
If the position is last then all the required local parameters in the host template will be appended to the list of required parameters for a particular method. All the optional parameters will appear after that.

We can definitely make "first" as the default position. The user can explicitly chose to add those parameters in the end.

AutoRest/Generators/CSharp/CSharp.Tests/Expected/AcceptanceTests/CustomBaseUriMoreOptions/Paths.cs
```diff
@@ -0,0 +1,203 @@
...
+        /// Headers that will be added to request.
+        /// </param>
+        /// <param name='cancellationToken'>
+        /// The cancellation token.
+        /// </param>
+        /// <return>
+        /// A response object containing the response body and response headers.
+        /// </return>
+        public async Task<HttpOperationResponse> GetEmptyWithHttpMessagesAsync(string vault, string secret, string keyName, string keyVersion = "v1", Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken))
+        {
```
amarzavery: @NiklasGustafsson - As you can see, vault and secret were added before the required parameter keyName.

When I changed the positionInOperation to "last" and locally generated it, the method signature looked like this:
```cs
public async Task<HttpOperationResponse> GetEmptyWithHttpMessagesAsync(string keyName, string vault, string secret, string keyVersion = "v1", Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken))
```
As you can see, vault and secret were added after the required local parameter keyName and all the optional parameters started after that.

NiklasGustafsson: OK, good, I now understand why the feature was designed this way. I definitely would make 'first' a default.

amarzavery: I agree, will make 'first', the default behavior

amarzavery:
@begoldsm - I am adding more properties to the "x-ms-parameterized-host" extension. Please review the changes.

Issue #909, opened 2016-03-31
IntDate is not supported in AutoRest


pomortaz:
IntDate is integer representation of UNIX origin time used in JSON Web Token(JWT). Key Vault REST API has an attribute class containing properties of type IntDate such as [exp and nbf](https://msdn.microsoft.com/en-us/library/azure/dn903634.aspx). However, IntDate is not supported in the AutoRest to convert the Unix time to DateTime which is consumable by the client side code. Is it possible to add IntDate support to AutoRest? This is similar to [this issue](https://github.com/Azure/autorest/issues/805).

To do so, we need to define a UnixTimeConverter that [converts type of long to DateTime](https://github.com/Azure/azure-sdk-for-net/blob/master/src/KeyVault/Microsoft.Azure.KeyVault/UnixEpoch.cs). 
[JsonConverter(typeof (UnixTimeConverter))]

markcowl:
@pomortaz Is there a specification somewhere for exactly what intDate is?  The JsonConverter is nice, but we will need to implement this in multiple languages, so a specification of the format would be useful

pomortaz:
According to [JSON Web Token (JWT) draft](https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32), InDate (which is called [NumericDate now](http://static.javadoc.io/org.bitbucket.b_c/jose4j/0.4.4/org/jose4j/jwt/IntDate.html)) is "A JSON numeric value representing the number of seconds from 1970-01-01T00:00:00Z UTC until the specified UTC date/time, ignoring leap seconds."

This is also known as [Unix Time](https://en.wikipedia.org/wiki/Unix_time).

PR #948, opened 2016-04-14
Adding support for Unix time formatted dates in C#

tbombach:
Addressing issue #909
Adding a new known primary type for Unix time formatted dates (which is in seconds since Unix Epoch). Wiring up code namers to handle this primary type and set the correct.

Implementing the JsonConverter for C# and adding to templates for serializing/deserializing a DateTime object with the Unix time format. Adding path and body tests in C# to validate this functionality

### DIFF
--- a/AutoRest/AutoRest.Core/Utilities/Extensions.cs
+++ b/AutoRest/AutoRest.Core/Utilities/Extensions.cs
@@ -244,7 +244,7 @@ namespace Microsoft.Rest.Generator.Utilities
         }
 
         /// <summary>
-        /// Returns true is the type is a PrimaryType with KnownPrimaryType matching typeToMatch.
+        /// Returns true if the type is a PrimaryType with KnownPrimaryType matching typeToMatch.
         /// </summary>
         /// <param name="type"></param>
         /// <param name="typeToMatch"></param>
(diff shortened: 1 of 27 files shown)


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

jianghaolu:
Please pull from upstream master to get a hot fix in Java. Thanks!

AutoRest/AutoRest.Core/Utilities/Extensions.cs
```diff
@@ -263,5 +263,54 @@ public static bool IsPrimaryType(this IType type, KnownPrimaryType typeToMatch)
             }
             return false;
         }
+
```
tbombach: @markcowl Mark, could you take a quick look at the changes I made in this last commit whenever it's convenient for you?

For context, MethodTemplateModel.GetSerializationSettingsReference() is getting too complex and failing a build rule for code complexity (and it's hard to read). I simplified this logic by adding methods to check if an IType either is a KnownPrimaryType or is a dictionary/sequence that is based on the KnownPrimaryType (previously these checks were repeated for each primary type that had special serialization behavior).

Tests are passing and the serialization is working the same as before. I'd like another set of eyes on it before I merge to make sure I kept the behavior the same with respect to the null checks. Thanks!

amarzavery:
The changes look good to me.

PR #969, opened 2016-04-22
Added support for required parameters constructors

brnleehng:
Proposal:
Currently, there are already static constructors that the user can use without the required parameters in the service client templates. By removing the readOnly attribute for parameters, the user can simply note the parameter to be required. This will then be added to the required parameter constructors. As of right now, setting a parameter to be readOnly and required to be true does not work.

Changes:
- Removed the readOnly extension
- Regenerated the CSharp/Azure CSharp tests
- Minor: Made doc strings in the Service Client template to be more accurate

### DIFF
--- a/AutoRest/Generators/CSharp/Azure.CSharp.Tests/Expected/AcceptanceTests/AzureBodyDuration/AutoRestDurationTestService.cs
+++ b/AutoRest/Generators/CSharp/Azure.CSharp.Tests/Expected/AcceptanceTests/AzureBodyDuration/AutoRestDurationTestService.cs
@@ -36,14 +36,14 @@ namespace Fixtures.Azure.AcceptanceTestsAzureBodyDuration
         public Uri BaseUri { get; set; }
 
         /// <summary>
-        /// Gets or sets json serialization settings.
+        /// Gets JSON serialization settings.
         /// </summary>
         public JsonSerializerSettings SerializationSettings { get; private set; }
 
         /// <summary>
-        /// Gets or sets json deserialization settings.
+        /// Gets JSON deserialization settings.
         /// </summary>
-        public JsonSerializerSettings DeserializationSettings { get; private set; }        
+        public JsonSerializerSettings DeserializationSettings { get; private set; }
 
         /// <summary>
         /// Gets Azure subscription credentials.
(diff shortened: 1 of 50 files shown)


azurecla:
Hi __@brnleehng__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (brhoan). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

brjohnstmsft:
Is there a related issue we can look at, preferably with examples, that explains what this change does?

brnleehng:
The example we were looking at was.. Subscription Id is not part of the required parameter constructor. One issue is if the user doesn't set the subscription id then their operation will fail. The subscription id should just be set in the constructor if that's the case. The user shouldn't need to know that they have to set the subscription id before calling operations. This ony happens when we set the parameter to have the required attribute in the spec.

Old:
BatchManagementClient(Uri baseUri, ServiceClientCredentials credentials, params DelegatingHandler[] handlers)

New:
BatchManagementClient(Uri baseUri, ServiceClientCredentials credentials, string subscriptionId, params DelegatingHandler[] handlers)

AutoRest/Generators/CSharp/Azure.CSharp.Tests/Expected/AcceptanceTests/AzureBodyDuration/AutoRestDurationTestService.cs
```diff
@@ -36,14 +36,14 @@ public partial class AutoRestDurationTestService : ServiceClient<AutoRestDuratio
         public Uri BaseUri { get; set; }
 
         /// <summary>
-        /// Gets or sets json serialization settings.
+        /// Gets JSON serialization settings.
```
matthchr: "Gets the JSON serialization settings." would be better?

AutoRest/Generators/CSharp/Azure.CSharp.Tests/Expected/AcceptanceTests/AzureBodyDuration/AutoRestDurationTestService.cs
```diff
@@ -36,14 +36,14 @@ public partial class AutoRestDurationTestService : ServiceClient<AutoRestDuratio
...
         /// </summary>
         public JsonSerializerSettings SerializationSettings { get; private set; }
 
         /// <summary>
-        /// Gets or sets json deserialization settings.
+        /// Gets JSON deserialization settings.
         /// </summary>
-        public JsonSerializerSettings DeserializationSettings { get; private set; }        
+        public JsonSerializerSettings DeserializationSettings { get; private set; }
```
matthchr: Same comment here about "the"

amarzavery:

jianghaolu:

PR #973, opened 2016-04-22
Revert "Escape data strings for odata queries."

hovsepm:
Reverts Azure/autorest#902


amarzavery:
@begoldsm - This PR reverts your PR #902. It caused breaking changes in the Resource Manager client.

PR #996, opened 2016-05-02
[Java & Java.Azure] Create service once in each client

jianghaolu:
To avoid the overhead of calling `someClient.someOps()`.


jianghaolu:

Issue #634, opened 2016-01-21
x-ms-parameter-grouping emits incorrect comment for parameter group type


brjohnstmsft:
Using `x-ms-parameter-grouping` in multiple operations results in a class with a summary comment that is derived from only the first operation in the Swagger file.

For example, here is the declaration of a parameter group class used by many operations, not just `CreateOrUpdate`:

```csharp
    /// <summary>
    /// Additional parameters for the CreateOrUpdate operation.
    /// </summary>
    public partial class SearchRequestOptions
```

@matthchr

PR #1002, opened 2016-05-03
Makes the documentation for the Parameter Groups more generic so that it no longer includes just a single operation name when there could be several

John-Hart:
Fixes issue #634 where only the first operation name was added to the documentation for a Parameter Group


AutoRest/Generators/Extensions/Extensions/Extensions.cs
```diff
@@ -410,13 +410,18 @@ public static void AddParameterGroups(ServiceClient serviceClient)
                         parameterGroupType = new CompositeType
                         {
                             Name = parameterGroupName,
-                            Documentation = "Additional parameters for the " + method.Name + " operation."
+                            Documentation = "Additional parameters for these operations: " + method.Name
```
brjohnstmsft: The list of operations could be potentially very long. In the limit, it could be every operation supported by the client. Do we really need to list them in the comment?

John-Hart: The documentation does wrap in the summary and intellisense truncates the text but we certainly don't need to list them.  How about just a simple summary of "Additional parameters for one or more operations" ?

brjohnstmsft: I am fine with that.

tbombach:
Looks good to me

tbombach:
Also, could you rename the PR, since we decided to summarize instead of appending the operation names?

Issue #993, opened 2016-04-30
Fix LRO PATCH and simplification


amarzavery:
This PR https://github.com/Azure/autorest/pull/875 fixed LRO PATCH in node.js and also simplified it. Similar changes need to be made in C#, Java, Ruby, Python so that every language is consistent.

tbombach:
Thanks for opening this. I'm working on the C# changes

annatisch:
@amarzavery  - are you able to clarify how PATCH operations were fixed? I'm having trouble finding it amongst the refactor changes in the PR diff....

amarzavery:
@annatisch - sure.

Previously implemented long running PATCH was incorrectly giving preference to provisioning state value.

For a PATCH, the resource has to be successfully created as a prerequisite. Hence, its provisioning state will always be succeeded. We were checking that would return without polling.

The right thing to do for a PATCH operation is to look for Azure Async Operation Header and Location Header in the initial response. If they are present then we should poll (get) on the link provided in those headers until a terminal state is reached.

We were doing this kind of polling for long running POST and DELETE. Hence, we modified the logic for PATCH to not look for provisioning state and do the right thing.

As a part of this modification we realized that everything can be simplified into one method named getLongRunningOperationStatus() (by checking the initial method and the statuscode) rather than having the previous two methods (getPutOrPatch.., getPostOrDelete..). Hence this was reduced to one.

We still kept the old methods for backwards compatibility. These methods internally call the new method.

Hope this helps :).

vishrutshah:
Removing Ruby Label as the Fix is in.

annatisch:
@amarzavery, thanks so much - that helped lots!!
I think I've finally got this resolved for Python - I'd appreciate it if you could take a quick glance over my latest PR.

jianghaolu:
*(Not for Java runtime GA as this won't be a breaking change. Will fix it later.)*

PR #1011, opened 2016-05-07
[In Review] Patch support for long running operation in ruby

vishrutshah:
This is regarding Issue #875 & #993 for Ruby. + Simplifying the long running operation results calls.

### DIFF
--- a/ClientRuntimes/Ruby/ms-rest-azure/lib/ms_rest_azure/azure_service_client.rb
+++ b/ClientRuntimes/Ruby/ms-rest-azure/lib/ms_rest_azure/azure_service_client.rb
@@ -15,99 +15,35 @@ module MsRestAzure
     attr_accessor :api_version
 
     #
-    # Retrieves the result of 'PUT' operation. Perfroms polling of required.
+    # Retrieves the result of 'POST','DELETE','PUT' or 'PATCH' operation. Performs polling of required.
     # @param azure_response [MsRestAzure::AzureOperationResponse] response from Azure service.
     # @param custom_deserialization_block [Proc] custom logic for response deserialization.
     #
     # @return [MsRest::HttpOperationResponse] the response.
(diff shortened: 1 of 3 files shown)


azurecla:

TTYL, AZPRBOT;

ClientRuntimes/Ruby/ms-rest-azure/lib/ms_rest_azure/azure_service_client.rb
```diff
@@ -64,74 +63,52 @@ def get_put_operation_result(azure_response, custom_deserialization_block)
...
+      polling_state.get_operation_response
+    end
+
+    # @TODO Update name of the method to get_put_or_patch_operation_result when introducing breaking change / updating major version of gem
+    # Retrieves the result of 'PUT' or 'PATCH' operation. Performs polling of required.
+    # @param azure_response [MsRestAzure::AzureOperationResponse] response from Azure service.
+    # @param custom_deserialization_block [Proc] custom logic for response deserialization.
+    #
+    # @return [MsRest::HttpOperationResponse] the response.
+    def get_put_operation_result(azure_response, custom_deserialization_block)
```
veronicagg: Let's check on whether we need to keep this one or not per our offline conversation.

veronicagg:
@vishrutshah Changes look good to me. Thanks!

ClientRuntimes/Ruby/ms-rest-azure/lib/ms_rest_azure/azure_service_client.rb
vishrutshah: As per our offline discussion, we found the use case when it's required for backward compatibility Example: *_mgmt_* gem uses old template (0.2.2) with ms_rest_azure moves to 0.2.2 to 0.2.3 with this change then we can break.

PR #1032, opened 2016-05-12
Fix for a null reference exception when not providing namespace

tbombach:
In JavaCodeNamer.cs, when the namespace parameter is not provided as a command line parameter, it throws a null reference exception when calling `ToLower()`. Even though namespace is necessary for Java code to be generated, this code path is still reached when calling AutoRest with no parameters (e.g. when viewing help).


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

PR #1039, opened 2016-05-13
Java: Use SNAPSHOT runtime as dependency

jianghaolu:
A tiny fix as beta1 is a bit outdated.
No need to merge in `release` branch as runtime releases are separate.


tbombach:
LGTM

PR #1040, opened 2016-05-13
Refactor parameter grouping and enable better comments

matthchr:
- Parameter grouping code is clearer.
  - Models for parameter grouping now more clearly state what
    operations they are used on.


matthchr:
Just ignore the second comment if you want to look at the AutoRest core changes.  Sample documentation output for the doc change can be seen in the second commit when I regenerate the expected files for test.

brjohnstmsft:
@matthchr The list of operations can get pretty long. Do we even need the list?

matthchr:
@brjohnstmsft I think the list is useful because without it every single operation group object has exactly the same comment. For my use case, we end up with a bunch of group objects which say "An operation group for some operations" -- and there are 30 of those... but none say what operation they are for.  Obviously you can sort of figure it out from the name, but having an explicit reference helps add clarity I think...

On the other hand, I see why having a list of 70 or 80 of these in the comment would be annoying...

Any ideas?  Maybe we need a description field in the Swagger specification which can be used to override having the list? (i.e. if you specify a description it'll just use that instead of doing the autogenerated thing)?

AutoRest/Generators/Extensions/Extensions/ParameterGroupExtensionHelper.cs
```diff
@@ -0,0 +1,204 @@
...
+                List<Parameter> parametersToRemoveFromMethod = new List<Parameter>();
+
+                foreach (ParameterGroup parameterGroup in parameterGroups)
+                {
+                    CompositeType parameterGroupType =
+                        generatedParameterGroups.FirstOrDefault(item => item.Name == parameterGroup.Name);
+
+                    IEnumerable<Method> methodsWhichUseGroup = GetMethodsUsingParameterGroup(serviceClient.Methods, parameterGroup);
+
+                    string methodList = string.Join(", ", methodsWhichUseGroup.Select(m => string.Format("{0}_{1}", m.Group, m.Name)));
```
itowlson: It looks like this could be moved inside the if (parameterGroupType == null) check?

itowlson: Given the Batch scenario where most types are used for a single operation, the "for the following operations: Foo." reads awkwardly to me.  We could produce more natural text at the expense of a small helper function:

static string OperationsText(IList<string> ops)
{
    if (ops.Count == 1) return ops[0] + " operation";
    if (ops.Count == 2) return ops[0] + " and " + ops[1] + " operations";
    return String.Join(", ", ops.Take(ops.Count - 1)) + ", and " + ops.Last() + " operations";
}

itowlson:
We could also put a count check in and use the 'generic' text if there are more than, say, five operations.

---
In reply to: [218937556](https://github.com/Azure/autorest/pull/1040#issuecomment-218937556) [](ancestors = 218937556)

matthchr:
@brjohnstmsft What do you think about @itowlson's suggestion about capping it at some fixed number of operations, after which it no longer includes the list but rather some generic text?

brjohnstmsft:
That could work. I like this better though:

> Any ideas? Maybe we need a description field in the Swagger specification which can be used to override having the list? (i.e. if you specify a description it'll just use that instead of doing the autogenerated thing)?

To give more detail on my scenario: I'm generating clients from two different Swagger specs -- one for data plane, one for "metadata plane" (not really the same as management plane). I want to have a parameter group shared across both of the generated clients' operations. I do this by ensuring that their declarations are identical in both Swagger specs, then I delete one of the generated files. The problem with listing operations for me is that I end up with operation names in the comment for only one of the planes (data or metadata). If it's purposefully an incomplete list, that's not so bad, but it would be better if I could just specify my own description text in the Swagger.

matthchr:
@markcowl @tbombach -- Can you take a look at this and merge it if it looks good?
I think @brjohnstmsft and @itowlson have had their comments addressed.

tbombach:
Looks good, the parameter grouping code is easier to follow now that some of the `AddParameterGroups` logic is split up into private methods

Issue #642, opened 2016-01-25
There is no way to avoid generating and using models and use shared objects as in wsdl.


yvdh:
in WSDL we could force the use of serializable shared client objects, i.e. full objects that are also DTO objects. This doesn't seem possible with AutoRest, and is a real miss: I don't want to include an extra set models and use AutoMapper to transform those to the 'real' full objects (which include functionality).
Any suggestions?

stankovski:
You can tag the model as [`x-ms-external`](https://github.com/Azure/autorest/blob/master/Documentation/swagger-extensions.md#x-ms-external) which will cause the model to be not generated.

yvdh:
Tried that before posting, did not work! Should have mentioned that, my bad. 
I currently use a script to do some source code tinkering and that works.

stankovski:
Ahh, I see that "x-ms-external" is in Azure extensions. Changing title to indicate that the solution should be to move the extension under generic code generator.

PR #1043, opened 2016-05-13
CSharp: add support for excluding models from a proxy by referencing …

jhancock93:
…external assembly

- Change CustomSettings to be dictionary of object to allow for some settings to be arrays (this is used for referencing one or more external assemblies)
- Allow CodeGenSettings to be passed in as a json file on the commandline rather than the settings in the original swagger document (better for this to be a decision by the person generating the proxy than the service provider)
- Change CSharpGenerator to be able to reference external assemblies and
namespaces when assembly is passed in on commandline (like WCF svcutil). Namespaces of the matching types are automatically referenced in the proxy so the proxy does not need to be hand-edited.

### DIFF
--- a/AutoRest/Generators/CSharp/CSharp/CSharpCodeGenerator.cs
+++ b/AutoRest/Generators/CSharp/CSharp/CSharpCodeGenerator.cs
@@ -1,13 +1,18 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Globalization;
 using System.Threading.Tasks;
 using Microsoft.Rest.Generator.ClientModel;
 using Microsoft.Rest.Generator.CSharp.Templates;
 using System.Linq;
+using System.Reflection;
+using Microsoft.Rest.Generator.CSharp.Properties;
+using Microsoft.Rest.Generator.Logging;
 
 namespace Microsoft.Rest.Generator.CSharp
 {
@@ -19,6 +24,7 @@ namespace Microsoft.Rest.Generator.CSharp
         public CSharpCodeGenerator(Settings settings) : base(settings)
         {
             _namer = new CSharpCodeNamer();
+            ReferencedNamespaces = new HashSet<string>();
             IsSingleFileGenerationSupported = true;
         }
 
@@ -36,6 +42,18 @@ namespace Microsoft.Rest.Generator.CSharp
         [SettingsAlias("syncMethods")]
         public SyncMethodsGenerationMode SyncMethods { get; set; }
 
+        /// <summary>
+        /// Allows a client to reference models from another assembly instead of generating new models
+        /// </summary>
+        [SettingsInfo("ExternalModelAssemblies")]
+        [SettingsAlias("modelAssemblies")]
+        public string[] ExternalModelAssemblies { get; set; }
+
+        /// <summary>
+        /// Referenced namespaces
+        /// </summary>
+        public HashSet<string> ReferencedNamespaces { get; private set; }
+
         public override string Name
         {
             get { return "CSharp"; }
@@ -49,7 +67,8 @@ namespace Microsoft.Rest.Generator.CSharp
 
         public override string UsageInstructions
         {
-            get {
+            get
+            {
                 return string.Format(CultureInfo.InvariantCulture,
                     Properties.Resources.UsageInformation, ClientRuntimePackage);
             }
@@ -72,12 +91,14 @@ namespace Microsoft.Rest.Generator.CSharp
         /// <param name="serviceClient"></param>
         public override void NormalizeClientModel(ServiceClient serviceClient)
         {
+            ReferencedNamespaces.UnionWith(ResolveExternalReferences(serviceClient));
             PopulateAdditionalProperties(serviceClient);
             Extensions.NormalizeClientModel(serviceClient, Settings);
             _namer.NormalizeClientModel(serviceClient);
             _namer.ResolveNameCollisions(serviceClient, Settings.Namespace,
                 Settings.Namespace + ".Models");
         }
+
 
         private void PopulateAdditionalProperties(ServiceClient serviceClient)
         {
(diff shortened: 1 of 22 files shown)


azurecla:
Hi __@jhancock93__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!

TTYL, AZPRBOT;

jhancock93:
This is some what related to issue #642 but is a different way of dealing with external types, more along the lines of how the WCF svcutil allowed types to be referenced from external assemblies.

jhancock93:
Rebased pull request to more recent version of master to avoid merge conflicts.

jhancock93:
I have one test that I added that passes on Windows but is failing when run by Travis. Any suggestions on how to reproduce the issue locally?

tbombach:
@azuresdkci test this please!

jhancock93:
I see there are now changes that require conflicts to be merged. I can rebase this branch on top of master again and fix the conflicts. Will one of the admins please respond as to whether I should go ahead and do that? This PR has been sitting idle for 2 weeks...

tbombach:
@jhancock93 I'll take a look at this today and leave any comments if there are any necessary changes before you have to rebase.

AutoRest/AutoRest.Core/Settings.cs
```diff
@@ -305,10 +320,30 @@ public static void PopulateSettings(object entityToPopulate, IDictionary<string,
                     {
                         try
                         {
-                            if (setting.Value.IsNullOrEmpty() && property.PropertyType == typeof(bool))
+                            if (setting.Value.ToString().IsNullOrEmpty() && property.PropertyType == typeof(bool))
```
tbombach: This removes the null check. If setting.Value is null, ToString() will throw, before the IsNullOrEmpty() call

AutoRest/AutoRest.Core/GlobalSuppressions.cs
```diff
@@ -60,3 +60,4 @@
 [assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1002:DoNotExposeGenericLists", Scope = "member", Target = "Microsoft.Rest.Generator.ClientModel.ServiceClient.#Methods")]
 [assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1002:DoNotExposeGenericLists", Scope = "member", Target = "Microsoft.Rest.Generator.ClientModel.ServiceClient.#Properties")]
 [assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1002:DoNotExposeGenericLists", Scope = "member", Target = "Microsoft.Rest.Generator.ClientModel.ParameterTransformation.#ParameterMappings")]
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Scope = "member", Target = "Microsoft.Rest.Generator.Settings.#Create(System.Collections.Generic.IDictionary`2<System.String,System.Object>)")]
```
tbombach: I don't think you need this suppression anymore, since you removed the try/catch from an earlier commit

AutoRest/Generators/CSharp/CSharp/CSharpCodeGenerator.cs
```diff
@@ -87,6 +108,67 @@ private void PopulateAdditionalProperties(ServiceClient serviceClient)
...
+                    modelTypesToRemove.Add(typeGroup.Key);
+                    modelNamespaces.Add(matchingExportedTypes[0].Namespace);
+                }
+                else if (matchingExportedTypes.Count > 1)
+                {
+                    Logger.LogError(new ArgumentException("ExternalModelAssemblies"),
+                            Properties.Resources.ConflictingTypesError, typeGroup.Key,
+                            string.Join(", ", referenceAssemblies.Select(a => a.GetName())));
+                }
+
```
tbombach: Minor: remove extra newline here

AutoRest/Generators/CSharp/CSharp/CSharpCodeGenerator.cs
```diff
@@ -87,6 +108,67 @@ private void PopulateAdditionalProperties(ServiceClient serviceClient)
             }
         }
 
+
```
tbombach: Minor: extra newline

AutoRest/AutoRest.Core/GlobalSuppressions.cs
jhancock93: Good catch! I have removed this line.

AutoRest/Generators/CSharp/CSharp/CSharpCodeGenerator.cs
jhancock93: Removed.

AutoRest/Generators/CSharp/CSharp/CSharpCodeGenerator.cs
jhancock93: Removed.

jhancock93:
Rebased on latest master and addressed code review comments.

hovsepm:

jhancock93:
I don't understand what might cause the "No test results found" error in the "default" check. Is there something I need to do from my end to fix the problem?

fearthecowboy:
Howdy, @jhancock93 -- We've been discussing this PR in our internal standup.

I really like the parts about using a response-file on the command line for pulling in the settings. If you give me a PR with that separately, we'll take that right away.

We're mulling over how we want to handle skipping code-gen of referenced classes. Currently you can mark a type as `x-ms-external ` ( see https://github.com/Azure/autorest/blob/master/Documentation/swagger-extensions.md#x-ms-external ) .

I've also got some ideas swirling around on how to do some declarations a bit more generically so we can apply the same pattern against multiple language generators.

Is there a something that `x-ms-external` doesn't do that you'd need in order to use that?

jhancock93:
Thanks for taking a look at this PR. Let me see what I can do to make a separate PR for the settings file.

Referencing external types is a complex issue, so I'm glad to have a discussion about it. I was aiming for functionality along the lines of what SvcUtil provides for WCF services. It provides similar functionality to exclude types from a proxy based on what it finds in a externally referenced DLL. I have a team that is transitioning a bunch of services from SOAP to WebAPI, so it was convenient to be able to reference the same contract assemblies for SOAP proxies generated by SvcUtil and WebAPI proxies generated by AutoRest. Before I implemented my changes, I looked at the x-ms-external tag and there were a few issues with it that I could see:

1. As I recall, x-ms-external only works with the Azure generator right now.
2. The x-ms-external tag doesn't say anything about what namespace the type is in. So unless the type is in a namespace that would already be referenced by the client, the client won't compile without manual modifications. In contrast, my PR finds the namespace associated with types with matching names in the external DLL and adds them to a namespaces property that then creates using statements in the client so it compiles with no manual code changes.
3. Because the x-ms-external tag is part of the swagger json, the decision whether to share types or not must be made by the author of the service API, not the client owner that consumes the API. In constrast, my change allows both of kinds of clients to be generated, one that generates the types as part of the client and one that references the external DLL depending on the needs of the client. 
4. For this particular use case of sharing a C# contracts assembly, we would only want to exclude types when doing CSharp proxy generation and would want to still generate model classes for other languages. That is something I can do with this methodology but not with x-ms-external because it is part of the swagger definition instead of a client generation option.

fearthecowboy:
>> As I recall, x-ms-external only works with the Azure generator right now.

IIRC, yeah, that's on my list of things to fix.

>> The x-ms-external tag doesn't say anything about what namespace the type is in. So unless the type is in a namespace that would already be referenced by the client, the client won't compile without manual modifications. In contrast, my PR finds the namespace associated with types with matching names in the external DLL and adds them to a namespaces property that then creates using statements in the client so it compiles with no manual code changes.

That was one of my assumptions, glad to see it validated.

>> Because the x-ms-external tag is part of the swagger json, the decision whether to share types or not must be made by the author of the service API, not the client owner that consumes the API. In constrast, my change allows both of kinds of clients to be generated, one that generates the types as part of the client and one that references the external DLL depending on the needs of the client.

Currently true, but I'm thinking of a way to allow you to specify that tangential to the swagger itself.

>> For this particular use case of sharing a C# contracts assembly, we would only want to exclude types when doing CSharp proxy generation and would want to still generate model classes for other languages. That is something I can do with this methodology but not with x-ms-external because it is part of the swagger definition instead of a client generation option.

And that's my other assumption validated. 

Again, I am thinking about a different workaround that would get us past the need to embed it in the swagger, but ... you're right, if the namespace isn't the same as the expected namespace, it's not going to find it...

Lemme get back to you.

jhancock93:
Any more discussion on this topic? I see that the team is currently working to reorganize the folder structure in the repo...

devigned:
@fearthecowboy and @tbombach can we get a thumbs up or down on this functionality prior to pulling in #1220?

fearthecowboy:
This will need to be refactored to work in the new code base (going into a branch `RefactoredCodeModel` today)

Fixing externals is something that's definitely in my short term planning.

PR #992, opened 2016-04-29
Implement x-ms-parameterized-host extension for ruby

vishrutshah:
This is regarding Issue #722. 

Implementing support for x-ms-parameterized-host in Ruby generator.

PR #1051, opened 2016-05-17
Get formatted reference value for only sequence type or return name

vishrutshah:
Ruby MethodTemplateModel used to have conditions like 
```
if (param.Type is SequenceType)  
{  
    encodedParameters.Add(string.Format("'{0}' => {1}", urlPathName, param.GetFormattedReferenceValue()));  
}  
else  
{  
    encodedParameters.Add(string.Format("'{0}' => {1}", urlPathName, variableName));  
}  

```
As part of the refactoring code in PR #992 that was removed and delegated to the method [GetFormattedReferenceValue](https://github.com/Azure/autorest/blob/master/AutoRest/Generators/Ruby/Ruby/ClientModelExtensions.cs#L178). That code was applying the ToString method on all the type that is not sequence type but it should supposed to return that parameter name as it.

Due to this change the URL request got changed in such a way that for nil values it did type conversion to ToString() and started constructing url with empty value of the key when key was nil causing the VCR tests to break.

### DIFF
--- a/AutoRest/Generators/Ruby/Ruby/ClientModelExtensions.cs
+++ b/AutoRest/Generators/Ruby/Ruby/ClientModelExtensions.cs
@@ -180,7 +180,7 @@ namespace Microsoft.Rest.Generator.Ruby.TemplateModels
             SequenceType sequence = parameter.Type as SequenceType;
             if (sequence == null)
             {
-                return parameter.Type.ToString(parameter.Name);
+                return parameter.Name;
             }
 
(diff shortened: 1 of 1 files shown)


azurecla:

TTYL, AZPRBOT;

veronicagg:
@vishrutshah Looks good to me. There were only those 2 references to this method, so it's fine to update the method itself. Thanks!

tbombach:
@azuresdkci retest this please!

vishrutshah:
Thanks for the review !

PR #1061, opened 2016-05-19
Fix generated file name

daschult:


tbombach:
The current Azure Schema Generator expected files were generated before this change, correct?: https://github.com/daschult/autorest/commit/5604cf9552433b824a70a8df517237665eefaffd#diff-93cfae68e5d9696c25fd9e0f2525425cR51 (which I assume introduced the issue with the file names)

The change in this PR does not require regeneration, because they are already the correct paths?

daschult:
Correct.

tbombach:
Merged

PR #1068, opened 2016-05-20
Fix Swagger SchemaResolver to allow indirectness/inheritence when matching

TimLovellSmith:
Because properties may be defined by chains of indirect references to some type which inherits from another compatible type found by chains of indirect references, and that should be fine!


azurecla:
Hi __@TimLovellSmith__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (tilovell). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

TimLovellSmith:
CI build seems to have encountered problems downloading something - how do I get it to try again?

TimLovellSmith:
It looks like I broke  ClientModelWithCircularDependencyThrowsError [FAIL]
      Assert.Throws() Failure
      Expected: typeof(System.ArgumentException)
      Actual:   (No exception was thrown)

I am really not sure if this is really because the change I made is wrong. The test case includes a circular referenced based on a cycle of 'allOf' references, but I think the problem is that the code of Dereference, is designed to follow the chain of 'ref' references to its ultimate conclusion: not 'allOf' references, and therefore doesn't detect the circularity.

TimLovellSmith:
Note that circular 'allOf' is not necessarily even impossible to resolve in the swagger model - you can just follow the circle until it repeats itself, and assume the result is the closure of all the things visited. Some language generators may have trouble with generating code to reflect that, however. :-)

TimLovellSmith:
And circular 'ref' from type to itself if via properties list is also not necessarily impossible to resolve - it should correspond to recursive data structures like binary trees, etc.

TimLovellSmith:
I'm going to go ahead and add some logic to satisfy the test for now, even though I'm not sure its a great test, it seems like path of least resistance.

tbombach:
@azuresdkci add to whitelist

TimLovellSmith:
Hrm. Now I see code analysis failed...

replace the call to 'string.ToLower(CultureInfo)' with String.ToUpperInvariant().

tbombach:
Could you simplify (and rename) the test swagger file so that it's more targeted for the scenario that it's validating? In the modeler tests, we're not testing against a real API specification, they're intended to be unit tests that make sure the correct model can be built for a given feature in Swagger

TimLovellSmith:
@tbombach I've revised the pull request to simplify the test swagger file, and to add a full set of assertions for all the expected inheritance relationships.

TimLovellSmith:
I'm hereby adding to the PR turning off CA1822 for test projects, because it makes the pull request process unnecessarily painful.

TimLovellSmith:
Is the CI build flaky or what? '

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated

'

tbombach:
@TimLovellSmith The AutoRest repo was just switched from DNX to CoreCLR RC2, so that might be causing some flakiness. I'll watch your builds and restart them if they're failing for a transient reason or if it's unrelated to your changes

TimLovellSmith:
@tbombach Thanks, CI checks passed!

tbombach:
Merged. Thanks @TimLovellSmith for the contribution (and for working through the test flakiness)!

Issue #669, opened 2016-02-01
Adding exception type to generated SDK code


kuangweizhang:
When user is using generated SDK, C# on Get request for example, user cannot find what kind of exception will be thrown if the resource cannot be found by the comments in the C# source code. The only possible way is to either read the source code a lot further or try it out to see what exception will be thrown. This is not a good experience.
I hope we can provide type of exception in the generated code comment. Similar idea with Exception section in this page: https://msdn.microsoft.com/en-us/library/z6c2z492(v=vs.110).aspx

stankovski:
We can add [`<exception>`](https://msdn.microsoft.com/en-us/library/w1htk11d.aspx) to XML doc

PR #1076, opened 2016-05-24
Issue 669 - added Exception xml doc comments

John-Hart:
Fix issue #669 by adding Exception xml doc comments to generated code for CSharp for known exceptions


kirthik:
Looks good to me. Can you squash all the commits into one commit so its easier to tag commit to issue?

PR #1089, opened 2016-05-25
Add DNS schema acceptance test

daschult:
I had to make some changes to the AzureResourceSchema generator to generate the DNS schema. Specifically, I had to support creating resources when the resource type included a reference to a enum path parameter ({recordType}). In this case, I had to create a new resource definition for each of the possible enum values of the enum parameter. This has the same effect as if the swagger document had a different method for each of the enum values.


hovsepm:

PR #1098, opened 2016-05-27
Adding script to can verify that developer system has correct setup

fearthecowboy:
And a script to install the missing Android SDK libraries (if at least the base of the SDK is installed)

(re-issuing PR after mistakenly killing my original branch)


fearthecowboy:

PR #1103, opened 2016-05-31
Changing nuget reference from YamlDotNet to YamlDotNet.Signed

tbombach:
Since we're referencing it from a signed assembly, we need to use the signed version


tbombach:
@fearthecowboy Whenever you get a chance, please let me know if you see any issues with this change.

fearthecowboy:
:+1: 
Looks good to me.

PR #1104, opened 2016-05-31
 Fix bug with custom client request id generation

matthchr:
- Fix bug where parameter grouping broke custom client request id generation.
- Add tests to check above case


matthchr:
For context, this happened because of the way the code handled the extension `x-ms-client-request-id`.

Note that this extension is specified on the _parameter_ not on the method.

**What happened (for operations which _didn't_ use parameter grouping):**
1. Extension was specified in swagger.
2. During code generation, the code generator called the helper method `AzureExtensions.GetClientRequestIdString` which looked through the `method.Parameters` property and found the one with the `x-ms-client-request-id` extension, and used the value specified there.

**What happened (for operations which _did_ use parameter grouping)**
1. Extension was specified in swagger
2. Parameter grouping ran, moved this parameter into the parameter group, and removed it from the `method.Parameters` collection.
3. Code generation called the `AzureExtensions.GetClientRequestIdString` which looked through all the parameters in `method.Parameters` for those with the `x-ms-client-request-id` extension, but since that parameter was removed by parameter grouping it wasn't there anymore.

The fix is to process the extension and "move it up" to the `Method` level provided it meets some criteria (specifically there should be at most one occurrence of the `x-ms-client-request-id` extension across all of the parameters.

matthchr:
Travis seems to be being flaky, not sure why.

matthchr:
Hmm, it looks like Travis either re-ran itself or somebody re-ran it for me, since now it's passing.

matthchr:
@tbombach @hovsepm @markcowl -- Can one of you take a look at this and merge it if it looks okay?

tbombach:
@matthchr I'll look at it today

tbombach:
Merged

PR #1107, opened 2016-05-31
Changing YamlDotNet signing exception

tbombach:
Since the dll doesn't follow the same naming convention as Newtonsoft


fearthecowboy:
:+1: 

Looks good to me.

PR #1108, opened 2016-06-01
requestLibrarySink should be called with requestOptions if no additional arguments specified

karataliu:
In requestPipline.create, requestLibrarySink should be called with requestOptions if no additional arguments specified.

In following commit
https://github.com/Azure/autorest/commit/876f74b86c9b9b3ffe876e9d6de64c50a0a2e499#diff-c89a34dee81e6cb2f17bfc58c53cc484L117

requestLibrarySink is changed to return a function taking requestOptions, but only the case when 'arguments.length === 0 is false' is updated, should also update the remaining case.

### DIFF
--- a/ClientRuntimes/NodeJS/ms-rest/lib/requestPipeline.js
+++ b/ClientRuntimes/NodeJS/ms-rest/lib/requestPipeline.js
@@ -142,7 +142,7 @@ exports.requestLibrarySink = function (requestOptions) {
 exports.create = function (requestOptions) {
   return function () {
     if (arguments.length === 0) {
-      return exports.createWithSink(exports.requestLibrarySink);
+      return exports.createWithSink(exports.requestLibrarySink(requestOptions));
     }
     // User passed filters to add to the pipeline.
     // build up appropriate arguments and call exports.createWithSink


azurecla:
Hi __@karataliu__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (doliu). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

hovsepm:

tbombach:
@amarzavery This LGTM, I'll merge it if it doesn't raise any flags for you.

ClientRuntimes/NodeJS/ms-rest/lib/requestPipeline.js
```diff
@@ -142,7 +142,7 @@ exports.requestLibrarySink = function (requestOptions) {
 exports.create = function (requestOptions) {
   return function () {
     if (arguments.length === 0) {
-      return exports.createWithSink(exports.requestLibrarySink);
+      return exports.createWithSink(exports.requestLibrarySink(requestOptions));
```
amarzavery: How does this help?
If arguments.length === 0; which means nothing was passed as an argument while calling the create method. Then what would you gain by passing undefined (as requestOptions would be undefined in this scenario) to requestLibrarySink.

If something was passed as an argument while calling the create method then that something will be passed to  requestLibrarySink in the return statement on line 149.

Issue #1057, opened 2016-05-18
JsonExtensionData needs to be supported by AutoRest


pomortaz:
For the swagger JSON fields that are not well-defined and schema-less, we can add "additionalProperties" field as follow:

    "def1": {
      "properties": {
       "field1": {
          "type": "string"
       }
      },
      "additionalProperties": {
        "type": "object"
     }
    }

In C# this should be translated to a property, tagged with [JsonExtensionData] attribute similar to [this.](https://github.com/pomortaz/azure-sdk-for-net/blob/master/src/KeyVault/Microsoft.Azure.KeyVault/Models/Messages.cs#L58) 

        [JsonExtensionData]
        public Dictionary<string, object> AdditionalProperties { get; set; }

So if there is a schema-less JSON string in the service response, similar to field2 in the following example (for def1 in the swagger spec above), field2 should be serialized to ‘Dictionary<string, object> AdditionalProperties’ property:
 
    {
        "def1": {
            "field1": "this is field 1",
            "field2": {
                "field3": "this is field 3"
             }
         }
    }

fearthecowboy:
Hey @pomortaz ... I'm looking into this, just working out how the same thing can be equally implemented in the other languages, I just have to verify what the behavior would be if there is flattening going on.

I should have some more information in the next day or so.

amarzavery:
@pomortaz  Extending your example to explain @fearthecowboy 's point.

Let's assume that field2 and field3 were also mapped.
- Actual response payload
```
{
    "def1": {
        "field1": "this is field 1",
        "field2": {
            "field3": "this is field 3",
           "field4": "I am not mapped in swagger",
           "field5": "I too am not mapped in swagger. What to do? Where should I go?"
         }
     }
}
```


- Swagger schema definition
```
"def1": {
  "properties": {
   "field1": {
      "type": "string"
   },
   "field2": {
     "x-ms-client-flatten": true,
     "properties": {
         "field3": {
           "type": "string" // Think of this as something like provisioning state
         }
     }
  },
  "additionalProperties": {
    "type": "object"
 }
}
```



So after deserialization,
```
{
    "def1": {
      "field1": "this is field 1", // This will be deserialized properly
      "field3": "this is field 3", //This will be flattened during deserialization
      "additionlaProperties": {
       
//How should unmapped properties look like over here?

     A) Should they be flattened like their siblings?
         "field4": "I am not mapped in swagger",
[...]

PR #1112, opened 2016-06-01
Support additionalProperties that accept all unbound properties. (see #1057)

fearthecowboy:
This adds the support for a catch-all `additionalProperties` when there are other properties in the object.

Support in .net added by using JSON.NET's `JsonExtensionData` attribute.

Other languages will need to determine how to direct all other unbound properties into the 'AdditionalProperties' member.


AutoRest/Modelers/Swagger/SchemaBuilder.cs
```diff
@@ -38,9 +38,12 @@ public override IType BuildServiceType(string serviceTypeName)
             _schema = Modeler.Resolver.Unwrap(_schema);
 
             // If primitive type
-            if ((_schema.Type != null && _schema.Type != DataType.Object) ||
-                _schema.AdditionalProperties != null)
+            if (_schema.Type != null && _schema.Type != DataType.Object || ( _schema.AdditionalProperties != null && _schema.Properties.IsNullOrEmpty() ))
             {
+                // Notes:
```
tbombach: Could you update the [documentation here](https://github.com/Azure/autorest/blob/master/Documentation/defining-clients-swagger.md#dictionaries) too? Doesn't necessarily have to be with this PR if you just want to merge this as-is, but just mentioning it so that we don't forget to do it later

tbombach:
LGTM. Nice!

AutoRest/Modelers/Swagger/SchemaBuilder.cs
fearthecowboy: Will do. Thanks!

Issue #1121, opened 2016-06-01
The unit test MappingExtentionsTests.TestInputMapping is failing due to changes in code gen.


John-Hart:
The test is failing due to changes in code gen but the test wasn't updated.  

This is probably because this and other tests in AutoRest.Generator.Extensions.Tests and AutoRest.Generator.Azure.Extensions.Tests are missing from the gulp task test:xunit so it wasn't run.

PR #1123, opened 2016-06-01
Fixed broken test MappingExtensionTests.TestInputMappings

John-Hart:
Fixed broken test TestInputMappings for issue #1121
Added missing extensions test dlls to xunittestdlls in gulpfile.js
Updated expected files for tests added after commit  6d06584


fearthecowboy:
LGTM.

:+1:

AutoRest/Generators/Extensions/Extensions.Tests/MappingTests.cs
```diff
@@ -33,10 +33,9 @@ public void TestInputMapping()
             generator.Generate(clientModel).GetAwaiter().GetResult();
             string body = settings.FileSystem.ReadFileAsText("X:\\Payload.cs");
```
fearthecowboy: I think this will fail on non-windows ... (which is maybe why this was set to skip?)

John-Hart: I talked to Dan and he disabled it because it was failing due to the code gen change

John-Hart: I fixed the test that had been disabled to not have a hard coded path and everything is now passing @fearthecowboy can you take a quick look?

fearthecowboy:
Looks good to me. 
:+1:

Issue #1143, opened 2016-06-07
Generated python code fails to handle the model with "." in the property name


xingwu1:
Python client runtime threat "." in the property name as flatten property. However, if the property name contains ".", ex: "odata.nextLink". Client runtime will think this is the "odata" : {"nextLink" : ...}}.

PR #1144, opened 2016-06-07
Fix python code gen bug in handling "." in property name.

xingwu1:


xingwu1:
This is the fix for issue #1143

AutoRest/TestServer/swagger/paging.json
```diff
@@ -18,7 +18,7 @@
   "paths": {
     "/paging/single": {
       "get": {
-        "x-ms-pageable": { "nextLinkName": "nextLink", "itemName": "values" },
+        "x-ms-pageable": { "nextLinkName": "next.Link", "itemName": "values" },
```
amarzavery: This will cause a change in all the languages. Have you generated code for all the languages?

xingwu1: @amarzavery Yes.

tbombach: @xingwu1 Could the '.' scenario be tested in its own scenario in the test server? It makes it harder to know when this specific behavior breaks if it's part of an existing paging test. I would prefer to see it as a test in a new path in a swagger file, not a modification of an existing one.

xingwu1:
@tbombach I have added a new endpoint with the additional model to repro the bug.

tbombach:
Thanks @xingwu1. Merged

PR #1150, opened 2016-06-08
Fix bug of parameterized host in paging next

jianghaolu:


jianghaolu:

Issue #1156, opened 2016-06-10
[Ruby] All attributes should have return type documented


devigned:
Attribute readers generated by Ruby AutoRest don't have the type returned and thus there is no intellisense provided in the IDE.

For example, the resource management client doesn't provide the type returned for it's underlying method groups. See: http://www.rubydoc.info/gems/azure_mgmt_resources/0.3.1/Azure%2FARM%2FResources%2FResourceManagementClient%3Aresource_groups. The return type is object since nothing is specified in the documentation.

At a minimum, it should be:
``` ruby
# @return [Azure::ARM::Resources::ResourceGroups] resource_groups
attr_reader :resource_groups
```

I think this is systemic in the generator and needs to be fixed globally.

vishrutshah:
Fix has been checked-in to AutoRest. Closing this bug. Thanks.

PR #1158, opened 2016-06-11
Ruby - Adding Model name on the properties of the ServiceClient

vishrutshah:
Addressing Issue #1156 

As per the existing [ServiceClient ](https://github.com/Azure/azure-sdk-ruby/blob/master/resource_management/azure_mgmt_resources/lib/azure_mgmt_resources/models/deployment_properties.rb) & [Models](https://github.com/Azure/azure-sdk-ruby/blob/master/resource_management/azure_mgmt_resources/lib/azure_mgmt_resources/models/deployment_properties.rb), we can see that properties of the model has been documents correctly but not all the properties of ServiceClient. For example, In the [resource_management_client.rb](resource_management_client.rb) properties generated from MethodGroups does not specify correct return types like for resources, tags, deployment_operations. 

This change will add the return type on properties generated from MethodGroups in the ServiceClients.

### DIFF
--- a/AutoRest/Generators/Ruby/Ruby/Templates/ServiceClientTemplate.cshtml
+++ b/AutoRest/Generators/Ruby/Ruby/Templates/ServiceClientTemplate.cshtml
@@ -35,7 +35,7 @@ module @Settings.Namespace
 
 @foreach (var operation in Model.MethodGroups)
 {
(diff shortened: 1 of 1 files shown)


azurecla:

TTYL, AZPRBOT;

veronicagg:
@vishrutshah LGTM!

vishrutshah:
Thanks for the review !!

Issue #1166, opened 2016-06-14
Generated code does not handle a null response when there's no default in the responses section


tbombach:
If an operation doesn't define a default in the responses section, our fallback logic for handling unexpected status codes is to try to read the response content stream as a string:
https://github.com/Azure/autorest/blob/master/AutoRest/Generators/CSharp/CSharp/Templates/MethodTemplate.cshtml#L296

```
            if ((int)_statusCode != 200 && (int)_statusCode != 204)
            {
                var ex = new CloudException(string.Format("Operation returned an invalid status code '{0}'", _statusCode));
                _responseContent = await _httpResponse.Content.ReadAsStringAsync().ConfigureAwait(false);
                ex.Request = new HttpRequestMessageWrapper(_httpRequest, _requestContent);
                ex.Response = new HttpResponseMessageWrapper(_httpResponse, _responseContent);
```

However, if the Content property is null, this will throw a null reference exception.

tbombach:
@pscofield When you get a chance, could you add more info to this bug (e.g. Swagger)

Copy & pasted from an offline conversation:

The strange thing is that we have tests that go to server endpoints that return null ([that's the expected behavior in this test case](https://github.com/Azure/autorest/blob/master/AutoRest/Generators/CSharp/CSharp.Tests/AcceptanceTests.cs#L269) and the [associated model code](https://github.com/Azure/autorest/blob/master/AutoRest/Generators/CSharp/CSharp.Tests/Expected/AcceptanceTests/BodyString/StringModel.cs#L126)). They try to deserialize into a model, but they always run Content.ReadAsStringAsync() first (which should repro this issue), before checking for null. That's why it's weird to me that Content is null in this case, rather than just a stream with no content.

Also, I'm not seeing anything when searching online about `HttpResponseMessage.Content` being null. There seems to be something strange in this specific case with how `HttpClient.Send()` is working. I was looking through our test suite and there are lots of tests that run `HttpResponseMessage.Content.ReadAsStringAsync()` when the server response has no body.

pscofield:
What's the ETA on this? It blocks my check-in for LogicApps SDK since the tests are broken. Thanks.

tbombach:
@pscofield I sent out a PR that I believe fixes this issue. Could you check if the null check fixes the problem in your code? (see my comment here for instructions: https://github.com/Azure/autorest/pull/1170#discussion_r67376399)

PR #1170, opened 2016-06-15
Adding a null check for reading Content from a response as a string

tbombach:
Addressing issue #1166. In some cases, `_response.Content` can be null if the server returns an empty body. If the user doesn't specify a `default` schema in the `responses` section, the code will just try to read the `Content` property as a string, without checking if it's null first.

This PR adds a null check and just sets the `_responseContent` to the empty string if it's unable to read the `HttpResponseMessage.Content` property

### DIFF
--- a/AutoRest/Generators/CSharp/Azure.CSharp.Tests/Expected/AcceptanceTests/AzureSpecials/XMsClientRequestIdOperations.cs
+++ b/AutoRest/Generators/CSharp/Azure.CSharp.Tests/Expected/AcceptanceTests/AzureSpecials/XMsClientRequestIdOperations.cs
@@ -143,7 +143,12 @@ namespace Fixtures.Azure.AcceptanceTestsAzureSpecials
             if ((int)_statusCode != 200)
             {
                 var ex = new CloudException(string.Format("Operation returned an invalid status code '{0}'", _statusCode));
-                _responseContent = await _httpResponse.Content.ReadAsStringAsync().ConfigureAwait(false);
+                if (_httpResponse.Content != null) {
+                    _responseContent = await _httpResponse.Content.ReadAsStringAsync().ConfigureAwait(false);
+                }
+                else {
+                    _responseContent = string.Empty;
+                }
                 ex.Request = new HttpRequestMessageWrapper(_httpRequest, _requestContent);
                 ex.Response = new HttpResponseMessageWrapper(_httpResponse, _responseContent);
                 if (_httpResponse.Headers.Contains("x-ms-request-id"))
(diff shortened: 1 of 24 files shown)


AutoRest/Generators/CSharp/Azure.CSharp.Tests/Expected/AcceptanceTests/HeadExceptions/HeadExceptionOperations.cs
```diff
@@ -141,7 +141,12 @@ internal HeadExceptionOperations(AutoRestHeadExceptionTestService client)
             if ((int)_statusCode != 200)
             {
                 var ex = new CloudException(string.Format("Operation returned an invalid status code '{0}'", _statusCode));
-                _responseContent = await _httpResponse.Content.ReadAsStringAsync().ConfigureAwait(false);
+                if (_httpResponse.Content != null) {
```
fearthecowboy: Just idle curiosity, but why not simplify statements like this into things like: 

```  csharp 
_httpResponse _httpResponse.Content != null ?  await _httpResponse.Content.ReadAsStringAsync().ConfigureAwait(false) : string.Empty; 
```

or better yet:

``` csharp
// would guarantee that we always get back a string, even if ReadAsStringAsync() returns null...
_httpResponse = await _httpResponse?.Content.ReadAsStringAsync().ConfigureAwait(false) ?? string.Empty; 
```

but, I'm funny that way :)

fearthecowboy:
👍 LGTM

AutoRest/Generators/CSharp/Azure.CSharp.Tests/Expected/AcceptanceTests/HeadExceptions/HeadExceptionOperations.cs
tbombach: Agreed, that's my preferred way of structuring. Always a debate with myself over how efficient vs readable to make it. Since the coalesce makes sense to you in this case too, I'll use that here.

AutoRest/Generators/CSharp/Azure.CSharp.Tests/Expected/AcceptanceTests/AzureSpecials/XMsClientRequestIdOperations.cs
```diff
@@ -143,7 +143,12 @@ internal XMsClientRequestIdOperations(AutoRestAzureSpecialParametersTestClient c
             if ((int)_statusCode != 200)
             {
                 var ex = new CloudException(string.Format("Operation returned an invalid status code '{0}'", _statusCode));
-                _responseContent = await _httpResponse.Content.ReadAsStringAsync().ConfigureAwait(false);
+                if (_httpResponse.Content != null) {
```
tbombach: @pscofield Could you see if this change fixes the issue you were seeing? (you can just copy and paste the change in this file into your generated code to see if it works). I'll merge this into master if you no longer get the null reference exception

pscofield: @tbombach Works! Thanks for fixing.

PR #1171, opened 2016-06-16
Reuse the global autorest setting.

xingwu1:
Python codegen has a individual setting called "Version" which is exactly same usage as global setting "PackageVersion". This fix is to let Python codegen reuse the global setting.


fearthecowboy:
:+1: LGTM.

tbombach:
Merged

PR #1176, opened 2016-06-16
Summary comments contain improperly escaped backslashes (Issue #1152)

### LINKED ISSUE #1152: (C#) Summary comments contain improperly escaped backslashes
The C# code generator seems to needlessly escape backslashes when generating ///<summary> comments. For example, this Swagger:

```JSON
        "pattern": {
          "type": "string",
          "default": "\\W+",
          "description": "A regular expression pattern to match token separators. Default is \"\\W+\"."
        },
```

Yields this code:

```csharp
        /// <summary>
        /// Gets or sets a regular expression pattern to match token
        /// separators. Default is "\\\\W+".
        /// </summary>
        [JsonProperty(PropertyName = "pattern")]
        public string Pattern { get; set; }
```

The comment should actually look like this:

```csharp
        /// <summary>
        /// Gets or sets a regular expression pattern to match token
        /// separators. Default is "\W+".
        /// </summary>
```

fearthecowboy:
Fixes #1152 
- includes adding new unit test project (supports test-first bug fix model :+1:  )


AutoRest/Generators/CSharp/CSharp.Unit.Tests/TestExtensions.cs
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.IO;
+using Microsoft.Rest.Generator;
+using Microsoft.Rest.Generator.CSharp;
+using Microsoft.Rest.Generator.Extensibility;
+using Microsoft.Rest.Generator.Utilities;
+
+namespace AutoRest.Csharp.Unit.Test
```
tbombach: Could you change this namespace to be `Microsoft.Rest.Generator.CSharp.Unit.Test` so it's consistent with the other projects? (Or I'd prefer `Microsoft.Rest.Generator.CSharp.UnitTest` or even `Microsoft.Rest.Generator.CSharp.Tests`, since it could share a namespace with our other test project)

tbombach:
LGTM after changing the namespace.

Awesome, this will make it much simpler to fix bugs (and have test cases for them)!

AutoRest/Generators/CSharp/CSharp.Unit.Tests/TestExtensions.cs
fearthecowboy: Done.

PR #1184, opened 2016-06-21
Added parameter documentation to model constructors

brnleehng:
Parameters documentation that contains Title or Description or both in the specs will be added to the model constructor. If Title and Description is omitted in the model specs, then the parameter documentation is also omitted in the model constructor.


Samples/azure-storage/Azure.Ruby/azure_storage/models/storage_account_list_result.rb
```diff
@@ -13,9 +13,6 @@ class StorageAccountListResult
       # their properties.
       attr_accessor :value
 
-      # @return [String]
```
fearthecowboy: Any idea why this was deleted?

Samples/azure-storage/Azure.Ruby/azure_storage/models/usage_list_result.rb
```diff
@@ -12,9 +12,6 @@ class UsageListResult
       # @return [Array<Usage>] Gets or sets the list Storage Resource Usages.
       attr_accessor :value
 
-      # @return [String]
```
fearthecowboy: Any idea why this was deleted?

brnleehng: Not sure. In my PR, I only made changes to CSharp generator.

vishrutshah: looks like commit [64a7b867315c](https://github.com/Azure/autorest/commit/64a7b867315c5e6b23222ca86dab5ca3c7d2580e) has changed the samples but I don't see any changes in Ruby samples with that check-ins.

AutoRest/Generators/CSharp/CSharp/TemplateModels/ModelTemplateModel.cs
```diff
@@ -187,6 +198,49 @@ private static string CreateSignature(IEnumerable<ConstructorParameterModel> par
                 return string.Join(", ", declarations);
             }
 
+            private static IEnumerable<string> CreateSignatureDocumentation(IEnumerable<ConstructorParameterModel> parameters)
+            {
+                var declarations = new List<string>();
+                IEnumerable<ConstructorParameterModel> parametersWithDocumentation =
+                    parameters.Where(p => !(string.IsNullOrEmpty(p.UnderlyingProperty.Summary) && string.IsNullOrEmpty(p.UnderlyingProperty.Documentation)));
+
+                foreach (var property in parametersWithDocumentation.Where(p => !p.UnderlyingProperty.IsConstant).Select(p => p.UnderlyingProperty))
```
matthchr: Could move the LINQ in here out of the loop and into it's own variable (or pop it up into the other LINQ statement you have above) so it reads like:

```
 IEnumerable<ConstructorParameterModel> parametersWithDocumentation = 
    parameters.Where(p => !(string.IsNullOrEmpty(p.UnderlyingProperty.Summary) &&
    string.IsNullOrEmpty(p.UnderlyingProperty.Documentation)) && !p.UnderlyingProperty.IsConstant).Select(p.UnderlyingProperty);
```

It's minor but it gets all the selection/filtering stuff out of the way early so that you can just do your loop.

AutoRest/Generators/CSharp/CSharp/TemplateModels/ModelTemplateModel.cs
```diff
@@ -187,6 +198,49 @@ private static string CreateSignature(IEnumerable<ConstructorParameterModel> par
...
+                {
+                    string documentation = string.Empty;
+                    string formatString = "<param name=\"{0}\">{1}</param>";
+
+                    if (string.IsNullOrEmpty(property.Documentation))
+                    {
+                        documentation = string.Format(
+                            CultureInfo.InvariantCulture,
+                            formatString,
+                            char.ToLower(property.Name[0]) + property.Name.Substring(1),
```
matthchr: Refactor this into a variable so you don't repeat it 3 times

AutoRest/Generators/CSharp/CSharp/TemplateModels/ModelTemplateModel.cs
```diff
@@ -187,6 +198,49 @@ private static string CreateSignature(IEnumerable<ConstructorParameterModel> par
...
+                            string.Empty);
+                    }
+                    else
+                    {
+                        documentation = string.Format(
+                            CultureInfo.InvariantCulture,
+                            formatString,
+                            char.ToLower(property.Name[0]) + property.Name.Substring(1),
+                            property.Documentation);
+                    }
```
matthchr: This whole `if` `else` is actually identical except for the last parameter -- instead of doing `if` `else` like this, just do:
```
var documentationInnerText = property.Summary ?? property.Documentation ?? string.Empty
documentation = string.Format(
                            CultureInfo.InvariantCulture,
                            formatString,
                            char.ToLower(property.Name[0]) + property.Name.Substring(1),
                            documentationInnerText);
```

Samples/azure-storage/Azure.Ruby/azure_storage/models/storage_account_list_result.rb
xingwu1: This is due to my commit (https://github.com/Azure/autorest/commit/64a7b867315c5e6b23222ca86dab5ca3c7d2580e) which revert my previous commit (https://github.com/Azure/autorest/commit/6b8360de308a91706f08a648cb9256a7f968617c). Somehow my full code regen didn't revert the ruby file.

AutoRest/Generators/CSharp/CSharp/TemplateModels/ModelTemplateModel.cs
```diff
@@ -201,39 +201,20 @@ private static string CreateSignature(IEnumerable<ConstructorParameterModel> par
...
-                            property.Summary);
-                    }
+                foreach (var property in parametersWithDocumentation)
+                {
+                    var documentationInnerText = property.Summary ?? property.Documentation;
+                    var documentation = string.Format(
+                        CultureInfo.InvariantCulture,
+                        "<param name=\"{0}\">{1}</param>",
+                        char.ToLower(property.Name[0], CultureInfo.InvariantCulture) + property.Name.Substring(1),
+                        documentationInnerText);
```
matthchr: This still allows there to be a `<param name="Foo"\></param>` -- you shouldn't allow such a case I think.

You need a check for if `documentationInnerText` is `NullOrEmpty` and if it is don't even add it to the `declarations`

brnleehng: I think it's fine because in my linq query, I check for NullOrEmpty for the summary/description. There has to be some documentation. Otherwise, it will just omit it.

PR #1200, opened 2016-06-27
[DO NOT MERGE] Temp

matthchr:


fearthecowboy:
Killing this PR -- trying to reduce the noise as I get close to the great reorg.

PR #1205, opened 2016-06-29
[experiments] Refactoring validation model to use attributes

tbombach:
Changing how validation rules are described to using attributes on model classes. Previously, validation loop was an imperative traversal of the ServiceDefinition tree, with validation occurring at each level (and rules defined there). Instead, the tree is automatically traversed and each validation rule is applied for the property that it decorates.


azurecla:
    <span>
    </span>

TTYL, AZPRBOT;

PR #1210, opened 2016-06-29
Migration to CoreCLR RTM

hovsepm:
* Updated all the package references to CoreCLR Rtm versions.
* Increased C# ClientRuntime patch versions
* Moved part of LroHappyPathTests that is failing on Linux build to LroHappyPathTestsRest and skipped the latter one.


AutoRest/Generators/CSharp/Azure.CSharp.Tests/project.json
```diff
@@ -24,14 +24,14 @@
...
+      "type": "platform",
+      "version": "1.0.0"
     }, 
-    "Microsoft.NETCore.Platforms": "1.0.1-rc2-24027",
+    "Microsoft.NETCore.Platforms": "1.0.1",
     "CSharp.Tests": "2.0.0",
-    "Microsoft.Rest.ClientRuntime.Azure": "[3.3.0,4.0.0)",
-    "xunit": "2.1.0",
-    "dotnet-test-xunit": "1.0.0-rc2-build10015"
+    "Microsoft.Rest.ClientRuntime.Azure": "[3.3.1,4.0.0)",
```
fearthecowboy: FYI, for peer references like:
``` json
   "Microsoft.Rest.ClientRuntime.Azure": "[3.3.1,4.0.0)", 
```

I've been using:

``` json
   "Microsoft.Rest.ClientRuntime.Azure": {  "target": "project" } 
```
Which tells dotnet-cli to use the sibling project. 

I don't think you should change this, at this point, but I will check out what happens in the final package if you do it with `target:project`...

fearthecowboy:
👍  LGTM

Issue #1212, opened 2016-06-29
THE GREAT CODE REORG


fearthecowboy:
In order to attack `MAX_PATH` issues, I'm reorganizing the whole repo today.

We have far too much replication in path names, and excessive nesting leads to copy/paste errors when dealing with multiple projects nested inconsistently.

Thomas and I have discussed the options, and we're comfortable with the following layout:

``` bash
# Title: Rename (and change namespaces!) and put in a subfolder of /src
# Note, this does mean that I have to change NAMESPACES in all the C# code (with Resharper) 
# But this will trim a lot of excessive redundancy from the whole project 

# Core
src\core\AutoRest                                         # was Microsoft.Rest.Generator.Cli , and before that AutoRest
src\core\Autorest.Core                                    # was Microsoft.Rest.Generator
src\core\AutoRest.Core.Tests                              # was Microsoft.Rest.Generator.Tests
src\core\AutoRest.Extensions                              # was Microsoft.Rest.Generator.Extensions      
src\core\AutoRest.Extensions.Azure                        # was Microsoft.Rest.Generator.Extensions.Azure
src\core\AutoRest.Extensions.Azure.Tests                  # was Microsoft.Rest.Generator.Extensions.Azure.Tests
src\core\AutoRest.Extensions.Tests                        # was Microsoft.Rest.Generator.Extensions.Tests

# Generators
src\generator\AutoRest.AzureResourceSchema        # was Microsoft.Rest.Generator.AzureResourceSchema
src\generator\AutoRest.AzureResourceSchema.Tests  # was Microsoft.Rest.Generator.AzureResourceSchema.Tests
src\generator\AutoRest.CSharp                     # was Microsoft.Rest.Generator.CSharp
src\generator\AutoRest.CSharp.Azure               # was Microsoft.Rest.Generator.CSharp.Azure
src\generator\AutoRest.CSharp.Azure.Tests         # was Microsoft.Rest.Generator.CSharp.Azure.Tests
src\generator\AutoRest.CSharp.Tests               # was Microsoft.Rest.Generator.CSharp.Tests
src\generator\AutoRest.CSharp.Unit.Tests          # was Microsoft.Rest.Generator.CSharp.Unit.Tests
src\generator\AutoRest.Java                       # was Microsoft.Rest.Generator.Java
src\generator\AutoRest.Java.Azure                 # was Microsoft.Rest.Generator.Java.Azure
[...]

PR #1220, opened 2016-06-30
Re-organize entire code repo

fearthecowboy:
See #1212


azurecla:
Hi __@fearthecowboy__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

tbombach:
A heads-up for our contributors: this PR will be rearranging almost everything in the AutoRest repository. We want to give you fair warning that folder structure will change. Garrett's made sure that everything compiles and that tests are passing still, but we want to avoid unexpected surprises the next time you pull after this is merged.

cc: @annatisch @jianghaolu @amarzavery @daschult @lmazuel @vishrutshah @begoldsm @mattchr @xingwu1 @veronicagg @NiklasGustafsson @brnleehng

tbombach:
@fearthecowboy I don't think NodeJSGeneratorTests.sln got updated. It still references projects called `Azure.NodeJS.Tests` and `Azure.NodeJS.Tests/Azure.NodeJS.Tests.njsproj`

tbombach:
@fearthecowboy I know you're disabling the nuget test for now, but I think the path should be updated for client runtime:
https://github.com/fearthecowboy/autorest/blob/reorg/PackageTest/NugetPackageTest/package.json

tbombach:
@fearthecowboy Do the namespaces/output folder prevent CSharp client runtimes from living in a folder under `/src/client/CSharp`? It seems strange for all the other languages to have their own folder while C# doesn't

https://github.com/fearthecowboy/autorest/tree/reorg/src/client

tbombach:
@fearthecowboy I think the python .gitignore file needs to be updated to prevent python client runtime build artifacts from being checked into source. It's possible that the gitignore rules never were correct though, since they start at the location of .gitignore. I could be wrong though
https://github.com/fearthecowboy/autorest/blob/reorg/src/client/Python/.gitignore
and
https://github.com/fearthecowboy/autorest/blob/reorg/gitignore-python (btw, this file looks like a clone of /src/client/Python/.gitignore and I don't think it will be used by git, correct?)

(see the lines like: `ClientRuntimes/Python/Python35-64-ENV/Scripts/activate.bat`)

tbombach:
@fearthecowboy The AutoRest nuget .nuspec needs to be updated to reflect the namespace changes for generators, modelers, etc (this is the nuget for acquiring AutoRest.exe itself):

https://github.com/fearthecowboy/autorest/blob/reorg/src/core/AutoRest.Core/AutoRest.nuspec

tbombach:
Paths for the gulp command to sync dependencies versions needs to be updated. The paths to the source files for some of the code generators refer to the old places:
https://github.com/fearthecowboy/autorest/blob/reorg/Tools/gulp/gulp-runtime-version-sync.js

fearthecowboy:
@tbombach 

> I don't think NodeJSGeneratorTests.sln got updated. It still references projects called Azure.NodeJS.Tests and Azure.NodeJS.Tests/Azure.NodeJS.Tests.njsproj, NodeJS.Tests, etc
> https://github.com/fearthecowboy/autorest/blob/reorg/NodeJSGeneratorTests.sln

Ah. Good point. Lemme go fix that.

<hr>

> Another NodeJS related update, the included XML files that start with AutoRest.Generator.*. It could be that these aren't used, because AutoRest.Generator.Azure.Common doesn't seem to be the name of any > assembly before the reorg.
> https://github.com/fearthecowboy/autorest/blob/reorg/src/generator/AutoRest.NodeJS.Azure.Tests/Azure.NodeJS.Tests.njsproj

Hmm. Missed that. I'll go look into it. 

<hr>

> The Python.Tests.Azure.pyproj should be called Python.Azure.Tests.pyproj and the <RootNamespace> should be AutoRest.Python.Azure.Tests
> https://github.com/fearthecowboy/autorest/blob/reorg/src/generator/AutoRest.Python.Azure.Tests/Python.Tests.Azure.pyproj

Yeaah.. I'll fix that too.

<hr>

> @fearthecowboy I know you're disabling the nuget test for now, but I think the path should be updated for client runtime:
> https://github.com/fearthecowboy/autorest/blob/reorg/PackageTest/NugetPackageTest/package.json

I really just moved all those files into the PackageTest folder, it's really not going to run, and I didn't want to bother spending the time cleaning that up at all yet since it's going away with my next PR.

> I'm not sure, but I think the AutoRest.nuget.proj folder needs to be updated too:
> https://github.com/fearthecowboy/autorest/blob/reorg/src/core/AutoRest.Core/AutoRest.nuget.proj

> @fearthecowboy The AutoRest nuget .nuspec needs to be updated to reflect the namespace changes for generators, modelers, etc (this is the nuget for acquiring AutoRest.exe itself):
> https://github.com/fearthecowboy/autorest/blob/reorg/src/core/AutoRest.Core/AutoRest.nuspec


Hmm. Yeah. um... look over there! what's that! <runs away>

I guess I should update that, just so that the nightly won't break for the bit between now and when I get the next PR in.

<hr>

> @fearthecowboy Do the namespaces/output folder prevent CSharp client runtimes from living in a folder under /src/client/CSharp? It seems strange for all the other languages to have their own folder while C# doesn't https://github.com/fearthecowboy/autorest/tree/reorg/src/client

So.... I toyed around with that a bit, and here's the problem: 

I *reaaaaaaaaly* wanted the `.csproj` (and soon to be `.xproj` ) files to be at the same relative depth from the root, which makes it infinitely easier to maintain relative paths to common files and locations.
I would have unbundled the language-specific ones ... but there's often stuff in the root of those folders, and didn't want to complicate the `client` folder. 

<hr>

> @fearthecowboy I think the python .gitignore file needs to be updated to prevent python client runtime build artifacts from being checked into source. It's possible that the gitignore rules never were correct though, since they start at the location of .gitignore. 
> I could be wrong though https://github.com/fearthecowboy/autorest/blob/reorg/src/client/Python/.gitignore and
> https://github.com/fearthecowboy/autorest/blob/reorg/gitignore-python (btw, this file looks like a clone of /src/client/Python/.gitignore and I don't think it will be used by git, correct?)
> (see the lines like: ClientRuntimes/Python/Python35-64-ENV/Scripts/activate.bat)

Ouch, that file (`gitignore-python`) wasn't supposed to be there. I'll remove that.
I ran all the tests and didn't end up with any extra files, but I'll double check it all.

<hr>

> @fearthecowboy Paths for the gulp command to sync dependencies versions needs to be updated. The paths to the source files for some of the code generators refer to the old places:
> https://github.com/fearthecowboy/autorest/blob/reorg/Tools/gulp/gulp-runtime-version-sync.js

Ouch. forgot about that too. Fixing...

tbombach:
LGTM

Issue #1228, opened 2016-07-05
Code does not compile when model property docs have line breaks


stankovski:
This bug was introduced by #1184

stankovski:
Issue with C# code gen

PR #1229, opened 2016-07-05
Fix for an issue #1228

stankovski:
Fix for an issue #1228


fearthecowboy:
I applied the fix to the new reorg'd HEAD and merged it in #1243

PR #1245, opened 2016-07-10
Ruby - Polymorphic discriminator hashmap should have exact model name

vishrutshah:
Polymorphic discriminator hashmap should have exact model name as hash value instead of lowercased in Ruby.

### DIFF
--- a/src/client/Ruby/ms-rest/lib/ms_rest/serialization.rb
+++ b/src/client/Ruby/ms-rest/lib/ms_rest/serialization.rb
@@ -146,7 +146,7 @@ module MsRest
           parent_class = get_model(mapper[:type][:class_name])
           discriminator = parent_class.class_eval("@@discriminatorMap")
           model_name = response_body["#{mapper[:type][:polymorphic_discriminator]}"]
-          model_class = get_model(discriminator[model_name].capitalize)
+          model_class = get_model(discriminator[model_name])
         else
           model_class = get_model(mapper[:type][:class_name])
         end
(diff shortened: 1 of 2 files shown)


azurecla:

TTYL, AZPRBOT;

fearthecowboy:

veronicagg:
@vishrutshah how did you find this was a problem? do we need to add a test for the case? 
why did we choose to capitalize before? Change looks good to me, thanks!

vishrutshah:
@veronicagg With current implementation of polymorphic discriminator we are generating hash map with value as the model name but lower cased names. This caused me to capitalize the name in the serialization refactoring assuming that the map naming is correct but it turns out that assumption was wrong. 

I found this with the test of DataLake Analysis gem where we have [JobProperties](https://github.com/vishrutshah/azure-sdk-ruby/pull/3/files#diff-a160d991bf17b34f669c1bd9557a7e4eR15) has map with model names lower cased but it contains more than on words like (USqlJobProperties / HiveJobProperties) where capitalization doesn't work. As per my knowledge, the only way to test this is seems like having swagger with these kind of case with x-ms-polymorphic-discriminator but that seems like corner case, so i did not pursue that. What are you views on this?

veronicagg:
@vishrutshah Thanks for the clarification. I agree is it's a special case, though I would add this issue/case to our backlog, so when we work on test coverage for autorest we can include a case to cover this. thanks!

vishrutshah:
Sounds good. Added it to the backlog list. 

@veronicagg thanks for the review!!

PR #1249, opened 2016-07-11
[Ruby - In Review] Adding kind property on the resource model in ms_rest_azure. 

vishrutshah:
This is to resolve issue https://github.com/Azure/azure-sdk-ruby/issues/18.

### DIFF
--- a/src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/resource.rb
+++ b/src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/resource.rb
@@ -23,6 +23,10 @@ module MsRestAzure
     # @return [Hash{String => String}] the tags attached to resources (optional).
     attr_accessor :tags
 
+    # @return [String] the kind of the resource.
+    # FIXME kind attribute was introduced because of SiteInstance model of azure_mgmt_web gem has kind attribute.
+    attr_accessor :kind
+
     def self.mapper
       {
         required: false,
@@ -73,6 +77,13 @@ module MsRestAzure
                         name: 'String'
                       }
                   }
+                }
+              },
+              kind: {
+                required: false,
+                serialized_name: 'kind',
+                type: {
+                  name: 'String'
                 }
               }
           }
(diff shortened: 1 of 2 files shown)


azurecla:

TTYL, AZPRBOT;

src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/resource.rb
```diff
@@ -23,6 +23,10 @@ class Resource
     # @return [Hash{String => String}] the tags attached to resources (optional).
     attr_accessor :tags
 
+    # @return [String] the kind of the resource.
+    # FIXME kind attribute was introduced because of SiteInstance model of azure_mgmt_web gem has kind attribute.
```
veronicagg: If 'kind' becomes part of resource, we may not need a 'fixme' comment.

veronicagg:
@vishrutshah left one comment, change looks good to me, thanks!

src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/resource.rb
vishrutshah: @veronicagg I completely agree but i haven't found any swagger other than web doing this, so i was thinking to keep it until 'kind' officially becomes part of all resource.

devigned: Kind is in this [storage spec](https://github.com/Azure/azure-rest-api-specs/blob/master/arm-storage/2016-01-01/swagger/storage.json). Thoughts?

vishrutshah: @devigned I see the "kind" but it is not part of the Resource definition. azure_arm_web is adding "kind" in the [resource ](https://github.com/Azure/azure-rest-api-specs/blob/master/arm-web/2015-08-01/swagger/service.json#L21441)itself. Isn't that different?

vishrutshah:
@veronicagg @devigned Thanks for the feedback. Closing this PR with respect to https://github.com/Azure/autorest/pull/1263 that'll avoid failures on unknown properties of the object while serializing.

PR #1255, opened 2016-07-12
Ruby - ms_rest should throw validation error instead of NameError

vishrutshah:
ms_rest should throw validation error instead of NameError when property not found on ruby object.

Reference: http://apidock.com/ruby/Object/instance_variable_get

### DIFF
--- a/src/client/Ruby/ms-rest/lib/ms_rest/serialization.rb
+++ b/src/client/Ruby/ms-rest/lib/ms_rest/serialization.rb
@@ -315,7 +315,12 @@ module MsRest
 
         unless model_props.nil?
           model_props.each do |key, value|
-            instance_variable = object.instance_variable_get("@#{key}")
+            begin
+              instance_variable = object.instance_variable_get("@#{key}")
+            rescue NameError
+              fail ValidationError, "instance variable '#{key}' is expected on '#{object}'."
+            end
+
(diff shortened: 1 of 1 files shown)


src/client/Ruby/ms-rest/lib/ms_rest/serialization.rb
```diff
@@ -315,7 +315,12 @@ def serialize_composite_type(mapper, object, object_name)
 
         unless model_props.nil?
           model_props.each do |key, value|
-            instance_variable = object.instance_variable_get("@#{key}")
+            begin
+              instance_variable = object.instance_variable_get("@#{key}")
+            rescue NameError
+              fail ValidationError, "instance variable '#{key}' is expected on '#{object}'."
```
veronicagg: should we print object.class?

veronicagg:
@vishrutshah left one comment, rest looks good to me. Thanks!

src/client/Ruby/ms-rest/lib/ms_rest/serialization.rb
vishrutshah: Yup that's right. Corrected.

veronicagg:
@vishrutshah LGTM!

vishrutshah:
Thanks for the review!!

PR #1262, opened 2016-07-13
Throw code gen error if pageable operation does not return pageable type

veronicagg:
If return type of a pageable operation does not contain a collection, there should be a mismatch in the spec, and we can't find the element type of the collection that should be returned by the operation.


azurecla:
Hi __@veronicagg__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
        This seems like a small (but important) contribution, so no contribution license agreement is required at this point. Real humans will now evaluate your PR.
    </span>

TTYL, AZPRBOT;

vishrutshah:
@veronicagg LGTM.

PR #1265, opened 2016-07-13
Remove HEAD operation status code warning

matthchr:
- HEAD methods now only LogInfo instead of LogWarning when
   pointing out that you might be mistaken if you're missing
   two status codes.

### DIFF
--- a/src/core/AutoRest.Extensions.Azure/AzureExtensions.cs
+++ b/src/core/AutoRest.Extensions.Azure/AzureExtensions.cs
@@ -111,7 +111,7 @@ namespace AutoRest.Extensions.Azure
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


matthchr:
@Porges This should resolve our issue

matthchr:
For context: Since AutoRest allows both `boolean` and non `boolean` returns for HEAD methods, it seems silly to flag up a warning saying that "you might be mistaken" -- and those warnings cause problems if you run AutoRest during a build with warnings as errors turned on.

Moved this to be an informative statement, rather than a warning for the above reasons.

fearthecowboy:
LGTM
:+1:

Porges:
Thanks! Is 0.17 going to be released anytime soon, or should we just pull a nightly for now?

PR #1278, opened 2016-07-15
Sanitize inputs

fearthecowboy:
This includes 
- some utilities classes that make linq a lot more useful.
- changes `Summary`, `Documentation` and `Description` properties in the model to actively strip out control characters 
- Added `Debug` level severity for messages.
- refactored Validation framework a bit. 
  - Validations now directly control the `Severity` and the `TemplateMessage`  instead of relying on static registration in some constants
  - refactored the logic to use simple LINQ statements
  - added some comments in some of the classes.

- Added three specific validations:
  - `MissingValidator` - applies reports when a Property doesn't have a validator (`Severity` == `Debug`)
  - `IsIdentifier` - applies to Properties that should be considered an Identifier (`Severity` == `Error`)
  - `NoControlCharacters` - applies to all strings, does not permit control characters in strings. (`Severity` == `Error`)


azurecla:
Hi __@fearthecowboy__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

src/core/AutoRest.Core/Validation/RecursiveObjectValidator.cs
```diff
@@ -1,85 +1,77 @@
...
+                return Enumerable.Empty<ValidationMessage>();
+            }
+
+            // ensure that the rules can be re-enumerated without re-evaluating the enumeration.
+            var collectionRules = rules.ReEnumerable();
+
+            var list = entity as IList;
+            if (list != null)
+            {
+                // Recursively validate each list item and add the
```
fearthecowboy: Sorry, I couldn't resist rewriting this a bit. 

The original implementation has an extra reallocation of the entire list (rather than being able to `Select` on it directly. 

Same thing with the IDictionary implementation.

src/core/AutoRest.Core/Validation/RecursiveObjectValidator.cs
```diff
@@ -1,85 +1,77 @@
...
-                            }
-                        }
-                    }
-                }
-                else if (entity.GetType().IsClass && entity.GetType() != typeof(string))
+                return Enumerable.Empty<ValidationMessage>();
+            }
+
+            // ensure that the rules can be re-enumerated without re-evaluating the enumeration.
+            var collectionRules = rules.ReEnumerable();
```
fearthecowboy: Without this, the `collectionRules ` enumerable would be constantly re-evaluating the original enumeration every time it was used. The `ReEnumerable` class caches the contents as they are pulled thru  so that subsequent uses of the same sequence are a lot cheaper. (and it doesn't do the work up front, like `ToList` or `ToArray` would.

src/core/AutoRest.Core/Validation/RecursiveObjectValidator.cs
```diff
@@ -1,85 +1,77 @@
...
+
+            // if this is a class, validate it's value and it's properties.
+            if (entity.GetType().IsClass && entity.GetType() != typeof(string))
+            {
+                // return the messages for both this value and its properties.
+                return ValidateObjectValue(entity, collectionRules).Concat(ValidateObjectProperties(entity));
+            }
+
+            // validate just the value.
+            return ValidateObjectValue(entity, collectionRules);
```
fearthecowboy: This was needed to validate string elements in an `IList`

src/core/AutoRest.Core/Validation/RecursiveObjectValidator.cs
```diff
@@ -1,85 +1,77 @@
...
         }
 
-        private IEnumerable<ValidationMessage> ValidateObjectValue(object entity, IEnumerable<RuleAttribute> collectionRules)
+        private IEnumerable<ValidationMessage> ValidateObjectValue(object entity,
+            IEnumerable<Rule> collectionRules)
         {
             // Get any rules defined for the class of the entity
-            var classRules = entity.GetType().GetCustomAttributes<RuleAttribute>(true);
+            var classRules = entity.GetType().GetValidationRules();
```
fearthecowboy: I pulled thru the `Rule`s themselves instead of holding onto the `RuleAttribute` because this way in the future we can instantiate more rules from other sources (like external annotation) without having to wrap them.

src/core/AutoRest.Core/Validation/RecursiveObjectValidator.cs
```diff
@@ -97,7 +89,17 @@ private IEnumerable<ValidationMessage> ValidateObjectProperties(object entity)
                 var value = prop.GetValue(entity);
 
                 // Get any rules defined on this property and apply them to the property value
-                var propertyRules = prop.GetCustomAttributes<RuleAttribute>(true);
+                var propertyRules = prop.GetValidationRules().ReEnumerable();
+
+                // if the property has no defined rules, let's supply some defaults
+                if (propertyRules.IsNullOrEmpty())
+                {
+                    propertyRules = DefaultRules.ReEnumerable();
```
fearthecowboy: Applied when there are none.

src/core/AutoRest.Core/Validation/RecursiveObjectValidator.cs
```diff
@@ -97,7 +89,17 @@ private IEnumerable<ValidationMessage> ValidateObjectProperties(object entity)
...
+                }
+
+                // add universal property rules
+                propertyRules = propertyRules.Concat(UniversalRules).ReEnumerable();
```
fearthecowboy: Applied always.

src/core/AutoRest.Core/Validation/Rule.cs
```diff
@@ -19,66 +16,23 @@ protected Rule()
         }
 
         /// <summary>
-        /// The name of the exception that describes this rule
+        /// The template message for this Rule.
```
fearthecowboy: I moved the connection of the `Message` and the  `Severity` to the Rule, so we didn't have to maintain a dictionary elsewhere.

src/core/AutoRest.Core/Validation/RulesExtensions.cs
```diff
@@ -0,0 +1,65 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using AutoRest.Core.Utilities.Collections;
+using Newtonsoft.Json;
+
+namespace AutoRest.Core.Validation
+{
+    internal static class RulesExtensions
```
fearthecowboy: Moved to it's own file.

src/core/AutoRest.Core/Validation/TypedRule.cs
```diff
@@ -38,9 +29,8 @@ public virtual IEnumerable<ValidationMessage> GetValidationMessages(T entity)
             object[] formatParams;
             if (!IsValid(entity, out formatParams))
             {
-                yield return CreateException(Exception, formatParams);
+                yield return new ValidationMessage(this, formatParams);
```
fearthecowboy: I cleaned this up and put the functionality in the constructor.

src/core/AutoRest.Core/Validation/ValidationExceptionName.cs
```diff
@@ -1,21 +0,0 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License. See License.txt in the project root for license information.
-
-namespace AutoRest.Core.Validation
-{
-    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Xms")]
```
fearthecowboy: I eliminated this because it required Validators to pre-register in this file, which would make it more of a pain for generators to add their own.

src/core/AutoRest.Core/Validation/ValidationMessage.cs
```diff
@@ -1,31 +1,63 @@
...
 {
+    /// <summary>
+    /// Represents a single validtion violation (can just a debug message, informational, warning, error, or fatal error)
+    /// </summary>
     public class ValidationMessage
     {
-        private IList<string> _path = new List<string>();
+        public ValidationMessage(Rule rule)
+        {
+            Type = rule.GetType();
```
fearthecowboy: Using the `Type` instead of an `Enum` makes this more flexible.

fearthecowboy:
@tbombach I, uh... ran with it. 

I still have a bit more I'd like to do with this, but I wanted to get the changes in before I left for vacation. 

If you're OK with it, go ahead and merge it.

tbombach:
Merged with PR #1291

PR #1289, opened 2016-07-20
Taking over PR 1278 (Sanitize inputs)

tbombach:
Cleaning up a few items from PR #1278

- Code formatting
- Excluding newline characters from being stripped for right now after talking through it with PM. Since a lot of existing strings rely on them, and they are already escaped by the WrapComment() method and the Swagger spec allows these characters (GFM is permitted), it doesn't make sense to just strip them without providing a way of outputting newlines. We should revisit this (since it's up to each generator template to escape the newlines themselves), but simply stripping them would be a big change in how we handle these strings without warning for people who rely on this behavior.
- Renaming some rule classes to match previous exception names


azurecla:
    <span>
    </span>

TTYL, AZPRBOT;

PR #1301, opened 2016-07-21
Ruby bug fixes

vishrutshah:
1. Client Runtime should not throw on unknown provisioning state from server
2. Enum model names must be PascalCase
3. Ruby model should fall back to the current class model when no discriminator found from server
4. Invalid enum values should not throw exceptions. only warnings
5. Default values for enum for EnumType should be escape quoted
6. Client template should not include models


azurecla:

TTYL, AZPRBOT;

veronicagg:
@vishrutshah LGTM!

vishrutshah:
Thanks for the review !!

PR #1320, opened 2016-08-01
Changing linting rules for $ref with siblings and requiring default responses

tbombach:
- Changing $ref with siblings rule to info, since we don't handle it correctly in AutoRest yet 
- Changing the default responses rule to just require a response (since the default behavior of using code and message is fine)
- Making the "AvoidNestedProperties" rule respect the x-ms-client-flatten extension value


azurecla:
    <span>
    </span>

TTYL, AZPRBOT;

Issue #470, opened 2015-11-08
Allow Models directory name argument


corydeppen:
My API has a `Model` type that generates a `Models` class containing the operations relative to that type. This creates an error `The namespace 'MyNamespace' already contains a definition for 'Models'`. If the autorest executable was able to accept an argument for the `Models` directory name, it would allow the default name of `Models` to be overridden if a developer knew there would be a naming conflict with an existing type.

PR #1324, opened 2016-08-03
Added ModelsName Setting parameter and update CSharp to use this in place of the hard coded "Models"

John-Hart:
This Fixes issue #470 for CSharp. 
CSharp will now use the ModelsName for models output folder and models namespace


azurecla:
Hi __@John-Hart__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

src/core/AutoRest.Core/Settings.cs
```diff
@@ -81,6 +82,13 @@ public Settings()
         public string Input { get; set; }
 
         /// <summary>
+        /// Gets or sets a name for the generated client models Namespace and Models output folder
+        /// </summary>
+        [SettingsInfo("Name to use for the generated client models namespace and folder name. Not supported by all code generators.")]
```
fearthecowboy: I'm wondering if we _want_ to force the `models` folder as a sub-folder/sub-namespace of the parent (ie, by doing it as you've done here, declared as just the last `.part` ), or if we should allow them to specify the whole namespace/foldername. (ie, `-ModelsName Foo.Bar.Bin.Baz` ) 

That would require the code generator to calculate the folder name a bit more carefully  (ie, generated relative between the `client` namespace/path and the `models` namespace/path.

And does this let us choose the same namespace for the client as the model itself? Is it possible that someone **would** want that?

(food for thought...)

src/core/AutoRest.Core/Settings.cs
```diff
@@ -81,6 +82,13 @@ public Settings()
...
+        [SettingsAlias("mname")]
+        public string ModelsName { get; set; }
```
fearthecowboy: And if this was the whole namespace (as opposed to just the suffix) then you wouldn't have to do string smashing down below...

src/generator/AutoRest.CSharp/Templates/EnumTemplate.cshtml
```diff
@@ -2,7 +2,7 @@
 @inherits AutoRest.Core.Template<AutoRest.CSharp.TemplateModels.EnumTemplateModel>
 @Header("// ")
 @EmptyLine
-namespace @(Settings.Namespace).Models
+namespace @(Settings.Namespace).@(Settings.ModelsName)
```
fearthecowboy: (the string smashing I was referring to)

src/generator/AutoRest.CSharp.Unit.Tests/Bug470.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace AutoRest.CSharp.Unit.Tests
+{
+    public class Bug470 : BugTest
```
fearthecowboy: (ignore this comment, just checking for codeflow vs github differences)

fearthecowboy:
LGTM :+1:

PR #1331, opened 2016-08-05
[Ruby] Adding known Azure Environments in ruby runtime for easy discovery

vishrutshah:
NodeJS Reference: https://github.com/Azure/autorest/blob/master/src/client/NodeJS/ms-rest-azure/lib/azureEnvironment.js

### DIFF
--- a/src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/active_directory_service_settings.rb
+++ b/src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/active_directory_service_settings.rb
@@ -19,10 +19,7 @@ module MsRestAzure
     #
     # @return [ActiveDirectoryServiceSettings] settings required for authentication.
     def self.get_azure_settings
-      settings = ActiveDirectoryServiceSettings.new
-      settings.authentication_endpoint = 'https://login.windows.net/'
-      settings.token_audience = 'https://management.core.windows.net/'
-      settings
+      get_settings(MsRestAzure::AzureEnvironments::Azure)
     end
 
     #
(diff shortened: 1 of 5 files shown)


azurecla:

TTYL, AZPRBOT;

src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/azure_environment.rb
```diff
@@ -0,0 +1,163 @@
...
+      attr_reader :active_directory_resource_id
+
+      # @return [String] the Active Directory resource ID
+      attr_reader :active_directory_graph_resource_id
+
+      # @return [String] the Active Directory resource ID
+      attr_reader :active_directory_graph_api_version
+
+      # @return [String] the endpoint suffix for storage accounts
+      attr_reader :storage_endpoing_suffix
```
veronicagg: typo: endpoint

veronicagg:
@vishrutshah Left a minor comment, rest looks good to me. Per our conversation yesterday, we should update https://github.com/Azure/autorest/blob/master/src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/active_directory_service_settings.rb as well. Thanks!

src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/azure_environment.rb
vishrutshah: Thanks for catching. Corrected.

vishrutshah:
Yup. Updated !!

src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/active_directory_service_settings.rb
```diff
@@ -19,20 +19,44 @@ class ActiveDirectoryServiceSettings
     #
     # @return [ActiveDirectoryServiceSettings] settings required for authentication.
     def self.get_azure_settings
-      settings = ActiveDirectoryServiceSettings.new
-      settings.authentication_endpoint = 'https://login.windows.net/'
```
veronicagg: This value seems to be different with the new settings:  'https://login.microsoftonline.com/', which one should it be?

vishrutshah: Yes https://login.microsoftonline.com/ one is new endpoint that we should use. The one that we were using (https://login.windows.net/) is the old endpoint and as per my conversation with @amarzavery it's for backward compatibility.

veronicagg:
LGTM!

vishrutshah:
@veronicagg Thanks for the review! As per our offline talk, i have used the old endpoint for now. I'll make another PR to be merged in just before releasing ruby runtime to give us time to fix recordings.

PR #1335, opened 2016-08-08
[In Review] Fix for inconsistency produced in generated code when operation is no…

veronicagg:
Small change to address inconsistency in generated code when operation is not marked as pageable and therefore does not contain pageable methods, even if return type looks pageable. Without this change we make a call to a non-existent "as_lazy" method.

### DIFF
--- a/src/generator/AutoRest.Ruby.Azure/AzureRubyCodeGenerator.cs
+++ b/src/generator/AutoRest.Ruby.Azure/AzureRubyCodeGenerator.cs
@@ -92,15 +92,14 @@ namespace AutoRest.Ruby.Azure
                 {
                     PageableExtension pageableExtension = JsonConvert.DeserializeObject<PageableExtension>(method.Extensions[AzureExtensions.PageableExtension].ToString());
                     if (pageableExtension != null && !method.Extensions.ContainsKey("nextLinkMethod") && !string.IsNullOrWhiteSpace(pageableExtension.NextLinkName))
-                    {
-                        
-                        if (serviceClient.Methods[i].Extensions.ContainsKey("nextMethodName"))
+                    {            
+                        if (method.Extensions.ContainsKey("nextMethodName"))
                         {
-                            serviceClient.Methods[i].Extensions["nextMethodName"] = CodeNamer.GetMethodName((string)method.Extensions["nextMethodName"]);
+                            method.Extensions["nextMethodName"] = CodeNamer.GetMethodName((string)method.Extensions["nextMethodName"]);
                         }
                         else if (!string.IsNullOrWhiteSpace(pageableExtension.OperationName))
                         {
-                            serviceClient.Methods[i].Extensions["nextMethodName"] = CodeNamer.GetMethodName(pageableExtension.OperationName);
+                            method.Extensions["nextMethodName"] = CodeNamer.GetMethodName(pageableExtension.OperationName);
                         }
                         if (!method.Extensions.ContainsKey(AzureExtensions.LongRunningExtension))
                         {
--- a/src/generator/AutoRest.Ruby.Azure/AzureRubyCodeGenerator.cs
+++ b/src/generator/AutoRest.Ruby.Azure/AzureRubyCodeGenerator.cs
@@ -93,14 +93,25 @@ namespace AutoRest.Ruby.Azure
                     PageableExtension pageableExtension = JsonConvert.DeserializeObject<PageableExtension>(method.Extensions[AzureExtensions.PageableExtension].ToString());
                     if (pageableExtension != null && !method.Extensions.ContainsKey("nextLinkMethod") && !string.IsNullOrWhiteSpace(pageableExtension.NextLinkName))
                     {
-                        serviceClient.Methods.Insert(i, (Method)method.Clone());
+                        
                         if (serviceClient.Methods[i].Extensions.ContainsKey("nextMethodName"))
                         {
                             serviceClient.Methods[i].Extensions["nextMethodName"] = CodeNamer.GetMethodName((string)method.Extensions["nextMethodName"]);
                         }
-                        i++;
+                        else if (!string.IsNullOrWhiteSpace(pageableExtension.OperationName))
+                        {
+                            serviceClient.Methods[i].Extensions["nextMethodName"] = CodeNamer.GetMethodName(pageableExtension.OperationName);
+                        }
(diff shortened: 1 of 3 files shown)


vishrutshah:
:shipit:

src/generator/AutoRest.Ruby.Azure/AzureRubyCodeGenerator.cs
```diff
@@ -93,14 +93,25 @@ private void AddRubyPageableMethod(ServiceClient serviceClient)
                     PageableExtension pageableExtension = JsonConvert.DeserializeObject<PageableExtension>(method.Extensions[AzureExtensions.PageableExtension].ToString());
                     if (pageableExtension != null && !method.Extensions.ContainsKey("nextLinkMethod") && !string.IsNullOrWhiteSpace(pageableExtension.NextLinkName))
                     {
-                        serviceClient.Methods.Insert(i, (Method)method.Clone());
+                        
                         if (serviceClient.Methods[i].Extensions.ContainsKey("nextMethodName"))
```
vishrutshah: >serviceClient.Methods[i] [](start = 28, length = 24)

should we be using `method` instead of `serviceClient.Methods[i]`

src/generator/AutoRest.Ruby.Azure/AzureRubyCodeGenerator.cs
```diff
@@ -93,14 +93,25 @@ private void AddRubyPageableMethod(ServiceClient serviceClient)
                     PageableExtension pageableExtension = JsonConvert.DeserializeObject<PageableExtension>(method.Extensions[AzureExtensions.PageableExtension].ToString());
                     if (pageableExtension != null && !method.Extensions.ContainsKey("nextLinkMethod") && !string.IsNullOrWhiteSpace(pageableExtension.NextLinkName))
```
vishrutshah: >!method.Extensions.ContainsKey("nextLinkMethod") [](start = 53, length = 48)

should we be fine removing `!method.Extensions.ContainsKey("nextLinkMethod")` condition?

vishrutshah:
>                 {

In general, It'd be great simplify this with two paths 1. Has PageableExtension then do something and 2. Has LongRunningExtension do something. What do you think about that?

---
Refers to: src/generator/AutoRest.Ruby.Azure/AzureRubyCodeGenerator.cs:92 in a3429f8. [](commit_id = a3429f8c9f1be81081c46ac8b9e556a68a0f9af7, deletion_comment = False)

src/generator/AutoRest.Ruby.Azure/TemplateModels/AzureMethodTemplateModel.cs
```diff
@@ -273,7 +273,8 @@ public override string OperationReturnTypeString
         {
             get
             {
-                if (Extensions.ContainsKey("nextMethodName") && !Extensions.ContainsKey(AzureExtensions.PageableExtension))
+                if (Extensions.ContainsKey("nextMethodName") && (!Extensions.ContainsKey(AzureExtensions.PageableExtension) ||
+                    (Extensions.ContainsKey(AzureExtensions.PageableExtension) && Extensions.ContainsKey(AzureExtensions.LongRunningExtension))))
```
vishrutshah: looks like we don't need this `(!Extensions.ContainsKey(AzureExtensions.PageableExtension) || (Extensions.ContainsKey(AzureExtensions.PageableExtension)`

src/generator/AutoRest.Ruby.Azure/TemplateModels/AzureMethodTemplateModel.cs
```diff
@@ -119,7 +119,7 @@ public override string ResponseGeneration()
                 if (ReturnType.Body is CompositeType)
                 {
                     CompositeType compositeType = (CompositeType)ReturnType.Body;
-                    if (compositeType.Extensions.ContainsKey(AzureExtensions.PageableExtension))
+                    if (compositeType.Extensions.ContainsKey(AzureExtensions.PageableExtension) && this.Extensions.ContainsKey("nextMethodName"))
```
vishrutshah: >&& this.Extensions.ContainsKey("nextMethodName") [](start = 96, length = 48)

From line 124, looks like we don't need `&& this.Extensions.ContainsKey("nextMethodName")` condition or we don't need line 124

src/generator/AutoRest.Ruby.Azure/TemplateModels/AzureMethodTemplateModel.cs
veronicagg: there's a parenthesis enclosing the last two conditions, so this is needed.

---
In reply to: [74297529](https://github.com/Azure/autorest/pull/1335#discussion_r74297529) [](ancestors = 74297529)

src/generator/AutoRest.Ruby.Azure/AzureRubyCodeGenerator.cs
veronicagg: nextLinkMethod, indicates whether the method is the "next" method, nextMethodName is the name of the next method for a pageable operation, so they refer to difference things.

---
In reply to: [74296420](https://github.com/Azure/autorest/pull/1335#discussion_r74296420) [](ancestors = 74296420)

src/generator/AutoRest.Ruby.Azure/TemplateModels/AzureMethodTemplateModel.cs
veronicagg: similar to the other comment, "nextMethodName" is different form "nextLinkMethod"

---
In reply to: [74298007](https://github.com/Azure/autorest/pull/1335#discussion_r74298007) [](ancestors = 74298007)

veronicagg:
>                 {

In this case, I actually need to special case for when pageable and longrunning extensions are used together, in order to avoid 2 copies of the primary method.

---
In reply to: [238950237](https://github.com/Azure/autorest/pull/1335#issuecomment-238950237) [](ancestors = 238950237)

---
Refers to: src/generator/AutoRest.Ruby.Azure/AzureRubyCodeGenerator.cs:92 in a3429f8. [](commit_id = a3429f8c9f1be81081c46ac8b9e556a68a0f9af7, deletion_comment = False)

vishrutshah:
@veronicagg LGTM!

PR #1341, opened 2016-08-09
Updating runtime authentication library due to signing issue

markcowl:
This is solely about updating the version so we can sign and publish the package.


fearthecowboy:
Closing this PR -- we're moving client runtime to sdk-for-net.

Issue #919, opened 2016-04-05
Node.js sample results in "RangeError: Maximum call stack size exceeded"


mcoimbrac:
Hi all.

I've run into an Exception using the node.js sample provided on the 0.15.0 release, which I think is a bug.

It happens on a recursive loop function, using setImmediate, process.nextTick or a setTimeout wrapper.

The following scripts are simple ways of replicating my problem. This script:
```
var petStoreApi = require('./petStoreSampleApi.js');
var msRest = require('ms-rest');
var credentials = new msRest.TokenCredentials("special-key");
var petStoreClient = new petStoreApi(credentials);

var petId = 1;

function loop( i, callback ) {
    if( i < 10000 ) {
        petStoreClient.getPetById(petId, function(err, result, httpReq, response) {
            console.log(i, new Date());
            process.nextTick( function() {
                loop( i+1, resume ); 
            });
        });
    } else {
        callback();
    }
}

console.log('start at ', new Date());
loop( 0, function() {
    console.log('end at ', new Date());
});
```
results in:
```
marcelo@local:~/opt/autorest$ node getPetById-github-issue-code-used.js 
start at  Tue Apr 05 2016 14:26:33 GMT-0300 (BRT)
0 Tue Apr 05 2016 14:26:34 GMT-0300 (BRT)
1 Tue Apr 05 2016 14:26:34 GMT-0300 (BRT)
(...)
4951 Tue Apr 05 2016 15:02:21 GMT-0300 (BRT)
4952 Tue Apr 05 2016 15:02:22 GMT-0300 (BRT)
url.js:426
  if (search && search.charAt(0) !== '?') search = '?' + search;
                                                       ^
RangeError: Maximum call stack size exceeded
    at String.replace (native)
    at Url.format (url.js:428:23)
    at Url.parse (url.js:365:20)
    at Object.urlParse [as parse] (url.js:104:5)
    at Request.init (/opt/autorest/node_modules/ms-rest/node_modules/request/request.js:250:20)
    at new Request (/opt/autorest/node_modules/ms-rest/node_modules/request/request.js:141:8)
    at request (/opt/autorest/node_modules/ms-rest/node_modules/request/index.js:55:10)
    at /opt/autorest/node_modules/ms-rest/node_modules/request/index.js:101:12
    at /opt/autorest/node_modules/ms-rest/node_modules/request/index.js:101:12
    at /opt/autorest/node_modules/ms-rest/node_modules/request/index.js:101:12
```

[...]

PR #1342, opened 2016-08-10
Fix Node.js Infinite Request Wrap Stack Overflow

sedouard:
Hey there. There isn't a contributing document so I'm not sure what specific information you need.

We're hitting [this issue](https://github.com/Azure/autorest/issues/919) A LOT and its causing big problems as it basically causes an unrecoverable error in our backend workers. This is on the latest version of `ms-rest`.

The issue happens when you call azure node.js sdk calls enough (think within the thousands of requests). Because you are wrapping the global `request` module within this module every time a client calls `requestLibrarySink` it adds an [additional layer on the stack](https://github.com/request/request/blob/v2.72.0/index.js#L84).

So you should just use a new variable every time you call `request.defaults`. This prevents the stack from getting too big.

### DIFF
--- a/src/client/NodeJS/ms-rest/lib/requestPipeline.js
+++ b/src/client/NodeJS/ms-rest/lib/requestPipeline.js
@@ -83,11 +83,11 @@ exports.createWithSink = function(sink) {
 exports.requestLibrarySink = function (requestOptions) {
 
   return function (options, callback) {
-    request = request.defaults(requestOptions);
+    var defaultRequest = request.defaults(requestOptions);
     var requestStream;
     var bodyStream;
     if (options.headersOnly) {
-      var requestHeaderStream = request(options);
+      var requestHeaderStream = defaultRequest(options);
       requestHeaderStream.on('error', function (err) {
         return callback(err);
       });
@@ -101,9 +101,9 @@ exports.requestLibrarySink = function (requestOptions) {
       if (options.body && typeof options.body.pipe === 'function') {
         bodyStream = options.body;
         options.body = null;
-        requestStream = bodyStream.pipe(request(options));
+        requestStream = bodyStream.pipe(defaultRequest(options));
       } else {
-        requestStream = request(options);
+        requestStream = defaultRequest(options);
       }
       requestStream.on('error', function (err) {
         return callback(err);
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@sedouard__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (sedouard). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

sedouard:
Here's an easy reproduction of the issue, completely independent of `ms-rest`:

```js
'use strict';
var request = require('request');
for(let i = 0; i < 100000; i++) {
request = request.defaults({});
}
// calling request leads into an SO because of too many wraps!
request({url: 'http://http.cat'}, (err, res) => { console.log(res);})
```

sedouard:
Can I get a an `eta` for how long it would be to get this out on npm with the nodejs sdks updated to the latest `ms-rest` version? This package is pretty far down the stack so I'm going to assume it will take a long time.

PS: I'm not employed by msft anymore so someone should check the CLA bot synchronization to user github orgs.

devigned:
@amarzavery and @haocs please give this some attention.

@sedouard thank you for the PR!

amarzavery:
Hi @sedouard - I had a hard time debugging this. The fix seems logical. Thanks for providing a fix for this. We will take this as a high priority item. Test the fix and if everything goes well publish the updated ms-rest package before this weekend.

azurecla:
__@sedouard__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

amarzavery:

amarzavery:
@sedouard - The new runtime packages for ms-rest-azure and ms-rest have been published. Please let me know if you find any issues with it.

sedouard:
Wow that was fast! Thanks @amarzavery! I pushed a private version of ms-rest on my internal npm server but I'll unpublish and refresh my node_modules cache and see if its updated. (should though the sdks are setup to pick up `1.15.0` on npm install.

amarzavery:
Yes, all the individual service libraries are set up to pick the latest version of runtime packages. Hence cleaning up the node_modules folder and doing an npm install again should get you up and running.

amarzavery:
Thanks again for taking the time to investigate and providing a fix for this 💯

PR #1358, opened 2016-08-16
Added Support for multi collectionFormat in Queries for CSharp

### LINKED ISSUE #717: Support "multi" collectionFormat
Please look at the collectionFormat definition and the set of valid values over here https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameter-object 

At bare minimum we need to support multi so that swagger petstore json http://petstore.swagger.io/v2/swagger.json can be run flawlessly for generating samples.

John-Hart:
This fixes #717 for CSharp


tbombach:
LGTM!

PR #1375, opened 2016-08-24
Fix for CSharp portion of Issue #1285 - added support for Deprecated property for operations

John-Hart:


fearthecowboy:
Looks good to me :+1:

PR #1377, opened 2016-08-24
Changing logging to still log validation when client generation fails

tbombach:
Instead of relying on client generation not throwing, we'll now log validation messages after catching


fearthecowboy:
Looks good to me :+1:

PR #1386, opened 2016-08-30
Adding a test to make sure our generated code does not cause collisions

tbombach:
This test would have failed before `usings` directives were removed from generated code. The `HttpOperationResponse` model is a bit silly, but `Action` is a better example. The user has to avoid this model name (for all of the generated languages), simply because it's defined in `System` in C#.

Once we add some more processing to remove the fully qualified names from generated code as much as possible, this test should still pass (since it checks errors and warnings).


src/generator/AutoRest.CSharp.Unit.Tests/NameCollision.cs
```diff
@@ -0,0 +1,70 @@
...
+
+                var result = await Compile(fileSystem);
+
+                // filter the warnings
+                var warnings = result.Messages.Where(
+                    each => each.Severity == DiagnosticSeverity.Warning
+                            && !SuppressWarnings.Contains(each.Id)).ToArray();
+
+                // use this to dump the files to disk for examination 
+                fileSystem.SaveFilesToTemp("name_collision");
```
John-Hart: do you want to save this file every time the test is run?

tbombach: Nope, I don't. Thanks for the catch!

tbombach:
@azuresdkci retest this please!

fearthecowboy:
LGTM :+1:

tbombach:
@azuresdkci retest this please!

PR #1390, opened 2016-08-31
[do not merge]python: surface real error message on LRO operation failure

yugangw-msft:
fix azure/azure-cli#754.
Notes:
1. xplat-cli doesn't have this issue because node runtime appears to check out the server error first. The related code is [here](https://github.com/Azure/autorest/blob/master/src/client/NodeJS/ms-rest-azure/lib/pollingState.js#L141). This PR follows the same logic.
2. Mark it DNM as I submit it for initial evaluation

### DIFF
--- a/src/client/Python/msrestazure/msrestazure/azure_exceptions.py
+++ b/src/client/Python/msrestazure/msrestazure/azure_exceptions.py
@@ -106,24 +106,26 @@ class CloudError(ClientException):
         self.status_code = self.response.status_code
         self.request_id = None
 
-        if error:
-            self.message = error
-            self.error = response
+        default_error_message = error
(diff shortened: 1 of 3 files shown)


yugangw-msft:
/CC:@AnnaTisch, could you take a look and comment it?

annatisch:
@yugangw-msft, I actually just addressed this issue in PR #1392 - could you please check this to see if it meets your requirements?

yugangw-msft:
@annatisch this is great! I will check out

yugangw-msft:
@annatisch your change also works. closing mine

annatisch:
@yugangw-msft  - cool, thanks for checking!

PR #1394, opened 2016-09-01
Making SchemaResolver more flexible

tbombach:
Making `SchemaResolver` usable outside of a `SwaggerModeler`. This will be used in validation rules to unwrap references for a given schema, letting us look at all parameters/definitions, either local or referenced in rules (other PRs to come).

- Adding a constructor to instantiate it without a modeler (which was only used to get the `Modeler.ServiceDefinition`). Useful when we only have a `ServiceDefinition` (which has all the info to dereference)
- Exposing the `FindProperty()` method that finds a property by name in a schema's inheritence chain. Standardizing the naming so that you can call this by schema $ref string or with the actual schema
- Tweaking the method documentation for `FindProperty()`, because references to "ancestor" or "parent" come from the usage of the class in `SwaggerModeler`. The actual logic of the method searches the referenced schema, as well as its ancestors.
- Also adding an extension method to make it easy to get the first ancestor of a given type from the context when writing a validation rule.

### DIFF
--- a/src/modeler/AutoRest.Swagger/SchemaResolver.cs
+++ b/src/modeler/AutoRest.Swagger/SchemaResolver.cs
@@ -16,7 +16,6 @@ namespace AutoRest.Swagger
     public class SchemaResolver : ICloneable
     {
         private const int MaximumReferenceDepth = 40;
-        private readonly SwaggerModeler _modeler;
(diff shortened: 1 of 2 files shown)


fearthecowboy:
LGTM :+1:

PR #1403, opened 2016-09-02
Adding pageable validation rules and tests

tbombach:
Enforced rules:
- Using `x-ms-pageable` requires a 200 response defined
- The value specified by `nextLinkName` property in `x-ms-pageable` must refer to a property that exists

- Adding a base class for Extension rules to make it easy to write rules to validate extensions
- Adding a base class for Pageable rules to make it easy to write rules to validate `x-ms-pageable`


fearthecowboy:
LGTM :+1:

PR #1411, opened 2016-09-08
[Go] Add code for name conflict resolution and added new APIs.

garimakhulbe:
This PR Includes..
- name conflicts resolution related change (go sdk pr: [#394](https://github.com/Azure/azure-sdk-for-go/pull/394))
- add version string (major.minor.patch) as input argument to rake file and used -pv option in AutoRest command line to pass version in Go Generator. Version argument is required for go generator and added exception when set to null while generating sdk. Used version 0.0.0 for testing.  
- New APIs added - ServerManagement, Datalake-analytics/job, Datalake-analytics/account, EventHub.
- Updated to new version - Network, Subscriptions.
- Removed - Container Service. We are still in discussion on this API.


Tools/gulp/gulp-regenerate-expected.js
```diff
@@ -93,6 +93,12 @@ function gulpRegenerateExpected(options, done) {
       args.push(opts.syncMethods);
     }
 
+    //package version
+    if(opts.pv){
```
vishrutshah: minor: needs a space between `)` and `{`. I'd recommend using Document Formatter in VS.

src/generator/AutoRest.Go/Extensions.cs
```diff
@@ -89,6 +91,26 @@ public static string[] ToWords(this string value)
             return SplitPattern.Split(value).Where(s => !string.IsNullOrEmpty(s)).ToArray();
         }
 
+        public static string Singularize(this string value)
```
vishrutshah: Could you please add comments.

src/generator/AutoRest.Go/Extensions.cs
```diff
@@ -89,6 +91,26 @@ public static string[] ToWords(this string value)
...
+        {
+            PluralizationService ps = PluralizationService.CreateService(CultureInfo.GetCultureInfo("en-us"));
+            return ps.Singularize(value);
+        }
+
+        /// <summary>
+        /// This method checks if MethodGroupName is plural of package name.
+        /// if yes, it does not remove package name from group name.
```
vishrutshah: looks like this method only checks so this comments seems appropriate on the above method.

src/generator/AutoRest.Go/GoCodeGenerator.cs
```diff
@@ -60,7 +60,8 @@ public override void NormalizeClientModel(ServiceClient serviceClientModel)
         public override async Task Generate(ServiceClient serviceClient)
         {
             string packageName = GoCodeNamer.PackageNameFromNamespace(Settings.Namespace);
-            
+            string[] version = GoCodeNamer.SDKVersionFromPackageVersion(Settings.PackageVersion);
```
vishrutshah: Version fields should be considered optional so that we do not need it in testing.

src/generator/AutoRest.Go/TemplateModels/VersionTemplateModel.cs
```diff
@@ -15,12 +15,14 @@ public class VersionTemplateModel : ServiceClient
         public static readonly List<string> StandardImports = new List<string> { "fmt" };
 
         public readonly string PackageName;
+        public readonly string[] VersionString;
 
-        public VersionTemplateModel(ServiceClient serviceClient, string packageName)
+        public VersionTemplateModel(ServiceClient serviceClient, string packageName, string[] versionString)
```
vishrutshah: minor: I'd just name `versionString` to `version` as IMO `versionString` misleading name for string array.

src/generator/AutoRest.Go/rakefile.rb
```diff
@@ -196,8 +200,9 @@ def to_services(m, h, *p)
 def generate(service)
 	s = "Generating #{service.fullname}.#{service.version} "
 	puts "#{s} #{"=" * (80 - s.length)}"
+
 	delete(service)
-	s = `#{AUTOREST} -AddCredentials -CodeGenerator Go -Header MICROSOFT_APACHE -Input #{service.input_path} -Namespace #{service.namespace} -OutputDirectory #{service.output_path} -Modeler Swagger`
+	s = `#{AUTOREST} -AddCredentials -CodeGenerator Go -Header MICROSOFT_APACHE -Input #{service.input_path} -Namespace #{service.namespace} -OutputDirectory #{service.output_path} -Modeler Swagger -pv #{ENV['VERSION']}`
```
vishrutshah: Can you use `SDK_VERSION` variable in the command instead of accessing environment variable again?

src/generator/AutoRest.Go/rakefile.rb
```diff
@@ -13,18 +16,18 @@
...
 	# containerservice: {version: "2016-03-30", swagger: "containerservice"},
 	# commerce: {version: "2015-06-01-preview"},
-	# datalake_analytics: {
-	# 	account: {version: "2015-10-01-preview"},
-	# 	catalog: {version: "2015-10-01-preview"},
-	# 	job:  {version: "2016-03-20-preview"}
-	# },
+	datalake_analytics: {
+		account: {version: "2015-10-01-preview"},
+		# catalog: {version: "2016-06-01-preview"},
```
vishrutshah: job and catalog are not azure resource manager api of DataLake. They belong to the data plane of the DataLake resource provider.

src/generator/AutoRest.Go/rakefile.rb
```diff
@@ -13,18 +16,18 @@
...
+		job:  {version: "2016-03-20-preview"}
+	},
 	datalake_store: {
 		account: {version: "2015-10-01-preview"},
 		filesystem: {version: "2015-10-01-preview"}
```
vishrutshah: filesystem are not azure resource manager api of DataLake Store. They belong to the data plane of the DataLake Store resource provider.

vishrutshah:
@garimakhulbe I've left some minor comments. Please have a look and also verify why travis is broken or just re-kick or the build.

mcardosos:
>                 "autorest", "client", "date", "err", "req", "resp", "result", "sender"

"validation" and "to" are missing

---
Refers to: src/generator/AutoRest.Go/GoCodeNamer.cs:216 in 458388f. [](commit_id = 458388fb6c236b9ab003115cf4d0589ea0949e1d, deletion_comment = False)

src/generator/AutoRest.Go/rakefile.rb
garimakhulbe: These are in arm in [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs) and in [.net sdk](https://github.com/Azure/azure-sdk-for-net/tree/AutoRest/src/ResourceManagement/DataLake.Store/Microsoft.Azure.Management.DataLake.Store/Generated) repos.

src/generator/AutoRest.Go/rakefile.rb
garimakhulbe: These are in arm in azure-rest-api-specs and in .net sdk repos.

src/generator/AutoRest.Go/GoCodeGenerator.cs
garimakhulbe: If we make it optional then we need to put some default value in version like 0.0.0. There is a chance if we do that we might release SDK with wrong default version so I made it required to be sure what version we pass the version all the time. But if you still want I can make this optional?

fearthecowboy:
What's the status here? Close or Merge? :dart:

mcardosos:
@garimakhulbe Ping

garimakhulbe:
Will update for comments by today eod and probably merge by tomorrow.

garimakhulbe:
@fearthecowboy saw your message on channel on freezing code for development on core. I will merge this PR by today.

Tools/gulp/gulp-regenerate-expected.js
garimakhulbe: version is optional now so this change is removed.

src/generator/AutoRest.Go/Extensions.cs
garimakhulbe: This method is removed. Not using now.

src/generator/AutoRest.Go/Extensions.cs
garimakhulbe: This is updated.

src/generator/AutoRest.Go/GoCodeGenerator.cs
garimakhulbe: Version is optional now and 0.0.0 for tests.

src/generator/AutoRest.Go/TemplateModels/VersionTemplateModel.cs
garimakhulbe: This is changed to version.

src/generator/AutoRest.Go/rakefile.rb
garimakhulbe: This is updated to SDK_Version.

src/generator/AutoRest.Go/rakefile.rb
garimakhulbe: Not changed.

src/generator/AutoRest.Go/rakefile.rb
garimakhulbe: Not changed.

garimakhulbe:
@vishrutshah @mcardosos I have updated PR for review comments. Please check. We need to merge this today as we are freezing code for autorest core changes.

- Updated the version to 5.0.0 since we removed container service from sdk.
- Reverted resources/subscription to older version for now because new swagger has `SubscriptionState` enum type in package `Subscriptions`. Since we are trimming package name, `SubscriptionState` becomes `tate` which incorrect. If we don't remove it then all types will have package which result in golint error for stuttering types. We need discussion on how we are going to handle different name scenarios so not updating it in this PR.

mcardosos:
LGTM, as soon as Travis build is fixed, I'm ok if this PR gets merged :3

src/generator/AutoRest.Go/rakefile.rb
vishrutshah: As per my talk with Ben Goldsmith, They do not wanted to include Data Plane services into ARM namespace intentionally but they did for .net and nodeJS. Now they cannot remove it with reasons on break their existing customer and for backward compatibilities. 

Go language still have a chance to avoid this case by not including them in the ARM namespace. I'd still prefer not to include those in the ARM namespace today or we'll be stuck in the same situation. We should only generate management planes for Store and Analytics services. What are your views on this?

---
In reply to: [79500366](https://github.com/Azure/autorest/pull/1411#discussion_r79500366) [](ancestors = 79500366)

src/generator/AutoRest.Go/TemplateModels/VersionTemplateModel.cs
```diff
@@ -15,12 +15,14 @@ public class VersionTemplateModel : ServiceClient
         public static readonly List<string> StandardImports = new List<string> { "fmt" };
 
         public readonly string PackageName;
+        public readonly string[] Version;
```
vishrutshah: minor: any reason for creating own structure for version instead of using the System.Version to provide this functionalities? [https://msdn.microsoft.com/en-us/library/system.version(v=vs.110).aspx](https://msdn.microsoft.com/en-us/library/system.version(v=vs.110).aspx)

src/generator/AutoRest.Go/rakefile.rb
mcardosos: Maybe those could have their own folders just like storage data plane has?

vishrutshah:
@garimakhulbe Changes looks good. Please make sure travis passes otherwise this will break all subsequent PR's into master. Thanks!

garimakhulbe:
I am not sure whats going on here. Windows build is fine but Linux is failing with error....

`
 /home/travis/build/Azure/autorest/src/generator/AutoRest.Go/project.json(25,39): error NU1001: The dependency System.Data.Entity.Design could not be resolved. (from travis logs)
`

System.Data.Entity.Design is new dependency I added to go generator. I tried to add dependency in project.json through following ways but nothing works for linux.

`
 "frameworks": {
    "net451": {
      "frameworkAssemblies": {
        "System.Data.Entity.Design": "4.0.0.0"
      }
    }
`
or
`
 "frameworks": {
    "net451": {
      "frameworkAssemblies": {
        "System.Data.Entity.Design": ""
      }
    }
`

Can anybody help?

vishrutshah:
@tbombach / @fearthecowboy Do you guys have any idea or pointer for @garimakhulbe on the issue with Travis Linux builds while adding new `System.Data.Entity.Design ` dependency?

jhendrixMSFT:
@garimakhulbe This is a .NET desktop assembly and there doesn't appear to be a .NET Core equivalent (well I searched around a bit and couldn't find one but perhaps I missed it).  Mono does appear to have support for this functionality but I don't know what it would take to wire into the build (is this really how we'd want to fix it though).  Can you find a .NET Core equivalent?

tbombach:
@garimakhulbe, @jhendrixMSFT Is correct: https://apisof.net/catalog/System.Data.Entity.Design. I believe you included `System.Data.Entity.Data` to use `PluralizationService` - it looks like this functionality is removed from .NET Core compatible Entity Framework: https://github.com/aspnet/EntityFramework/issues/2506#issuecomment-127315643

Do you need to use the pluralization service? It seems like you use it to know whether or not to add the package name to the name of the client - would you mind explaining the desired behavior a bit?

garimakhulbe:
I have removed dependency and fixed it by workaround. @jhendrixMSFT please take a look. 

@tbombach the issue is we have a package `EventHub` which has methodgroup `EventHubs`. When we remove package name from group name, it becomes `EventHubs -> s` which is incorrect. For that, I keep group name as it is if its plural of package.
Some context in this. We remove package name for methodgroup name to avoid stuttering type errors in go. You will find more about it [here](https://blog.golang.org/package-names). It is if type has package name in it, then go gives golint error. For example...
`network.NetworkInterfacesClient  // error - package name repeat`
`network.InterfacesClient               // good name`


Build is failing for ruby install. It seems `https://rubies.travis-ci.org/ubuntu/14.04/x86_64/ruby-2.3.0.tar.bz2` is unreachable at the moment. I will try again in sometime or if there is some other issue let me know.

jhendrixMSFT:
@garimakhulbe Your change looks reasonable however I would suggest updating the comment on IsNamePlural() to note that the function will return false for names it doesn't recognize.

garimakhulbe:
All checks passed. @jhendrixMSFT Can we merge?

jhendrixMSFT:
LGTM.  Side note: if the pluralization issue becomes systemic we should look at copying the implementation from EF6 (it's MIT license).

PR #1413, opened 2016-09-08
Using SchemaResolver to follow inheritance chain for validation rules

tbombach:
- Modifying `RequiredPropertiesMustExist` rule to follow allOf references
- Modifying `NextLinkMustExist` rule to follow allOf references
- Adding positive tests to make sure validation rules don't show when they shouldn't


tbombach:
Addressing https://github.com/Azure/autorest/issues/1383

PR #1419, opened 2016-09-12
Supporting Graph.RBAC scenario with fragment nextLink in C#

tbombach:
Since the `nextLink` provided by Graph.RBAC is not a full URL, it's modeled with x-ms-paths. Since that URL has query parameters, our C# logic creates a URL with two `"?"` characters, instead of correctly appending extra parameters with `"&"`.

- Modifying the C# template code to add this check when constructing the url for operations
- Adding a test that mimics the fragment scenario and modifying C# code to test this scenario
- Regenerating expected

Thanks to Anna for making the test closely repro the Graph scenario

### DIFF
--- a/Samples/azure-storage/Azure.CSharp/StorageAccountsOperations.cs
+++ b/Samples/azure-storage/Azure.CSharp/StorageAccountsOperations.cs
@@ -100,7 +100,7 @@ namespace Petstore
             }
             if (_queryParameters.Count > 0)
             {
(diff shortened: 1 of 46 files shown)


tbombach:
Update: reverted removal of Python files that occurred during first regenerate

fearthecowboy:
LGTM :+1:

tbombach:
Merged!

PR #1434, opened 2016-09-16
Fixing TestResults dir path

dsgouda:
Creating the TestResults directory under root dir like build.proj does

@fearthecowboy


tbombach:
@azuresdkci add to whitelist

tbombach:
LGTM

PR #1438, opened 2016-09-19
Supporting relative $ref paths in Swagger files

tbombach:
Currently, a $ref that refers to an external file with a relative path will try to resolve from the directory of AutoRest.exe, not the input Swagger file. This PR changes this behavior so that the files resolve from the input file.

- Re-using code that correctly resolves relative composite swagger file document paths
- Refactoring this code to put it in Settings and FileSystem.cs (see commit #2)
- Also fixing existing tests that validate the current (wrong) behavior

Note: This PR is a step in the right direction. It doesn't yet support multiple levels of $ref relative paths (e.g. Input spec has an absolute ref to spec A on the internet -> Spec A has a relative $ref to spec B -> This logic will try to use the folder of the Input spec file to fully resolve the $ref to spec B). This is because we don't store the file path of a file once we deserialize it. Therefore we don't have that path when we're resolving external $refs. I'm submitting this PR since it supports the most common scenario with relative $ref and it gets us closer to solving the more complicated scenario.

### DIFF
--- a/src/core/AutoRest.Core/Settings.cs
+++ b/src/core/AutoRest.Core/Settings.cs
@@ -61,6 +61,30 @@ Licensed under the MIT License. See License.txt in the project root for license 
         /// </summary>
         public IFileSystem FileSystem { get; set; }
 
+        private Uri _inputFolder = null;
+        /// <summary>
+        /// Gets the Uri for the path to the folder that contains the input specification file.
+        /// </summary>
+        public Uri InputFolder
+        {
+            get
+            {
+                if (_inputFolder == null)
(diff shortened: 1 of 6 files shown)


tbombach:
CC: @amarzavery: Let me know if you have any thoughts on this PR. I moved the `basepath` logic to a property to `Settings` and the logic that adds a relative path to the input folder into `FileSystem`. That way the same code is reused for CompositeSwagger document resolution and reference file path resolution.

fearthecowboy:
LGTM 👍

PR #1444, opened 2016-09-20
Refactor code model

fearthecowboy:
## Welcome to the great Code Model Refactor!

**[Still editing this doc]**

The overall purpose of this rather large refactoring is to make the whole `CodeModel` (aka, `ServiceClient`) smarter, more consistent, and drive behavior into the code model classes, and remove unnecessary mutability from the model. (a small example: if a property was using an `EnumType` as nullable, then it modified the actual `EnumType` to be nullable, and every use of that would be.)

There were also these huge 'normalization' steps which would go thru and rewrite parts of the model. This made it kinda tricky to handle certain kinds of behavior, and track down where stuff went...

After the refactoring, behavior is delegated to the actual type (so, rather than have a `normalize` method change the output name of a `PrimaryType`), that name is simply overrideable in the subclass directly. 
Everything now has a ref to their parent, (so a `parameter` can walk back up to the `method`, the `methodgroup` or even up to the top of the model now)

Plus, all 'transformation' of a model are now handled in discrete steps (and can be spit out to JSON and brought back in between steps.)

The foundation for a pluggable transformer model is in here too

Initially, this was not going to be such a radical change, but in order to accomplish certain things, one thing led to another, and, well, this gets us to a really nice spot.

### Dependency Injection System
With this patch, I've introduced the the `Least Offensive Dependency Injection System` :tm: (aka `LODIS`)
 See: https://github.com/Azure/autorest/wiki/Least-Offensive-Dependency-Injection-System

### `Fixable` values
The notion of values which the final value can be both derived or a fixed literal value: https://github.com/Azure/autorest/wiki/Fixable-T----When-a-value-is-both-calculated-and-or-fixed.

### Generator/Transformer/Namer model

Basics of the new Code Model: https://github.com/Azure/autorest/wiki/CodeModel-and-the-Language-specific-Generator-Transformer-Namer 

### Name Disambiguation

Learn a bit about the name disambiguation here: https://github.com/Azure/autorest/wiki/Name-Disambiguation


src/core/AutoRest.Core/CodeModelTransformer.cs
```diff
@@ -0,0 +1,105 @@
...
+            return new Context();
+        }
+
+        public IDisposable Activate() => _context.Activate();
+
+        public virtual CodeModel Load(string jsonText)
+        {
+            using (_context.Activate())
+            {
+                try { File.WriteAllText(@"c:\tmp\orig.json",jsonText); } catch { }
```
tbombach: Non-windows? Maybe use Path.GetTempPath?

src/core/AutoRest.Core/Model/IChild.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+//
```
tbombach: Very minor: a few of these files have extra comment lines in the header

src/core/AutoRest.Core/Model/IModelType.cs
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        public virtual string ClassName => Name.Value;
+
+        public virtual string DeclarationName => Name.Value;
+
+        /// <summary>
+        /// Allows a type to append extra documentation on a property
+        /// </summary>
+        public virtual string ExtendedDocumentation => null;
```
tbombach: I think ExtendedDocumentation is another one that we'll want fixable eventually, but I can't think of the use case right now. I imagine we'll convert them on an as-needed basis?

src/core/AutoRest.Core/CodeGenerator.cs
```diff
@@ -61,19 +69,25 @@ public virtual void PopulateSettings(IDictionary<string, object> settings)
...
+        }
 
         /// <summary>
         /// Generates code and outputs it in the file system.
         /// </summary>
-        /// <param name="serviceClient"></param>
+        /// <param name="codeModel"></param>
         /// <returns></returns>
-        public abstract Task Generate(ServiceClient serviceClient);
+        public virtual /* async */ Task Generate(CodeModel codeModel)
```
fearthecowboy: Yes

src/core/AutoRest.Core/Model/IModelType.cs
fearthecowboy: Yeah, we may make a lot of them fixable in the long run. First step was just to get it working, refinements can be made along the way.

fearthecowboy:
Merging into branch -- will promote to master once the other generators are converted

PR #1455, opened 2016-09-27
[Ruby] Fix dual credentials on Ruby Service Clients

vishrutshah:
_Problem_: https://github.com/Azure/azure-sdk-for-ruby/issues/479
_Solutions_: Based on the fix from [NodeJS ](https://github.com/Azure/autorest/blob/master/src/generator/AutoRest.NodeJS/NodeJSCodeGenerator.cs#L76) and [Python ](https://github.com/Azure/autorest/blob/master/src/generator/AutoRest.Python/PythonCodeGenerator.cs#L81) language generator we should only populate credentials if they are not already added.

### DIFF
--- a/Samples/azure-storage/Azure.Ruby/generated/azure_storage/storage_management_client.rb
+++ b/Samples/azure-storage/Azure.Ruby/generated/azure_storage/storage_management_client.rb
@@ -33,10 +33,6 @@ module Petstore
     # is generated and included in each request. Default is true.
     attr_accessor :generate_client_request_id
 
-    # @return Subscription credentials which uniquely identify client
-    # subscription.
-    attr_accessor :credentials
-
     # @return [StorageAccounts] storage_accounts
     attr_reader :storage_accounts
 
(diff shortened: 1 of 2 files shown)


vishrutshah:
@sarangan12 & @veronicagg Feel free to review this PR when you get a chance. Thanks!

veronicagg:
Only sample change that's included in the PR is azure-storage, was there also a change in petstore?

vishrutshah:
I've regened samples using `regenerate:expected:samples` but i didn't see any other changes but will double check. 

I am also looking into from where the duplication on credentials is occurring :)

vishrutshah:
@veronicagg I've found from where we are having the dual credentials. 

When we call [AzureExtensions.NormalizeAzureClientModel(serviceClient, Settings, CodeNamer);](https://github.com/Azure/autorest/blob/master/src/generator/AutoRest.Ruby.Azure/AzureRubyCodeGenerator.cs#L70), it indirectly calls [AddAzureProperties(serviceClient);](https://github.com/Azure/autorest/blob/master/src/core/AutoRest.Extensions.Azure/AzureExtensions.cs#L86) which adds Azure specific properties like `subscriptionId`, `acceptLanguage `, `LongRunningOperationRetryTimeout` along with `credentials`.

Let me know if you think we could have another route that can neatly fix this. Thanks!

vishrutshah:
Thanks for the review!!

PR #1457, opened 2016-09-27
Fix errant OOM on HEAD REQUEST.

fearthecowboy:
On Head Request, always use `HttpCompletionOption.ResponseHeadersRead` on SendAsync


tbombach:
LGTM :+1:

PR #1459, opened 2016-09-27
Generating C# project files based on new -Project flag

tbombach:
- New flag -Project, which controls whether or not project files are generated
- Adding a FileTemplate (outputs a bare file), an XProjTemplate (outputs an XProj) and ProjectJson/AzureProjectJson templates (might be merged into one template later depending on how much they differ)
- Adding a step in CSharpCodeGenerator and AzureCSharpCodeGenerator that outputs project.json and .xproj files for the generated code if the flag is included

Still todo: 
- Be smarter about overwriting existing project files - this code does not yet update an existing file with the correct info (e.g. ClientRuntime reference)
- Generate a test project for the Azure generator

### DIFF
--- a/src/core/AutoRest.Core/FileTemplate.cs
+++ b/src/core/AutoRest.Core/FileTemplate.cs
@@ -1,0 +1,30 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace AutoRest.Core
+{
+#line 1 "FileTemplate.cshtml"
+
+#line default
+#line hidden
+    using System.Threading.Tasks;
+
+    public class FileTemplate : Template<string>
+    {
+        #line hidden
+        public FileTemplate()
+        {
+        }
+
+        #pragma warning disable 1998
+        public override async Task ExecuteAsync()
+        {
+#line 3 "FileTemplate.cshtml"
+Write(Model);
+
+#line default
+#line hidden
+        }
+        #pragma warning restore 1998
+    }
+}
--- a/src/core/AutoRest.Core/Settings.cs
+++ b/src/core/AutoRest.Core/Settings.cs
@@ -54,6 +54,7 @@ Licensed under the MIT License. See License.txt in the project root for license 
             Modeler = "Swagger";
             ValidationLevel = LogEntrySeverity.Error;
             ModelsName = "Models";
+            Project = false;
         }
 
         /// <summary>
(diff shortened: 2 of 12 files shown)


src/core/AutoRest.Core/FileTemplate.cs
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
fearthecowboy: I thought we didn't check in the generated `.cs` files for `.cshtml` files since they are generated at build time?

src/core/AutoRest.Core/Settings.cs
```diff
@@ -275,6 +276,14 @@ public string Header
         public LogEntrySeverity ValidationLevel { get; set; }
 
         /// <summary>
+        /// The input validation severity level that will prevent code generation
+        /// </summary>
```
fearthecowboy: Are we expecting that all the generator should support this eventually?

src/generator/AutoRest.CSharp.Azure/Templates/AzureProjectJsonTemplate.cshtml
```diff
@@ -0,0 +1,53 @@
...
+    "netstandard1.5": {
+      "imports": ["dnxcore50"],
+      "buildOptions": { "define": [ "PORTABLE" ] },
+      "dependencies": {
+        "Microsoft.NETCore.Platforms": "1.0.1",
+        "System.Diagnostics.Tools": "4.0.1",
+        "System.Net.Http": "4.1.0",
+        "System.Runtime.Serialization.Primitives": "4.1.1",
+        "System.Threading.Tasks": "4.0.11",
+        "Microsoft.Rest.ClientRuntime.Azure": "[@Model.ClientRuntimeVersion,4.0.0)"
```
fearthecowboy: What? didn't want to bother calculating the next major version!? awwwwwwww.. . 😼

src/generator/AutoRest.CSharp/Templates/ProjectJsonTemplate.cshtml
```diff
@@ -0,0 +1,50 @@
...
+  "packOptions": {
+    "summary": "@Model.Description",
+    "tags": [],
+    "requireLicenseAcceptance": true
+  },
+
+  "buildOptions": {
+    "delaySign": true,
+    "publicSign": false,
+    "keyFile": "../../../../tools/MSSharedLibKey.snk"
```
fearthecowboy: are you sure we want these in the vanilla c# version? That path is very specific ...

PR #1471, opened 2016-09-30
Dev/daschult/update compute swagger

daschult:
Update computer.json swagger and add containerService.json swagger


tbombach:
Merged

PR #1476, opened 2016-09-30
Add notification hub schema tests

daschult:


src/generator/AutoRest.AzureResourceSchema.Tests/AzureResourceSchemaAcceptanceTests.cs
```diff
@@ -129,14 +129,19 @@ public static void Network_2016_03_30()
         {
             RunSwaggerTest("Network", "2016-03-30", "network.json");
         }
-
-        [Fact]
+
```
tbombach: I think a `[Fact]` annotation is missing here

daschult: Fixed.

tbombach:
Merged

PR #1481, opened 2016-10-03
[Ruby] Operation methods should not include constants in the signature

vishrutshah:
This change excludes usage of constant parameters in the operations documentation comments. 
Example Error: https://github.com/Azure/azure-sdk-for-ruby/pull/481#discussion-diff-81245530R134

### DIFF
--- a/src/generator/AutoRest.Ruby/TemplateModels/MethodTemplateModel.cs
+++ b/src/generator/AutoRest.Ruby/TemplateModels/MethodTemplateModel.cs
@@ -319,29 +319,16 @@ namespace AutoRest.Ruby.TemplateModels
 
         /// <summary>
         /// Get the parameters that are actually method parameters in the order they appear in the method signature
-        /// exclude global parameters
-        /// </summary>
-        public IEnumerable<ParameterTemplateModel> LocalParameters
-        {
-            get
-            {
-                //Omit parameter group parameters for now since AutoRest-Ruby doesn't support them
-                return
-                    ParameterTemplateModels.Where(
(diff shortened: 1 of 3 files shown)


vishrutshah:
@veronicagg & @sarangan12 Feel free to review this PR when you get a chance. Thanks!

src/generator/AutoRest.Ruby/TemplateModels/MethodTemplateModel.cs
```diff
@@ -334,6 +334,18 @@ public IEnumerable<ParameterTemplateModel> LocalParameters
         }
 
         /// <summary>
+        /// Get the parameters that are actually method parameters in the order they appear in the method signature
+        /// exclude global parameters and constants
+        /// </summary>
+        public IEnumerable<ParameterTemplateModel> MethodParameters
```
veronicagg: Are LocalParameters used as is anywhere? or could we modify LocalParameters method instead of creating a new one? 
Any thoughts on the existing comment in the LocalParameters method at https://github.com/Azure/autorest/pull/1481/files#diff-31d7014dc0acde174bbf87a6e0f80747R328 ? I would expect that comment to not be true anymore (not sure how the condition relates to the comment).

vishrutshah: Good catch. 
--Removed `LocalParameters` as it's a dead variable. Kept the name `MethodParamters` for clarity. 
--Also removed the comments as now we are supporting param grouping.

vishrutshah:
Thanks for the review!!

PR #1489, opened 2016-10-04
[Go]Paging tests and general paging fixes

mcardosos:
Paging now doesn't duplicate "NextLink" on structs.
Also, it only generates a "next" method when it is not already defined in swagger. This includes not generating a nextLinkPreparer when it is not needed.


azurecla:
Hi __@mcardosos__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (mcardoso). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

mcardosos:
@jhendrixMSFT @vishrutshah @garimakhulbe
Please take a look :D

src/generator/AutoRest.Go/rakefile.rb
```diff
@@ -9,17 +9,18 @@
 AUTOREST     = "../../core/AutoRest/bin/Debug/net451/win7-x64/AutoRest.exe"
 
 SWAGGER_VERSIONS = {
+	# asazure: {version: "2016-05-06"},
```
vishrutshah: Are you planning to regen the once that you updated in this Rakefile?

mcardosos: I could regen, yes. I updated the rake file just to keep it updated, and add comments on what we are missing.

src/generator/AutoRest.Go/TemplateModels/ModelsTemplateModel.cs
```diff
@@ -15,6 +16,7 @@ public class ModelsTemplateModel : ServiceClient
         public List<ModelTemplateModel> ModelTemplateModels { get; set; }
         public string PackageName { get; set; }
         public Dictionary<IType, string> PagedTypes { get; set; }
+        public List<IType> NextMethodUndefined { get; set; }
```
mcardosos: I added this for paged methods that its next method isn't defined in the swagger. How the generator handles this right now is adding a preparer in the models file, specific for the paged model. I am using this list to keep track of which models would need the preparer, so it is not generated when the next method is already defined. When the next method is defined in the swagger, the generator takes it as any other method and generates its preparer, sender and receiver next to the method and the other group methods.

src/generator/AutoRest.Go/TemplateModels/ModelTemplateModel.cs
```diff
@@ -21,6 +21,8 @@ public class ModelTemplateModel : CompositeType
         // (null or empty if the model is not paged).
         public string NextLink;
 
+        public bool PreparerNeeded;
```
mcardosos: To `false` I think. The general case is that models won't need a preparer (I explaned this better in the `NextMethodUndefined` comment).

src/generator/AutoRest.Go/GoCodeNamer.cs
```diff
@@ -533,6 +533,21 @@ private IType NormalizeDictionaryType(DictionaryType dictionaryType)
             return new MapType(NormalizeTypeReference(dictionaryType.ValueType));
         }
 
+        public static string PascalCaseWithoutChar(string name, char splitter)
```
vishrutshah: minor: should a comments

src/generator/AutoRest.Go/Extensions.cs
```diff
@@ -560,7 +562,13 @@ public static string Fields(this CompositeType compositeType)
                 &&  !String.IsNullOrEmpty((compositeType as ModelTemplateModel).NextLink))
             {
                 var nextLinkField = (compositeType as ModelTemplateModel).NextLink;
-                if (!properties.Any(p => p.Name == nextLinkField))
+                foreach (Property p in properties) {
```
jhendrixMSFT: Can you add a comment explaining why this is necessary.

src/generator/AutoRest.Go/TemplateModels/ServiceClientTemplateModel.cs
```diff
@@ -39,10 +39,9 @@ public ServiceClientTemplateModel(ServiceClient serviceClient, string packageNam
                                              : (MethodGroupName + "Client").TrimPackageName(PackageName);
             MethodScope = new MethodScopeProvider();
             MethodTemplateModels = new List<MethodTemplateModel>();
-
             Methods.Where(m => m.BelongsToGroup(MethodGroupName, PackageName))
-                    .OrderBy(m => m.Name)
-                    .ForEach(m => MethodTemplateModels.Add(new MethodTemplateModel(m, ClientName, PackageName, new MethodScopeProvider())));
+                .OrderBy(m => m.Name)
+                .ForEach(m => MethodTemplateModels.Add(new MethodTemplateModel(m, ClientName, PackageName, new MethodScopeProvider(), m.NextMethodExists(Methods))));
```
garimakhulbe: why you need to pass m.NextMethodExists(Methods) here? Can't you check this inside new MethodTemplateModel() since it will have the information you needed to nextlink?

src/generator/AutoRest.Go/TemplateModels/ModelsTemplateModel.cs
```diff
@@ -113,7 +119,12 @@ public ModelsTemplateModel(ServiceClient serviceClient, string packageName)
                     mtm.IsResponseType = true;
                     if (PagedTypes.ContainsKey(mtm))
                     {
-                        mtm.NextLink = PagedTypes[mtm];
+                        mtm.NextLink = GoCodeNamer.PascalCaseWithoutChar(PagedTypes[mtm], '.');
+                        if (NextMethodUndefined.Contains(mtm)) {
```
garimakhulbe: Not sure what PascalCaseWithoutChar does.

src/generator/AutoRest.Go/TemplateModels/ModelsTemplateModel.cs
```diff
@@ -100,6 +103,9 @@ public ModelsTemplateModel(ServiceClient serviceClient, string packageName)
                     {
                         PagedTypes.Add(m.ReturnValue().Body, m.NextLink());
                     }
+                    if (!m.NextMethodExists(serviceClient.Methods)) {
+                        NextMethodUndefined.Add(m.ReturnValue().Body);
+                    }
```
garimakhulbe: I do not get this change.

src/generator/AutoRest.Go/TemplateModels/ServiceClientTemplateModel.cs
garimakhulbe: ok. Got it. `Method`.

src/generator/AutoRest.Go/TemplateModels/ModelsTemplateModel.cs
garimakhulbe: >> With this changes, the generator no longer generates extra next operations.

don't generate extra preparer only when it is already defined in swagger. else generate always. I think you keep in account both cases. Right?

src/generator/AutoRest.Go/Extensions.cs
```diff
@@ -1022,7 +1048,7 @@ public static string NextLink(this Method method)
                     var nextLinkName = (string)pageableExtension["nextLinkName"];
                     if (!string.IsNullOrEmpty(nextLinkName))
                     {
-                        nextLink = GoCodeNamer.PascalCase(nextLinkName);
+                        nextLink = GoCodeNamer.PascalCaseWithoutChar(nextLinkName, '.');
                     }
```
garimakhulbe: Got it. Closed.

src/generator/AutoRest.Go/TemplateModels/ModelsTemplateModel.cs
mcardosos: It is for `odata.nextlink`
I think I like `PascalCaseWithoutSpecialChars`

src/generator/AutoRest.Go/TemplateModels/ModelsTemplateModel.cs
mcardosos: Yep yep
It is more obvious with these lines in the cshtml files...
[MethodTemplate](https://github.com/mcardosos/autorest/blob/32e755b0dc2f28b79dee2c68e8db9207cfd8f6b6/src/generator/AutoRest.Go/Templates/MethodTemplate.cshtml#L154)
[ModelTemplate]
(https://github.com/mcardosos/autorest/blob/32e755b0dc2f28b79dee2c68e8db9207cfd8f6b6/src/generator/AutoRest.Go/Templates/ModelTemplate.cshtml#L17)

src/generator/AutoRest.Go/TemplateModels/ModelsTemplateModel.cs
mcardosos: Yeah, it is for `odata.nextlink`
It could be `PascalCaseWithoutSpecialChars`

src/generator/AutoRest.Go/Extensions.cs
mcardosos: This foreach removes all dots in properties (for code to work), and if it finds the already existing next link, changes its name to how NextLink was added originally (NextLink is a ModelTemplateModel property and it is used in more places, including the razor tamplates)

mcardosos:
@fearthecowboy Hello :D
Will merging this to master cause problems with refactoring?

fearthecowboy:
LGTM 👍

PR #1494, opened 2016-10-06
[Do Not Merge] Using ruby version 2.2.0 in travis-ci

vishrutshah:


veronicagg:
@vishrutshah should we close this? Is the issue resolved?

vishrutshah:
@veronicagg Yup i should close this as that issues seemed to be transient and not happening now. Thanks for reminder :)

Issue #1467, opened 2016-09-29
NodeJS generator missing type in client .d.ts file 


christav:
When I generate with `-addCredentials true` and `CodeGenerator NodeJS` the generated typescript .d.ts file refers to ServiceClientCredentials, but neither defines the type or nor has an import to find it. In order for this file to compile, it has to be hand-edited to include a definition.

Ideally, it should import the interface from ms-rest. It's the same thing you do with the ServiceClientOptions type. Basically change the first generated line from:

````typescript
import { ServiceClientOptions, RequestOptions, ServiceCallback } from 'ms-rest';
```

to

```typescript
import { ServiceClientOptions,  ServiceClientCredentials, RequestOptions, ServiceCallback } from 'ms-rest';
```

PR #1495, opened 2016-10-06
nodejs: codegen + runtime fixes

amarzavery:
- Generator
 - Improvised url construction (no need to search for double forward slashes and remove them)
 - Achieved consistency in naming properties with dots in them
  `odata.type -> propertyName = odatatype and serializedName = "odata\\.type"`
 - added imports for type definitions exported from runtime
 - generated fragmented paging and added a test for it
- Runtime 
  - ServiceClientCredentials definition is properly defined in the d.ts file and gets imported correctly when `--AddCredentials` option is specified with Generic Node.js codegenerator.
  -  updated ts definitions and made some documentation changes in the azure runtime


amarzavery:
https://github.com/Azure/autorest/issues/1467

PR #1496, opened 2016-10-06
[Ruby - In review] Refactoring calls to create faraday connection

veronicagg:
Refactoring calls to create connections in client runtime, to centralize where we create new Faraday connections.

### DIFF
--- a/src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/polling_state.rb
+++ b/src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/polling_state.rb
@@ -99,9 +99,9 @@ module MsRestAzure
     def get_request(options = {})
       link = @azure_async_operation_header_link || @location_header_link
       faraday_options = {
-          :middlewares => request.middlewares,
-          :headers => request.headers,
-          :log => request.log
+          middlewares: request.middlewares,
+          headers: request.headers,
+          log: request.log
       }
       @connection ||= MsRest::HttpOperationRequest.create_faraday_connection(options[:base_uri], faraday_options)
       options[:connection] = @connection
--- a/src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/polling_state.rb
+++ b/src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/polling_state.rb
@@ -99,9 +99,10 @@ module MsRestAzure
     def get_request(options = {})
       link = @azure_async_operation_header_link || @location_header_link
       faraday_options = {
-          middlewares: request.middlewares,
-          headers: request.headers,
-          log: request.log
+          middlewares: @request.middlewares,
+          headers: @request.headers,
+          log: @request.log,
+          ssl: @request.ssl
       }
       @connection ||= MsRest::HttpOperationRequest.create_faraday_connection(options[:base_uri], faraday_options)
       options[:connection] = @connection
--- a/src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/polling_state.rb
+++ b/src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/polling_state.rb
@@ -98,7 +98,14 @@ module MsRestAzure
     
     def get_request(options = {})
       link = @azure_async_operation_header_link || @location_header_link
-      options[:connection] = create_connection(options[:base_uri])
+      faraday_options = {
+          :middlewares => request.middlewares,
+          :headers => request.headers,
+          :log => request.log
+      }
+      @connection ||= MsRest::HttpOperationRequest.create_faraday_connection(options[:base_uri], faraday_options)
+      options[:connection] = @connection
+
       MsRest::HttpOperationRequest.new(nil, link, :get, options)
     end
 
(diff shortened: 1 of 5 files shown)


src/client/Ruby/ms-rest/lib/ms_rest/http_operation_request.rb
```diff
@@ -121,7 +117,20 @@ def to_json(*a)
         log: log  
       }.to_json(*a)
     end
-    
+
+    def self.create_faraday_connection(base_uri, options = {})
+      Faraday.new(:url => base_uri, :ssl => MsRest.ssl_options) do |faraday|
```
vishrutshah: Should we also streamline how `ssl_options` are getting passed to the `create_faraday_connection`. Example: options should allow setting them or not?

src/client/Ruby/ms-rest/lib/ms_rest/http_operation_request.rb
```diff
@@ -121,7 +117,20 @@ def to_json(*a)
         log: log  
       }.to_json(*a)
     end
-    
+
+    def self.create_faraday_connection(base_uri, options = {})
```
vishrutshah: Any thoughts for why `create_faraday_connection` being a class method instead of instance method?

src/client/Ruby/ms-rest/lib/ms_rest/http_operation_request.rb
```diff
@@ -68,23 +68,19 @@ def initialize(base_uri, path_template, method, options = {})
...
-          middlewares.each{ |args| faraday.use(*args) } unless middlewares.nil?
-          faraday.adapter Faraday.default_adapter
-          logging = ENV['AZURE_HTTP_LOGGING'] || log
-          if logging
-            faraday.response :logger, nil, { :bodies => logging == 'full' }
-          end
-        end
+        options = {
+            :middlewares => middlewares,
+            :log => log
```
vishrutshah: This is not regarding this review but in general i am not sure how `log` property is being used in this context? Any idea?

src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/polling_state.rb
```diff
@@ -109,17 +116,5 @@ def get_request(options = {})
...
-        faraday.headers = request.headers
-        logging = ENV['AZURE_HTTP_LOGGING'] || request.log
-        if logging
-          faraday.response :logger, nil, { :bodies => logging == 'full' }
-        end
-      end
-    end
   end
-
 end
```
vishrutshah: Again this is not regarding your change but i just notices. We should add "a new line character" on line 120[last line].

src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/polling_state.rb
```diff
@@ -98,7 +98,14 @@ def get_operation_error
     
     def get_request(options = {})
       link = @azure_async_operation_header_link || @location_header_link
-      options[:connection] = create_connection(options[:base_uri])
+      faraday_options = {
+          middlewares: request.middlewares,
+          headers: request.headers,
+          log: request.log
```
vishrutshah: in general, I am curious & I don't know what are the best practices on when to use @ while accessing property and when not? 
Example: request being a property we are accessing it as `request` but in PollingState class azure_async_operation_header_link being property we are accessing it as `@azure_async_operation_header_link`

src/client/Ruby/ms-rest/lib/ms_rest/http_operation_request.rb
veronicagg: Faraday supports log options (https://github.com/lostisland/faraday/blob/master/lib/faraday/response/logger.rb), HttpOperationRequest creation allows for `:log` to be passed as part of the options. If set, then we should pass that when creating a Faraday connection.

src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/polling_state.rb
veronicagg: Here's some info:
- http://stackoverflow.com/questions/14319347/variables-in-ruby-on-rails 
- http://stackoverflow.com/questions/9809843/understanding-ruby-class-attributes-using-accessor-macros-and-self
I can update this to use @, since we're inside the class it's probably clearer to use it, though I've seen other cases already in the code that don't use it.

src/client/Ruby/ms-rest-azure/lib/ms_rest_azure/polling_state.rb
veronicagg: sure.

src/client/Ruby/ms-rest/lib/ms_rest/http_operation_request.rb
veronicagg: sure, I could include it, if it passed in options.

fearthecowboy:
Since this is for the Client Runtime, can we close this here and open it in the new home for the client runtime?

vishrutshah:
@veronicagg Closing this for @fearthecowboy :)

PR #1503, opened 2016-10-10
Add min length and max length support to schema generator

daschult:


src/generator/AutoRest.AzureResourceSchema/JsonSchema.cs
```diff
@@ -99,6 +99,16 @@ public string ResourceType
         public string Pattern { get; set; }
 
         /// <summary>
+        /// The minimum length that a string or an array matching this schema can have.
+        /// </summary>
+        public double? MinLength { get; set; }
```
daschult: I tried to keep the C# as close to the JSON as possible. Since the JSON value is a number, I thought it made sense to use C#'s equivalent data type.

Issue #1499, opened 2016-10-06
Validation rule for the produces/consumes types that AutoRest supports


tbombach:
Right now, AutoRest only supports JSON across the wire. Swagger specs can contain any MIME type, but we'd like to let the author know that AutoRest will not make requests using that MIME type.

This is indicated in Swagger files with produces and consumes sections. We should have a validation rule that each MIME type in that array is application/json. If it's not, we'll show a message that AutoRest won't use that MIME type in the generated code. I think it should be either a Warning or Info level message, since it doesn't break AutoRest to list extra MIME types.

brjohnstmsft:
What about supporting parameters on the `application/json` MIME type? For example, OData services can support varying levels of metadata. For example, `application/json;odata.metadata=none` returns less metadata than `application/json;odata.metadata=minimal`. Is there any reason AutoRest can't use MIME types like these?

tbombach:
Good point - the validation rule should make sure that it starts with `"application/json"`, not that it matches `"application/json"` exactly. This rule should use the same logic that's used during operation building: https://github.com/Azure/autorest/blob/master/src/modeler/AutoRest.Swagger/OperationBuilder.cs#L408 

Thanks @brjohnstmsft!

dsgouda:
Issue fixed here https://github.com/Azure/autorest/pull/1507

PR #1507, opened 2016-10-14
Adding validation rules for unique resource paths

dsgouda:
- Contains validation rules for Non app json media types and Non-Https schemes 
- Fixing documentation to account for changes to clean json when adding new validation rules
- Adding validation rule that ensures we reference a unique resource path within the given spec
Addressing issues https://github.com/Azure/autorest/issues/1499 , #1516 and #1517
@tbombach PTAL, also need you to verify the error message here


src/modeler/AutoRest.Swagger.Tests/Swagger/Validation/positive/clean-complex-spec.json
```diff
@@ -10,7 +10,7 @@
   "host": "petstore.swagger.io",
   "basePath": "/v1",
   "schemes": [
-    "http"
+    "Https"
```
fearthecowboy: Case? shouldn't this be `https` ?

And do we not support `http`?

src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -252,4 +252,13 @@
   <data name="InvalidConstraint" xml:space="preserve">
     <value>Constraint is not supported for this type and will be ignored.</value>
   </data>
+  <data name="NonAppJsonTypeNotSupported" xml:space="preserve">
+    <value>Non application json MIME type has limited support</value>
```
fearthecowboy: How about `Support for media types other than "application/json" has limited support` 

Points to note: 
- be explicit about what someone should use (`"application/json"` )
- `MIME type` is actually not the correct term anymore . It's `Media Type` (it not being used to describe Multipurpose Internet Mail Extensions anymore...)

src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -252,4 +252,13 @@
...
+  </data>
+  <data name="SupportedSchemesWarningMessage" xml:space="preserve">
+    <value>Non-HTTPS schemes have limited Autorest support</value>
```
fearthecowboy: o_O what are we saying about HTTP? We support HTTP, right?

Again, maybe be clearer `Support for schemes other than {"http","https" } have limited support `

src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -252,4 +252,13 @@
...
+  </data>
+  <data name="NetworkInterfacesApiWarningMessage" xml:space="preserve">
+    <value>Swagger API can reference either Microsoft.Compute or Microsoft.Network Apis and not both</value>
```
fearthecowboy: `Swagger` is deprecated. The new term is `OpenAPI` ... but maybe we should drop the reference to the format, and say something like `May not reference more than one of {"Microsoft.Compute","Microsoft.Network"}` 

- this makes it so we can add another item to the list without substantially changing the message later.

src/modeler/AutoRest.Swagger/Validation/NetworkInterfacesApiWarning.cs
```diff
@@ -0,0 +1,54 @@
...
+        private const string MicrosoftNetworkApi = "Microsoft.Network";
+        /// <summary>
+        /// This rule passes if the paths contain reference to either Microsoft.Network Apis or 
+        /// Microsoft.Compute Apis but not both
+        /// </summary>
+        /// <param name="paths"></param>
+        /// <returns></returns>
+        public override bool IsValid(Dictionary<string, Dictionary<string, Operation>> paths)
+        {
+            Regex ApiRegExp = new Regex(@"(/.+)+/((M|m)icrosoft.(C|c)ompute|(M|m)icrosoft.(N|n)etwork)/.+");
```
fearthecowboy: `MICROSOFT.COMPUTE` will pass thru that. 

Perhaps we want a case insensitive regex?

src/core/AutoRest.Core/Properties/Resources.resx
dsgouda: I believe @tbombach indicated we support https only (not http)

src/modeler/AutoRest.Swagger/Validation/NonAppJsonTypeWarning.cs
```diff
@@ -0,0 +1,35 @@
...
+    public class NonAppJsonTypeWarning : TypedRule<string>
+    {
+        private const string AppJsonType = "application/json";
+        /// <summary>
+        /// This rule passes if the entity contains application/json type 
+        /// </summary>
+        /// <param name="entity"></param>
+        /// <returns></returns>
+        public override bool IsValid(string entity)
+            => (entity.ToLowerInvariant().StartsWith(AppJsonType));
```
fearthecowboy: `StartsWith` ? 

Hmm. that's probably a good idea, good catch. I can't see a case where we shouldn't do that, but there may be one day.

fearthecowboy:
LGTM 👍

tbombach:
(As mentioned offline, Deepak is going to make this rule more generic. The rule will check that paths only contain one RP path, not for Network/Compute specifically)

tbombach:
Minor: make sure the PR title is updated, since the rule is more generic now

tbombach:
This rule should work for all Azure services, without having to add a resource type to the list to check. The general logic will be:

1. Create an empty list of resource types names.
2. Iterate through each path
3. Capture the RP resource type with a regex (i.e., given a path "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/microsoft.Foo", it will capture "Foo") and add it to the list of resource type names
  - Make sure that there's no exception thrown or problem if the path is not of the above form. That just means we should skip that path, because it won't affect whether or not this rule passes.
4. Make sure that the list of resource types is unique

You could optimize this by only storing the first resource type name that you find and failing the rule as soon as you see a different one.

fearthecowboy:
LGTM 👍

:shipit:

dsgouda:

src/modeler/AutoRest.Swagger/Validation/UniqueResourcePaths.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        /// <summary>
+        /// This rule passes if the paths contain reference to exactly one of the namespace resources
+        /// </summary>
+        /// <param name="paths"></param>
+        /// <returns></returns>
+        public override bool IsValid(Dictionary<string, Dictionary<string, Operation>> paths) 
+            => paths.Keys.Select(path => 
+                    { return apiRegExp.Match(path).Success ? apiRegExp.Match(path).Groups?[0]?.Value.ToString().ToLowerInvariant() : null; })
+                                      .Except(new List<string> { null }).Distinct().Count() <= 1;
```
tbombach: I think `.Where(p => p != null)` is preferable over `.Except` for readability, plus it doesn't instantiate a new List to filter nulls

tbombach:
Only things to change is the `.Except` to `.Where`, then it looks good to me (feel free to merge after updating the PR)

PR #1509, opened 2016-10-14
Introduced Plugin Model (refactored plugin classes)

fearthecowboy:
This branch is for the next phase of the Code Model Refactor - this updates the model to include a `Plugin` class, which is now the thing we load from a plugin assembly. 

The `Plugin` is the glue that binds all the pieces together. 
(aka, the jam in the AutoRest jelly roll)


tbombach:
LGTM - all questions/comments discussed offline

PR #1513, opened 2016-10-17
Adding validation rules for subscriptionId/api-version params

dsgouda:
Adding validation rules to enforce:
- Parameters section for operations should not define the subscriptionId and api-version parameters
- Service definition section must contain parameters api-version and subscriptionId

Modified all swagger validation test jsons to adhere to these validation rules
Fixing issue https://github.com/Azure/autorest/issues/1515

@tbombach


fearthecowboy:
If you're changing JSON files, make sure you pull all the latest code and do a `gulp regenerate:expected test` and then add in any changed files with a separate commit message 👍

src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -252,4 +252,10 @@
   <data name="InvalidConstraint" xml:space="preserve">
     <value>Constraint is not supported for this type and will be ignored.</value>
   </data>
+  <data name="OperationParametersNotAllowedMessage" xml:space="preserve">
+    <value>Parameters "subscriptionId" and "api-version" are not allowed in the operations section of azure swagger specs</value>
```
fearthecowboy: `Parameters "subscriptionId" and "api-version" are not allowed in the operations section` ?

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
```diff
@@ -0,0 +1,43 @@
...
+        private const string SubscriptionId = "subscriptionId";
+        private const string ApiVersion = "api-version";
+        /// <summary>
+        /// This rule passes if the parameters section contains both subscriptionId and api-version parameters
+        /// </summary>
+        /// <param name="paths"></param>
+        /// <returns></returns>
+        public override bool IsValid(Dictionary<string, SwaggerParameter> ParametersMap)
+        {
+            if (ParametersMap.Count==0 || ParametersMap == null)
```
fearthecowboy: Braces please. 😀 
``` csharp
if (ParametersMap.Count==0 || ParametersMap == null)
{
    return false;
}
```

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
```diff
@@ -0,0 +1,43 @@
...
+                return false;
+
+            var Names = ParametersMap.Values.Select(Param => Param.Name).ToList();
+            return Names.Contains(SubscriptionId) && Names.Contains(ApiVersion);
```
fearthecowboy: The code 
``` csharp 
var Names = ParametersMap.Values.Select(Param => Param.Name).ToList();
return Names.Contains(SubscriptionId) && Names.Contains(ApiVersion);
```
is iterating the `ParametersMap.Values` enumerable and allocating a `List` for no benefit. 
`.ToList()` should be reserved for when you actually need to manipulate the result set of an enumeration.

The following will iterate the set only once, and will return at the first `true` case. If there are a 1000 items in the list, and the first is matched, the iterating stops there. The happy path is the fast path ⏩ 
``` csharp
return ParametersMap.Values.Any( parameter => parameter.Name == SubscriptionId  || parameter.Name == ApiVersion );
```

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
fearthecowboy: Actually, you don't want to check the `count`; that presumes either that the dictionary implementation knows the actual item count (which not all do),  or you're forcing the dictionary to count the items. The iteration below will be nearly 0 time if the collection is empty anyway.

So replace this with 

``` csharp
if( ParametersMap == null ) 
{ 
    return false;
}
```

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
```
fearthecowboy: Now, I get to the bottom and see that you could simplify it a bit more, and use a single expression that leverages null-propagation operator (which turns the expression result into a nullable bool result `bool?` which can be trivially coaxed back into a bool by adding  `true == ` at the front of the expression.

``` csharp
  public override bool IsValid(Dictionary<string, SwaggerParameter> ParametersMap) => true == ParametersMap?.Values.Any( parameter => parameter.Name == SubscriptionId  || parameter.Name == ApiVersion );
```

src/modeler/AutoRest.Swagger/Validation/OperationParametersValidation.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        private const string SubscriptionId = "subscriptionId";
+        private const string ApiVersion = "api-version";
+        /// <summary>
+        /// This rule passes if the parameters are not subscriptionId or api-version
+        /// </summary>
+        /// <param name="paths"></param>
+        /// <returns></returns>
+        public override bool IsValid(SwaggerParameter Parameter)
+        {
```
fearthecowboy: rather than performing the separate checks on the `Parameter` and `Parameter.Name`, you can use the null-propagation operator to fold it into a single check and let the compiler to the heavy lifting. 

``` csharp 
public override bool IsValid(SwaggerParameter Parameter) => 
    SubscriptionId  != Parameter?.Name && 
    ApiVersion != Parameter?.Name;
```

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
dsgouda: This helps my limited knowledge of Linq 😄

src/modeler/AutoRest.Swagger/Validation/OperationParametersValidation.cs
dsgouda: New concepts, taking 🗒

fearthecowboy:
LGTM 👍

dsgouda:

PR #1521, opened 2016-10-19
[Go] Workaround golint stuttering

mcardosos:
When a model / enum / method name starts with the package name, golint will notice the stuttering. For example: `compute.ComputeVM`. Golint will suggest something like `compute.VM`. The problem is when the name of the model / enum /method is exaclty the same as the package. AutoRest currently generates `containerservice.ContainerServiceType`. With this change, it would generate `containerservice.TypeContainerService`

### DIFF
--- a/src/generator/AutoRest.Go/GoCodeNamer.cs
+++ b/src/generator/AutoRest.Go/GoCodeNamer.cs
@@ -260,21 +260,15 @@ namespace AutoRest.Go
                                             .Any(et => (et is IType && (et as IType).Name.Equals(name)) || (et is Method && (et as Method).Name.Equals(name)));
                         if (exported is EnumType)
                         {
-                            (exported as EnumType).Name = nameInUse
-                                                            ? name + "Enum"
-                                                            : name;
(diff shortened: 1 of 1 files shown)


mcardosos:
Please, take a look :3 @vishrutshah @jhendrixMSFT @garimakhulbe

jhendrixMSFT:
LGTM

ahmetalpbalkan:
You can disable stuttering warnings easily. They're not the end of the world. First of all `type containerservice.ContainerServiceType` sounds really weird. Why is there a `Type` at the end? That said, `containerservice.TypeContainerService` isn't any less weird. 

I'm okay with stuttering here.

mcardosos:
I agree it sounds super weird. For me, the ideal thing would be for it to be named just `ContainerService`. Do you know how to disable the stuttering warnings?

ahmetalpbalkan:
`go lint | grep -v stutter`

mcardosos:
Interesting thing, golint won't complain if the struct has exactly the same name as the package :D So yeah, the struct can be named `ContainerService` without problem :3

tbombach:
@fearthecowboy fyi, this is a change to the Go namer

PR #1530, opened 2016-10-24
New azure resource schema

fearthecowboy:
This adapts the `AzureResourceSchema` generator to work with the new code model.

Nothing too crazy here, just some weirdness with UTF8-BOM in some files -- I will follow up before merging.


fearthecowboy:
Branch problem... rebasing first.

PR #1531, opened 2016-10-24
Updated AzureResourceGenerator to work in new code model

fearthecowboy:
Not much changes here... mostly the required changes, fixed up the test running and regenerated the files (so that the UTF-8 BOM isn't there)

// cc @daschult  for review


daschult:
Looks good to me.

PR #1307, opened 2016-07-24
Removed most Using Statements from CSharp

John-Hart:
Removed all Using Statements except for those with extension methods and Models

PR #1538, opened 2016-10-26
Restore usings

tbombach:
Undoing change https://github.com/Azure/autorest/pull/1307 that removed usings.

The usings must be present alongside the fully qualified names so that Roslyn knows what can be reduced to shorter names (it doesn't move the fully qualified names to usings section). There will be a related .NET SDK PR that adds the reduction step to the generation cmd


tbombach:
Will reopen against refactor branch

PR #1539, opened 2016-10-26
Changes to the Python Code Generator for the Refactored Code Model 

fearthecowboy:
a few issues to resolve, should be in today.


fearthecowboy:
Having been reviewed with the relavent teams, I'm merging this with `RefactorCodeModel`

PR #1543, opened 2016-10-27
Fix for Java and Java.Fluent codegens

hovsepm:
Without this fix CDN Java code is generated incorrectly - ResourceGroupName global parameter is generated as client level property, but it should be method input parameter for Get/Put/Patch/Delete operations.


fearthecowboy:
LGTM 👍

PR #1551, opened 2016-10-28
Generate markdown help docs when generating template schemas.

jhendrixMSFT:
The AzureResourceSchema generator will now generate a help document
alongside the generated resource schema.  The doc contains an example
schema and tables describing the various properties; this is modeled from
the existing help docs that are hand-written.

### DIFF
--- a/src/core/AutoRest.Core/CodeGenerator.cs
+++ b/src/core/AutoRest.Core/CodeGenerator.cs
@@ -76,7 +76,7 @@ namespace AutoRest.Core
             {
                 await template.ExecuteAsync().ConfigureAwait(false);
             }
-            await Write(stringBuilder.ToString(), fileName);
+            await Write(stringBuilder.ToString(), fileName, true);
         }
 
         /// <summary>
@@ -84,8 +84,9 @@ namespace AutoRest.Core
         /// </summary>
         /// <param name="template"></param>
         /// <param name="fileName"></param>
+        /// <param name="skipEmptyLines"></param>
         /// <returns></returns>
-        public async Task Write(string template, string fileName)
+        public async Task Write(string template, string fileName, bool skipEmptyLines)
         {
             string filePath = null;
 
@@ -131,7 +132,7 @@ namespace AutoRest.Core
                     {
                         await textWriter.WriteLineAsync();
                     }
-                    else if (!string.IsNullOrWhiteSpace(line))
+                    else if (!skipEmptyLines || !string.IsNullOrWhiteSpace(line))
                     {
                         await textWriter.WriteLineAsync(line);
                     }
--- a/src/generator/AutoRest.AzureResourceSchema/JsonSchema.cs
+++ b/src/generator/AutoRest.AzureResourceSchema/JsonSchema.cs
@@ -13,6 +13,7 @@ namespace AutoRest.AzureResourceSchema
     /// </summary>
     public class JsonSchema
     {
+        private string resourceType;
         private IList<string> enumList;
         private IDictionary<string, JsonSchema> properties;
         private IList<string> requiredList;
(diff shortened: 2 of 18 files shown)


jhendrixMSFT:
@tbombach Here's the PR for the markdown generator.

tbombach:
Awesome, look good to me! @daschult any concerns before I merge?

src/core/AutoRest.Core/CodeGenerator.cs
```diff
@@ -139,7 +140,7 @@ public virtual void PopulateSettings(IDictionary<string, object> settings)
                     {
                         await textWriter.WriteLineAsync();
                     }
-                    else if (!string.IsNullOrWhiteSpace(line))
+                    else if (!skipEmptyLines || (skipEmptyLines && !string.IsNullOrWhiteSpace(line)))
```
daschult: The second check for skipEmptyLines is unnecessary since it can only be true if the second expression is evaluated.

src/generator/AutoRest.AzureResourceSchema/AzureResourceSchemaCodeGenerator.cs
```diff
@@ -50,7 +50,13 @@ public override void NormalizeClientModel(ServiceClient serviceClient)
                 ResourceSchemaWriter.Write(stringWriter, resourceSchemas[resourceProvider]);
 
                 string schemaPath = Path.Combine(Settings.OutputDirectory, resourceProvider + ".json");
-                await Write(stringWriter.ToString(), schemaPath);
+                await Write(stringWriter.ToString(), schemaPath, true);
+
+                stringWriter = new StringWriter();
+                ResourceMarkdownWriter.Write(stringWriter, resourceSchemas[resourceProvider]);
```
daschult: Why did you choose to make this part of the AzureResourceSchema generator, and not its own generator?

src/generator/AutoRest.AzureResourceSchema/JsonSchema.cs
```diff
@@ -13,6 +13,7 @@ namespace AutoRest.AzureResourceSchema
     /// </summary>
     public class JsonSchema
     {
+        private string resourceType;
```
daschult: Did you pull this out of the properties dictionary for performance reasons?

src/generator/AutoRest.AzureResourceSchema/JsonSchema.cs
```diff
@@ -48,38 +49,29 @@ public class JsonSchema
...
-                }
-
-                return result;
+                return resourceType;
             }
             set
             {
-                if (properties == null)
+                resourceType = value;
+                if (Properties != null && Properties.ContainsKey("type"))
```
daschult: Previously properties could be initialized in this (ResourceType) property's setter. Now it can't. Is that a regression?

src/generator/AutoRest.AzureResourceSchema/ResourceSchemaParser.cs
```diff
@@ -82,6 +82,7 @@ public static class ResourceSchemaParser
                     JsonSchema resourceDefinition = new JsonSchema();
                     resourceDefinition.JsonType = "object";
 
+                    resourceDefinition.ResourceType = resourceType;
```
daschult: Why not just look for the "type" property instead of creating a ResourceType property?

fearthecowboy:
The `RefactorCodeModel ` branch is about to be merged; this will have to wait until that branch is merged to `master`.

src/generator/AutoRest.AzureResourceSchema/ResourceSchemaParser.cs
jhendrixMSFT: This heuristic isn't reliable.  For example there are some swaggers that contain a property named "type" that is unrelated to resource type (I assume this is allowed).

src/generator/AutoRest.AzureResourceSchema/JsonSchema.cs
jhendrixMSFT: No, we still initialize it as required in the AddProperty() method.

src/generator/AutoRest.AzureResourceSchema/JsonSchema.cs
jhendrixMSFT: No, I removed it because relying on an entry with the name "type" is ambiguous (see my other comment regarding setting this value).

jhendrixMSFT:
@fearthecowboy Is the merge complete?

fearthecowboy:
If this passes tests and merges cleanly against master, I won't block it now :D

jhendrixMSFT:
@daschult Do you have any more comments?

daschult:
No, go ahead and merge it.

jhendrixMSFT:
@fearthecowboy If you're good with the changes could you please merge?

PR #1571, opened 2016-11-03
Extending test case for enum type arrays in CSharp

dsgouda:
@fearthecowboy PTAL

### DIFF
--- a/src/generator/AutoRest.CSharp.Azure/Model/MethodCsa.cs
+++ b/src/generator/AutoRest.CSharp.Azure/Model/MethodCsa.cs
@@ -88,7 +88,7 @@ namespace AutoRest.CSharp.Azure.Model
         /// <summary>
         /// Returns true if method has x-ms-long-running-operation extension.
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/generator/AutoRest.CSharp.Unit.Tests/Bug1161.cs
```diff
@@ -53,10 +54,16 @@ public Bug1161(ITestOutputHelper output) : base(output)
...
+                // try to load the assembly
+                var asm = Assembly.Load(result.Output.GetBuffer());
+                Assert.NotNull(asm);
                 
+                // verify that we have the class we expected
+                var recurrenceScheduleObject = asm.ExportedTypes.FirstOrDefault(each => each.FullName == "Test.Models.RecurrenceSchedule");
+                Assert.NotNull(recurrenceScheduleObject);
+                // verify the property is generated 
+                var prop = recurrenceScheduleObject.GetProperty("WeekDays");
+                Assert.NotNull(prop);
```
fearthecowboy: check the type (ie, prop.GetType() and make sure it is the expected type, and that it is nullable)

Issue #1534, opened 2016-10-24
ReadOnly Attributes should not be included in parameter flattening


annatisch:
Hello,

I'm not sure if other generated SDK's use parameter flattening, but in Python we use it to a threshold of 2, and it's caused some difficulties in the KeyVault client, where one of the Body Parameters includes a read-only attribute.

Read Only attributes should not be included in the count towards flattening threshold, nor should it be exposed as a parameter on the method which currently seems to be the case.

The issue seems to be here:
https://github.com/Azure/autorest/blob/master/src/core/AutoRest.Extensions/SwaggerExtensions.cs#L403

My fix would be:

```
if (bodyParameterType != null && (bodyParameterType.ComposedProperties.Count(p => !p.IsConstant  && !p.IsReadOnly) <= settings.PayloadFlatteningThreshold || bodyParameter.ShouldBeFlattened()))
{
    var parameterTransformation = new ParameterTransformation
    {
        OutputParameter = bodyParameter
    };
    method.InputParameterTransformation.Add(parameterTransformation);

    foreach (var property in bodyParameterType.ComposedProperties.Where(p => !p.IsConstant && p.Name != null  && !p.IsReadOnly))
    {
        var newMethodParameter = new Parameter();
        newMethodParameter.LoadFrom(property);
        ...
```

Let me know if you think this looks reasonable and I'll submit a PR.
Cheers,
Anna

FYI @lmazuel, @tbombach

annatisch:
@fearthecowboy - would this patch have any conflicts with your refactor?

fearthecowboy:
A smidgen... I can make the change in here too. It's just changing the `!p.IsConstant `  to `!p.IsConstant  && !p.IsReadOnly` on those two lines?

annatisch:
Yup, that's all it is - I think ReadOnly and Constant should be treated the same in this scenario - and no one has corrected me yet ;)

fearthecowboy:
I've put this change into the `RefactorCodeModel` branch already, and since it'll be `master` pretty quick, don't worry about doing anything with the current master.

annatisch:
Awesome - thank you!

annatisch:
@fearthecowboy - it looks like this patch somehow got lost in the merge to master.
So I've included it in a PR: #1572

PR #1572, opened 2016-11-03
Fix for issue #1534

annatisch:
Also made Python namespaces lowercase.


fearthecowboy:
Looks good to me. :+1:

fearthecowboy:
Ouch, looks like a test needs to be fixed.

fearthecowboy:
I'm ok with this. 

Ideally, we didn't have so many damn tests that have one parameter and then just constants, but fixing this is what you'd really want, right?

Merging it.

PR #1574, opened 2016-11-04
Recognize mac os darwin version 10.12

vishrutshah:
Me (the early adopter) using Mac OS version 10.12 ran into the problem. With this PR now AutoRest will recognize both version of Darwin OS 10.11 & 10.12

@fearthecowboy Please review as you get a chance. Thanks!


fearthecowboy:
I'm ok with this; we really need to make this better, I'm running into a problem on ubuntu too.

Still, we'll take this now, and fix it when we jump tools version next time.

LGTM 👍

Issue #1080, opened 2016-05-24
[WIP] Recognize vendor extensions for primitives


gitfool:
I recently submitted [Swashbuckle PR #776](https://github.com/domaindrivendev/Swashbuckle/pull/776). I'd appreciate any feedback.

Basically, if AutoRest recognized the `x-nullable` and `x-type` (or `x-type-dotnet`) vendor extensions I'm proposing, should they (or something similar) be enabled, then the AutoRest generated C# proxy could remove the type impedance mismatch we have today - whether or not a type is _nullable_ would not have to be conflated with whether or not a type is _required_, and secondly the original C# type could be preserved for C# clients. This should cover use in parameters, request / response types, composite types, arrays etc.

brjohnstmsft:
@devigned @Azure/adx-autorest-contributors Thoughts about having AutoRest support `x-nullable`? From the discussion [here](https://github.com/OAI/OpenAPI-Specification/issues/229) it seems there is consensus that it will be standardized in OpenAPI at some point.

`x-nullable` would solve a bunch of issues:

- https://github.com/Azure/autorest/issues/513
- https://github.com/Azure/autorest/issues/885
- https://github.com/Azure/autorest/issues/596

These are causing our team and others a bunch of pain. Solving it would eliminate nearly every case where we have to use `x-ms-external` in our specs today.

gitfool:
@brjohnstmsft Please also consider `x-type-dotnet` as this differentiates _all_ primitive .NET types including the commonly used `Decimal`, `DateTimeOffset` (with more fine-grained control than the `UseDateTimeOffset` command-line option), and `TimeSpan` system types.

The `x-type-dotnet` extension complements the `x-nullable` extension to provide full type fidelity for C# code generators. Generators for other languages would ignore this extension and continue to provide the same "lowest common denominator" types. :stuck_out_tongue_winking_eye:

fearthecowboy:
@gitfool I'm looking at this and related issues regarding nullability right now.

I'm going to review some of my notes with @devigned if he'd ever get back to his desk...

I think I am going to implement `x-nullable` in AutoRest, but in a discussion we had here a few days back it was suggested that `format` would be a better means to express additional type information rather than introducing an additional extension (ie, `x-type-dotnet`).

brjohnstmsft:
👍 to `x-nullable`. Glad to hear you're working on it!

gitfool:
Support for `x-nullable` would be much appreciated!

[...]

Issue #816, opened 2016-03-07
Nullable int property on response model for element that will always be present.


NiklasGustafsson:
I have a response-payload model with this property:

``` json
"_count": {
    "type": "integer",
    "readOnly": true,
    "default": 0,
    "description": "This property shows the number of databases returned as part of list operation."
},
```

Since it's a response model, whether it's required or not is not a big client-side concern, but getting good code generated is. The element will always be there, and it doesn't make any sense for it to be writable.

However, Swagger won't allow me to make it required **and** read-only, so I thought adding a default value might force it to use a non-nullable value type, but that's not the case:

``` csharp
        /// <summary>
        /// This property shows the number of databases returned as part of
        /// list operation.
        /// </summary>
        [JsonProperty(PropertyName = "_count")]
        public int? _count { get; private set; }
```

As far as I can tell, there is no means to force use of a regular value type here. I suggest that a readOnly element with an explicit default value be used to indicate that 'null' will never be an issue.

brjohnstmsft:
Duplicate https://github.com/Azure/autorest/issues/513

These are great suggestions! Please look for issues that have already been opened and +1 them; It may save both of us a bunch of time.

fearthecowboy:
@olydis -- can you write a unit test to reproduce this (if it still happens)?

Generate a few variations in a response type (read-only/not-read-only, with/without default value, required/not required) so we can see what variations are getting generated as what types, and then we can see *how* we want it fixed.

fearthecowboy:
I agree that the presence of a default value on a value type would *seem* to mean that it shouldn't be nullable, and the default value is the acceptable value if the service didn't return the value.

The other option, would be to allow `"x-nullable":false`  to forcibly say "no damn nullable, thank you very much" ...

brjohnstmsft:
@fearthecowboy Given that `x-nullable` is (or will soon be?) a thing, IMO it would be better to use it to control these client language-binding decisions, rather than overloading the semantics of standard Swagger elements too much.

olydis:
Quickly reproduced the behavior, here is a full list of what gets generated when:

| `required` | `default` | `readOnly` | `Nullable` |
|----------|---------|----------|----------|
| no       | no      | no       | yes      |
| no       | no      | yes      | yes      |
| no       | yes     | no       | yes      |
| no       | yes     | yes      | yes      |
| yes      | no      | no       | no       |
| yes      | yes     | no       | no       |

[...]

PR #1575, opened 2016-11-07
C# generator enforces x-nullable (also on lists and dictionaries)

### LINKED ISSUE #513: (C#, Java) We need a way to model non-nullable readOnly properties

### LINKED ISSUE #596: (C#) Dictionary value types are always nullable
This issue probably affects generated Java code as well; I haven't checked.

If you have Swagger like this:

```
"TextWeights": { 
  "properties": { 
    "weights": { 
      "type": "object", 
      "additionalProperties": { 
        "type": "number", 
        "format": "double" 
      }, 
    } 
  }, 
  "required": [ 
    "weights" 
  ], 
}, 
```

AutoRest generates a backing type in C# of `IDictionary<string, double?>` instead of `IDictionary<string, double>`. This is reasonable in cases where the REST API might not provide a value for every key, but this is not always the case. Sometimes the REST API guarantees that no values will be null, in which case it would be better for the client programming model to use a non-nullable primitive type.

Unfortunately I don't see an extensibility point in `additionalProperties` that would allow us to add an extension to solve this.

@devigned @Azure/adx-sdk-team Any thoughts on this? As a workaround, I'm going to use `x-ms-external` to suppress generation of any types affected by this issue, but it's a long-term maintenance problem whenever I do this.

### LINKED ISSUE #1549: Error generating C# code for string lists with formatted types
Array parameters with formatted string types result in generator error.

Steps:

```
  type: array
  items:
    type: string
    format: uuid
```

Expected:
Generates code to serialize `IList<Guid>`

Actual:
`Cannot generate a formatted sequence from a non-string List parameter`

olydis:
- always enforce `x-nullable`, not only if it is `true` (old behavior) - fixes #513, #816, #1080
- support `x-nullable` for `object/additionalProperties` - fixes #596
- support `x-nullable` for `array/items` - fixes #1549


fearthecowboy:
This PR's content included in other PR merged (#1578)

PR #1578, opened 2016-11-07
Fix and test nullable parameter behavior

### LINKED ISSUE #1088: (C#) Revise of parameter pre checking necessary?
Hello,

when I inspect Service Client method implementation, I find checking of null values of parameters even when the parameter is a struct based type e.g. a Guid.

Would it not better to check for default value here?

Regards
Rolf

### LINKED ISSUE #1554: C# codegen generates invalid code for boolean parameters
Given this parameter definition:
```json
"parameters": [
{
    "name": "myBool",
    "in": "query",
    "required": true,
    "type": "boolean"
}]
```
the following code is generated:
```csharp
if (myBool != null)
{
    _queryParameters.Add(string.Format("myBool={0}", Uri.EscapeDataString(SafeJsonConvert.SerializeObject(myBool, this.Client.SerializationSettings).Trim('"'))));
}
```
resulting in CS0472 error 

> C# The result of the expression is always 'true' since a value of type 'bool' is never equal to 'null' of type 'bool?'

Last tested with AutoRest 0.17.3

olydis:
There were some inconsistencies between parameters being nullable and null-checks in the generated code.

fixes #1088
fixes #1554


fearthecowboy:
always enforce x-nullable, not only if it is true (old behavior) - fixes #513, #816, #1080
support x-nullable for object/additionalProperties - fixes #596
support x-nullable for array/items - fixes #1549

Issue #1563, opened 2016-11-02
[Ruby - New Generator] Fix client runtime to not consider // in odata property


vishrutshah:
See the hack @ https://github.com/Azure/autorest/blob/master/src/generator/AutoRest.Ruby/ClientModelExtensions.cs#L731

PR #1586, opened 2016-11-09
[Ruby] Override Property-Rb to handle properties with .

vishrutshah:
Fix for https://github.com/Azure/autorest/issues/1563

### DIFF
--- a/src/generator/AutoRest.Ruby/Model/PropertyRb.cs
+++ b/src/generator/AutoRest.Ruby/Model/PropertyRb.cs
@@ -1,15 +1,33 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-using AutoRest.Core.Model;
+using AutoRest.Core;
+using AutoRest.Core.Utilities;
+using AutoRest.Extensions;
 
 namespace AutoRest.Ruby.Model
 {
     /// <summary>
     /// The model for property template.
     /// </summary>
-    public class PropertyRb : Property
+    public class PropertyRb : Core.Model.Property
     {
-        
+        protected PropertyRb()
+        {
+            // properties with dots by surrounding them with single quotes to represent 
+            // them as a single property and not as one being part of another. For example: 'odata.nextLink' 
+            Name.OnGet += name =>
+            {
+                // completely replacing base behavior.
+
+                // use either the overriden client name or the raw value
+                name = Extensions.GetValue<string>("x-ms-client-name").Else(Name.RawValue);
+
+                return CodeNamer.Instance.GetPropertyName(name);
+
+            };
+
+            SerializedName.OnGet += serializedName => (this.WasFlattened() ? serializedName : serializedName?.Replace(".", "\\\\."))?.Replace("\\\\\\\\","\\\\");
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


vishrutshah:
@veronicagg & @sarangan12 Feel free to take a look when you get a chance. Thanks!

fearthecowboy:
I'm ok with this if it's what you need :D

vishrutshah:
@fearthecowboy Yup that's what i think i need to keep as there is no other way i can think of if odata.nextlink is flattened too much to have more `.`s. Having `\\` would help us to find out that we should take `odata.nextLink` as one level in json response parsing instead of two.

src/generator/AutoRest.Ruby/Model/PropertyRb.cs
```diff
@@ -1,15 +1,33 @@
...
     /// </summary>
-    public class PropertyRb : Property
+    public class PropertyRb : Core.Model.Property
     {
-        
+        protected PropertyRb()
+        {
+            // properties with dots by surrounding them with single quotes to represent 
+            // them as a single property and not as one being part of another. For example: 'odata.nextLink' 
+            Name.OnGet += name =>
```
veronicagg: do we need to overwrite Name too?

src/generator/AutoRest.Ruby/Model/PropertyRb.cs
```diff
@@ -1,15 +1,33 @@
...
+                // completely replacing base behavior.
+
+                // use either the overriden client name or the raw value
+                name = Extensions.GetValue<string>("x-ms-client-name").Else(Name.RawValue);
+
+                return CodeNamer.Instance.GetPropertyName(name);
+
+            };
+
+            SerializedName.OnGet += serializedName => (this.WasFlattened() ? serializedName : serializedName?.Replace(".", "\\\\."))?.Replace("\\\\\\\\","\\\\");
```
veronicagg: it may be good to add a quick comment on the intention of this line, since there are 2 types of replacements

src/generator/AutoRest.Ruby/Model/PropertyRb.cs
```diff
@@ -1,15 +1,33 @@
...
     /// <summary>
     /// The model for property template.
     /// </summary>
-    public class PropertyRb : Property
+    public class PropertyRb : Core.Model.Property
     {
-        
+        protected PropertyRb()
+        {
+            // properties with dots by surrounding them with single quotes to represent
```
veronicagg: this comment seems to be missing something in the way it's written, let's update it

src/generator/AutoRest.Ruby/Model/PropertyRb.cs
vishrutshah: Actually i was trying to see where the changes of converting `\\\\\\\\` to `\\\\` affects, I did find any case as of today. For the case of keeping it simple i'll remove that case as categorizing it as dead code today. :) Let me know if you think otherwise as this wasn't the scenario i explained offline.

src/generator/AutoRest.Ruby/Model/PropertyRb.cs
vishrutshah: Nope. Removing it.

vishrutshah:
@veronicagg I've made changes as per the feedback. Feel free to review it again when you get a chance. Thanks!

src/generator/AutoRest.Ruby/Model/PropertyRb.cs
vishrutshah: @veronicagg As we discussed offline, I might have invested the hypothetical case for now but I am not certain. So let's remove it to be fail-fast in case that case appears. Thanks!

vishrutshah:
Thanks for the review!!

PR #1596, opened 2016-11-10
Extended tests in Bug1161.cs

dsgouda:
@fearthecowboy PTAL


fearthecowboy:
Something's still not right here. 

It shouldn't be deleting all those java files. When I edited just the one `.cs` file and the `gulpfile.js` and ran it, I got a lot less change.

can you make sure the `gulpfile.js` has this as the `regenerate:delete` task:

``` js
gulp.task('regenerate:delete', function(cb){
  del([
    'src/generator/AutoRest.CSharp.Azure.Tests/Expected',
    'src/generator/AutoRest.CSharp.Tests/Expected',
    'src/generator/AutoRest.NodeJS.Tests/Expected',
    'src/generator/AutoRest.NodeJS.Azure.Tests/Expected',
//    'src/generator/AutoRest.Java.Tests/src/main/java',
//    'src/generator/AutoRest.Java.Azure.Tests/src/main/java',
//    'src/generator/AutoRest.Java.Azure.Fluent.Tests/src/main/java',
    'src/generator/AutoRest.Python.Tests/Expected',
    'src/generator/AutoRest.Python.Azure.Tests/Expected',
    'src/generator/AutoRest.Go.Tests/src/tests/generated'
  ],
  cb);
});
```

PR #1600, opened 2016-11-11
Updated the Fluent C# generator to the new Code model

fearthecowboy:


src/generator/AutoRest.CSharp.Azure.Fluent/TransformerCsaf.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        public void NormalizeResourceTypes(CodeModelCsaf codeModel)
+        {
+            if (codeModel != null)
+            {
+                foreach (var model in codeModel.ModelTypes)
+                {
+                    if ((model.BaseModelType != null) && (model.BaseModelType.Extensions != null) &&
+                        model.BaseModelType.Extensions.ContainsKey(AzureExtensions.AzureResourceExtension) &&
+                        (bool) model.BaseModelType.Extensions[AzureExtensions.AzureResourceExtension])
```
olydis: `model.BaseModelType.Extensions[AzureExtensions.AzureResourceExtension]` guaranteed to be castable to `bool`?

src/generator/AutoRest.CSharp.Azure.Fluent/CodeGeneratorCsaf.cs
```diff
@@ -0,0 +1,113 @@
...
+            // Service client
+            var serviceClientTemplate = new AzureServiceClientTemplate { Model = codeModel };
+            await Write(serviceClientTemplate, $"{codeModel.Name}{ImplementationFileExtension}");
+
+            // Service client interface
+            var serviceClientInterfaceTemplate = new ServiceClientInterfaceTemplate { Model = codeModel };
+            await Write(serviceClientInterfaceTemplate, "I" + codeModel.Name + ImplementationFileExtension);
+
+            // Operations
+            foreach (MethodGroupCs group in codeModel.Operations)
```
olydis: 90% identical to `CodeGeneratorCsa.Generate` (and probably even `CodeGeneratorCs.Generate`). Would probably make sense to go `virtual GenerateOperations/GenerateModels/...`, call those here and only adjust them it if necessary.
Not only would it be less code, it would also be super clear what changes with each specialization (I diffed with `CodeGeneratorCsa` to figure that out)

src/generator/AutoRest.CSharp.Azure.Fluent/ClientModelExtensions.cs
```diff
@@ -1,29 +1,32 @@
...
+        public static bool IsResource(this IModelType type)
         {
-            CompositeType compositeType = type as CompositeType;
-            return compositeType != null && 
-                (type.Name == "Resource" || type.Name == "SubResource" ||
-                 type.Name == "Microsoft.Rest.Azure.Resource" || type.Name == "Microsoft.Rest.Azure.SubResource");
+            var compositeType = type as CompositeType;
+            return (compositeType != null) &&
+                   ((type.Name == "Resource") || (type.Name == "SubResource") ||
+                    (type.Name == "Microsoft.Rest.Azure.Resource") || (type.Name == "Microsoft.Rest.Azure.SubResource"));
```
olydis: These 4 strings appear throughout (2 of them aren't even short). Maybe centralize them somewhere? That would mean less refactoring work for future Garrett :D

src/generator/AutoRest.CSharp.Azure.Fluent/CodeGeneratorCsaf.cs
fearthecowboy: It's actually not my code... I believe that it's still 'in development' .. and generally I didn't want to tweak too much. 

I think I'm going to have to wait and see if this doesn't change much before they are done with it. I get the feeling it's still a looooong way to go.

src/generator/AutoRest.CSharp.Azure.Fluent/TransformerCsaf.cs
fearthecowboy: I can't see that necessarily being true (again, not my code) ... but I think we can wait until we scrub the way extensions are done.

src/generator/AutoRest.CSharp.Azure.Fluent/ClientModelExtensions.cs
fearthecowboy: If this was a mainstream generator,  I'd be right there with you. At this point it's a big prototype, and I'm not sure that it's ready to even bother yet. But good catch.

src/generator/AutoRest.CSharp.Azure.Fluent/TransformerCsaf.cs
olydis: scrubbing sounds good :D

olydis:
ah alright wasn't aware of the prototype status of this generator... of course, then we shouldn't spend more time on perfectionizing this while there is so much evil in more important components.

PR #1602, opened 2016-11-14
Multiple fixes in the code generator

amarzavery:
- Improved error handling in node.js
- Fixed a bug in the implementation of HEAD operations in C#, node.js and Ruby. (The generator was incorrectly checking for responsebody in a HEAD operation. This resulted in the the method returning void instead of a boolean value (true if positive status code, false if negative status code)


fearthecowboy:
Why did all those python files get regenerated?

fearthecowboy:
And what's with the .sln file changes. Did you pull from master beforehand? o_O

PR #1605, opened 2016-11-14
Fixing bug 1264

dsgouda:
Fixing paths to have collectionFormat only instead of parameterName:collectionFormat

@fearthecowboy PTAL


Samples/azure-storage/Azure.CSharp.Fluent/StorageAccountsOperations.cs
```diff
@@ -195,7 +189,7 @@ internal StorageAccountsOperations(StorageManagementClient client)
                 throw ex;
             }
             // Create Result
-            var _result = new Microsoft.Rest.Azure.AzureOperationResponse<CheckNameAvailabilityResultInner>();
+            var _result = new Microsoft.Rest.Azure.AzureOperationResponse<CheckNameAvailabilityResultCheckNameAvailabilityResultInner>();
```
fearthecowboy: Oh, that doesn't look quite right

PR #1608, opened 2016-11-14
Simplified C# async codegen

olydis:
For mysterious reasons the C# generator was launching a *fresh* task to start async operations, only to synchronously wait for it right away. 
It's like starting a program using `cmd /c` instead of just starting it.
Wat.


olydis:
@hovsepm

hovsepm:
@olydis as pointed by @stankovski [this](http://stackoverflow.com/questions/28305968/use-task-run-in-synchronous-method-to-avoid-deadlock-waiting-on-async-method) and [this ](https://blogs.msdn.microsoft.com/pfxteam/2012/04/13/should-i-expose-synchronous-wrappers-for-asynchronous-methods/) articles there could be deadlock issues with the change that your propose in some scenarios.

olydis:
I don't see how, but maybe I get TPL details wrong:
the deadlock occurs if context `A` tries to acquire the synchronization context `B`, which is currently waiting for `A` to finish. The well known examples you link to have `B` as the UI thread and `A` as some async method that (ooops) needs the UI to complete.

Do we have that scenario? *We are the library writer* in Stephen Toub's article, and we already follow all his advice regarding (as far as I can tell that's the `ConfigureAwait(false)` part), so I can think of no scenario in which our method deadlocks.
What we are currently doing is applying both the "we are the responsible library writer" and the "we are consumers of a poorly written library" approaches.

I'll write a mean test case with a UI thread to confirm, but it looks to me like we as the library writer have already done our job (if not, note that people who use our Async method in a bigger task and want to wait on that are still screwed)

fearthecowboy:
Having reviewed this a couple times, I'm agreeing with @olydis  -- we do indeed do the `.ConfigureAwait(false)` which is the prescribed guidance.  

Having the extra `Task.Run` in the generated code is sloppy and unnecessary. 

@olydis  - pull from master and regenerate again, and update the PR, and we'll run the full test suite. Unless that breaks, I think I'll merge this.

olydis:
@fearthecowboy done

Issue #1417, opened 2016-09-11
If an operation parameter name is the same as a definition name, the definition type is used


Osipion:
If an operation parameter name is the same as the name of a definition, AutoRest types the API method parameter as type described by the schema definition, rather than the type specified in the parameter declaration.

Example:

``` json
{
    "swagger" : "2.0",
    "info": {
      "title": "example",
      "description": "",
      "version": ""
    },
    "basePath" : "/a",
    "schemes" : [
        "http"
    ],
    "consumes" : [
        "application/json"
    ],
    "produces" : [
        "application/json"
    ],
    "paths" : {
        "/a/path" : {
            "get" : {
                "description" : "Returns a query result",
                "parameters" : [{
                        "name" : "query",
                        "in" : "query",
                        "type" : "string",
                        "required" : false
                    }
                ],
                "responses" : {
                    "200" : {
                        "description" : "queries",
                        "schema" : {
                            "$ref" : "#/definitions/query"
                        }
                    }
                },
                "tags" : [
                    "Skill Tag"
                ],
                "operationId" : "getQuery"
            }
        }
    },
    "definitions" : {
        "query": {
            "type" : "object",
            "properties": {
                "prop1": {
                    "type": "string"
                }
            }
        }
    }
}
```

The GET operation on /a/path takes a query parameter of type string, called query. However, AutoRest produces the following API method definition with a query parameter of type Query:

``` csharp
Task<HttpOperationResponse<Query>> GetQueryWithHttpMessagesAsync(Query query = default(Query), Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken));
```

If the spec is changed so that the definition name is different:

``` json
{
    "swagger" : "2.0",
    "info": {
      "title": "example",
      "description": "",
      "version": ""
    },
    "basePath" : "/a",
    "schemes" : [
        "http"
    ],
    "consumes" : [
        "application/json"
    ],
    "produces" : [
        "application/json"
    ],
    "paths" : {
        "/a/path" : {
            "get" : {
                "description" : "Returns a query result",
                "parameters" : [{
                        "name" : "query",
                        "in" : "query",
                        "type" : "string",
                        "required" : false
                    }
                ],
                "responses" : {
                    "200" : {
                        "description" : "queries",
                        "schema" : {
                            "$ref" : "#/definitions/query1"
                        }
                    }
                },
                "tags" : [
                    "Skill Tag"
                ],
                "operationId" : "getQuery"
[...]

PR #1614, opened 2016-11-16
Correctly disambiguate model types.

olydis:
Model types were compared exclusively by their name to decide equality, which is not the correct behavior.
A manifestation of this bug is issue #1417 which is now fixed.

Fixing this bug exposed two others, which are now fixed:
- property flattening depended on the order in which composite types are processed
- the composite swagger modeler did not merge method response types correctly


src/core/AutoRest.Core/Model/SequenceType.cs
```diff
@@ -42,32 +42,48 @@ public override string ToString()
             return Name;
         }
 
+        ///// <summary>
```
fearthecowboy: remove the commented out unused code?

src/core/AutoRest.Core/Model/SequenceType.cs
```diff
@@ -42,32 +42,48 @@ public override string ToString()
...
-        /// Serves as a hash function based on ElementType.
-        /// </summary>
-        /// <returns>
-        /// A hash code for the current object.
-        /// </returns>
-        public override int GetHashCode()
-        {
-            return ElementType.GetHashCode();
-        }
+        ///// <summary>
```
fearthecowboy: this too :D

PR #1617, opened 2016-11-16
Adds c# simplifier to AutoRest.exe 

fearthecowboy:
AutoRest will now shorten references and use `using` statements in generated c# code by default.

Can be disabled with `-DontSimplify`


src/core/AutoRest.Core/Utilities/FileSystem.cs
```diff
@@ -12,7 +12,7 @@ public class FileSystem : IFileSystem
     {
         public void WriteFile(string path, string contents)
         {
-            File.WriteAllText(path, contents, Encoding.UTF8);
+            File.WriteAllText(path, contents, new UTF8Encoding(false, true));
```
fearthecowboy: Suppresses UTF-8 BOM in outputs

src/core/AutoRest/Simplify/CSharpSimplifier.cs
```diff
@@ -0,0 +1,102 @@
...
+
+                return mscorlib;
+            }
+        }
+
+        public async Task Run()
+        {
+            var op = new AzureAsyncOperation();
+            var restOp = new RestException();
+
```
olydis: I assume those two guys make sure that `AppDomain.CurrentDomain.GetAssemblies()` contains everything we need?

src/core/AutoRest/Simplify/CSharpSimplifier.cs
```diff
@@ -0,0 +1,102 @@
...
+            var files = Settings.Instance.FileSystem.GetFiles(Settings.Instance.OutputDirectory, "*.cs",
+                    SearchOption.AllDirectories).
+                ToDictionary(each => each, each => Settings.Instance.FileSystem.ReadFileAsText(each));
+
+            var projectId = ProjectId.CreateNewId();
+            var solution = new AdhocWorkspace().CurrentSolution
+                .AddProject(projectId, "MyProject", "MyProject", LanguageNames.CSharp)
+                .AddMetadataReference(projectId, Mscorlib)
+                .AddMetadataReference(projectId, AppDomain.CurrentDomain.GetAssemblies()
```
olydis: would `typeof(AzureAsyncOperation).Assembly` do the same? Same with `RestException` below

src/core/AutoRest/Simplify/AdHocEqualityComparer.cs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// 
+
+using System;
+using System.Collections.Generic;
+
+namespace AutoRest.Simplify
+{
+    public class AdHocEqualityComparer<T> : IEqualityComparer<T>
```
olydis: Think we added one of those today :D 
see `AutoRest.Core.Utilities.EqualityComparer`

src/core/AutoRest/Simplify/CSharpSimplifier.cs
fearthecowboy: Will fix when moving to dotnet core for autorest.

Issue #1374, opened 2016-08-24
New rule: Check for properties marked `readOnly` & `required`


tbombach:
### Rule logic

Show an error/warning if a definition property is marked `readOnly: true` plus it appears under the `required` array of properties.
### Rule considerations

Since it depends on multiple values, you'll have to use **context** to traverse from the property up to the definition that it belongs to (to get the `required` array). See `RequiredPropertiesMustExist` rule for an example of how this can be done (that rule runs on each property in the `required` array, and it must access the set of properties that are in the model definition).
### Existing behavior

**Throws**: https://github.com/Azure/autorest/blob/1abf3d061580e2778efcec8f35d0789e1266d3ea/src/modeler/AutoRest.Swagger/SchemaBuilder.cs#L101
#### Why a linter rule?

If we add a validation rule, it will provide information about the location of the rule violation (e.g. which definition it occurred in), which the [VS Code extension](https://github.com/Azure/openapi-lint-extension) can use to underline the proper place in the file.

sarangan12:
Task Completed

PR #1622, opened 2016-11-17
Collection of new linter rules

olydis:
- adding readonly required rule (#1374)


olydis:
Hey @azurecla, it's still me!

PR #1626, opened 2016-11-18
Allow "$schema" property in spec files

bgold09:
* Allowing "$schema" enables intelligent JSON editors like Visual
    Studio to automatically detect the schema to use to validate the
    current file and provide autocompletion.
  * In the AutoRest world, the value of this property should always be
    the URL of some version of schema containing AutoRest extensions to
    the Swagger spec.

### DIFF
--- a/schema/swagger-extensions.json
+++ b/schema/swagger-extensions.json
@@ -15,6 +15,10 @@
     }
   },
   "properties": {
+    "$schema": {
+      "type": "string", 
+      "format": "uri"
+    },
     "swagger": {
       "type": "string",
       "enum": [


fearthecowboy:
The irony of course, is that the moment you do that, it's not technically valid swagger.

bgold09:
@fearthecowboy that's a good point of course. Do you think this is valid then? 

The desire for this change came from the (somewhat minor) frustration of opening specs in VS and always having to paste in the URL for the AutoRest schema since you cannot specify it in the spec file itself.

fearthecowboy:
I'm debating in my brain a lot right now. 

I was talking about this with someone else today too. 

They didn't even add it to OpenAPI 3. Very Sad.

I suspect we may support this, but the Schema at the URL would have to have that change too, otherwise it's not going to be legal swagger .

bgold09:
Interesting. In that case should we close this PR? I suppose I and everyone else will have to continue living with this (somewhat minor) annoyance.

fearthecowboy:
We could close the PR; but I'm starting to lean towards providing a canonical URL that has a tweaked version of the schema just so it *could* be used.

The more I think about it, the more that this arbitrary decision by the OAI is pretty stupid. It doesn't *hurt* to permit the `$schema`, and it would enable tooling where there is nothing now.

bgold09:
Yea I'm not sure that "you should use an editor that supports swagger" is a good enough reason not to allow the $schema property for AutoRest.

fearthecowboy:
I think I'm going to close this; I can't in good faith merge it with the OAI not putting it into their schema, however, I can, at the very least make AutoRest passively accept the presence of it.

We're doing an overhaul over the parsing of swagger (and other content) right now, I'll make sure that we passively permit the `$schema`, but for un-enlightened editor purposes,  we'll have to host a modified version of the Swagger Schemas (2.0 and 3.0) that can be referenced by an editor.

Perhaps if you're up to it, you could grab the swagger 2.0 JSON Schema ( https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v2.0/schema.json ) and add the support for `$schema` into it, and create a PR here in some place like  `/docs/schemas/swagger/2.0/schema.json` ?

PR #1627, opened 2016-11-18
Fixing C# simplifier, and line endings in generated code. 

fearthecowboy:


fearthecowboy:
@olydis take a quick look at this for me -- just commit tagged `Fixed linefeeds, better c# simplifier` -- the rest is regenerated code.

src/core/AutoRest.Core/CodeGenerator.cs
```diff
@@ -117,24 +117,30 @@ private void ResetFileList()
...
                 }
                 FileList.Add(filePath);
-                    Settings.Instance.FileSystem.DeleteFile(filePath);
+                Settings.Instance.FileSystem.DeleteFile(filePath);
             }
             // Make sure the directory exist
             Settings.Instance.FileSystem.CreateDirectory(Path.GetDirectoryName(filePath));
 
+            var lineEnding = filePath.LineEnding();
+
```
olydis: on the other hand, just saw below that we're also using `StringBuilder` and stuff, that's not customizable AFAIK... may just as well be explicit everywhere

src/core/AutoRest/Simplify/CSharpSimplifier.cs
```diff
@@ -92,9 +92,20 @@ private static MetadataReference Mscorlib
                         // get rid of any BOMs 
                         .Trim('\x00EF', '\x00BB', '\x00BF', '\uFEFF', '\u200B');
 
-                    // Write out the files back to their original location
-                    var output = Path.Combine(Settings.Instance.FileSystem.CurrentDirectory, document.Name);
-                    Settings.Instance.FileSystem.WriteFile(output, text);
+
+                    // special cases the simplifier can't handle.
+                    text = text.
```
olydis: and we're sure that no one ever says `default: "[Newtonsoft.Json.JsonObject]"` or `enum: ["Newtonsoft.Json.JsonObject]", "cowbell"]` which would end up in the source code and then get replaced here? ⚡️

fearthecowboy: Yeah, that would not be likely. 
As it is, this is somewhat temporary until we get to write a better simplifier that's more agressive, but not until after dotnet cli RTM ships.

PR #1631, opened 2016-11-22
Various minor bug fixes in the Go generator.

jhendrixMSFT:
Added a plural entry for "containerservice" to prevent type names from
getting improperly truncated (e.g. "ContainerServices" became "s").
Add imports based on property types in the service client template model.
Use GetParameterName() to handle the case where the parameter is a
property on the client type.


jhendrixMSFT:
@mcardosos Here are a few fixes required to generate the SDK with updated swaggers.

fearthecowboy:
You want this to be a PR against the `legacy` branch.

jhendrixMSFT:
@fearthecowboy will fix and submit new PR, thanks!

PR #1632, opened 2016-11-22
Various minor bug fixes in the Go generator.

jhendrixMSFT:
Added a plural entry for "containerservice" to prevent type names from
getting improperly truncated (e.g. "ContainerServices" became "s").
Add imports based on property types in the service client template model.
Use GetParameterName() to handle the case where the parameter is a
property on the client type.
Add support for x-ms-client-flatten by embedding child structs as an
anonymous field.
Add support for x-ms-parameter-location extension.

### DIFF
--- a/src/generator/AutoRest.Go/Extensions.cs
+++ b/src/generator/AutoRest.Go/Extensions.cs
@@ -10,6 +10,7 @@ using Newtonsoft.Json;
 
 using AutoRest.Core.ClientModel;
 using AutoRest.Core.Utilities;
+using AutoRest.Extensions;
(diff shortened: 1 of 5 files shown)


jhendrixMSFT:
@mcardosos Here's the updated PR against the legacy branch.

fearthecowboy:
FYI; make the following change in your branch : https://github.com/Azure/autorest/pull/1635/files 

An update to a dependent package is breaking with the old version of TypeScript

jhendrixMSFT:
@mcardoso I updated the PR with the naming fix.

mcardosos:
LGTM

PR #1634, opened 2016-11-23
Refactored pipeline

olydis:
- `ITransformer` interface for everything that does stuff
- `Schedule` for setting up pipeline (or tree) of transformers
- first class `async` treatment
- uniform error handling (`Schedule` knows which transformer to blame for what, no more try-catch cascades in the generator)


fearthecowboy:
@azuresdkci  retest this please

PR #1635, opened 2016-11-23
Change version of typescript

fearthecowboy:
A recent update to `moment.js` causes older versions of `tsc` to fail during the build/test of nodejs projects.


fearthecowboy:
@amarzavery  any problems with me doing this? (it fixes the build :D)

PR #1641, opened 2016-11-29
Literate YAML/JSON parser

olydis:
Defines a `LiterateYamlParser` that allows mixing markdown with YAML/JSON.
Already used for loading swagger specs if the file ending is `.md`


src/modeler/AutoRest.Swagger/SwaggerParser.cs
```diff
@@ -22,14 +23,19 @@ public static ServiceDefinition Load(string path, IFileSystem fileSystem)
                 throw new ArgumentNullException("fileSystem");
             }
 
-            return SwaggerParser.Parse(fileSystem.ReadFileAsText(path));
+            var swaggerDocument = fileSystem.ReadFileAsText(path);
+            if (path.EndsWith(".md"))
```
fearthecowboy: Make it guess content instead of requiring an extension

olydis: done! funny enough, I caught a bug in a test case with that: `src/dev/TestServer/swagger/azure-special-properties.json` contained *duplicate* `required` keys, which made it invalid YAML/JSON... so it was parsed as markdown, which failed.
Sadly, the `JsonDeserializer` swallowed that bug so far, now it was exposed by the explicit check.

src/modeler/AutoRest.Swagger/SwaggerParser.cs
```diff
@@ -22,14 +23,22 @@ public static ServiceDefinition Load(string path, IFileSystem fileSystem)
                 throw new ArgumentNullException("fileSystem");
             }
 
-            return SwaggerParser.Parse(fileSystem.ReadFileAsText(path));
+            var swaggerDocument = fileSystem.ReadFileAsText(path);
```
fearthecowboy: I'm ok with this for now, but I think long-term we're going to want to get to the point where we can hand the `Parser` a DOM of some sort, and isolate it from even knowing about 'files'

PR #1646, opened 2016-11-30
Code changes to avoid returning nil values from synchrnous wrappers

sarangan12:
This is related to avoid nil return values in the synchronous wrappers. I will send out the detailed explanation seperately. 

 @salameer @veronicagg @vishrutshah Please review.

Related to https://github.com/Azure/azure-sdk-for-ruby/issues/566

### DIFF
--- a/src/generator/AutoRest.Ruby.Azure.Tests/RspecTests/lro_spec.rb
+++ b/src/generator/AutoRest.Ruby.Azure.Tests/RspecTests/lro_spec.rb
@@ -49,8 +49,7 @@ describe 'Long Running Operation' do
   end
 
   it 'should retry on 202 server response in POST request' do
-    result = @lros_client.post202retry200_async(@product).value!
-    expect(result.body).to be_nil
+    result = @lros_client.post202retry200_async(@product).value!    
     expect(result.response.status).to eq(200)
   end
 
@@ -149,7 +148,6 @@ describe 'Long Running Operation' do
 
   it 'should succeed for delete no header in retry' do
     result = @lros_client.delete_no_header_in_retry_async().value!
-    expect(result.body).to be_nil
     expect(result.response.status).to eq(204)
   end
 
@@ -284,13 +282,11 @@ describe 'Long Running Operation with retry' do
 
   it 'should retry DELETE request on 500 response' do
     result = @lroretrys_client.delete202retry200_async().value!
-    expect(result.body).to be_nil
     expect(result.response.status).to eq(200)
   end
 
   it 'should retry POST request on 500 response' do
     result = @lroretrys_client.post202retry200_async(@product).value!
-    expect(result.body).to be_nil
     expect(result.response.status).to eq(200)
   end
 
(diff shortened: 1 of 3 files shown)


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

fearthecowboy:

src/generator/AutoRest.Ruby.Azure.Tests/RspecTests/lro_spec.rb
```diff
@@ -49,8 +49,7 @@
   end
 
   it 'should retry on 202 server response in POST request' do
-    result = @lros_client.post202retry200_async(@product).value!
-    expect(result.body).to be_nil
+    result = @lros_client.post202retry200_async(@product).value!
```
veronicagg: @sarangan12 what are we expecting to come back in the response body? is there something we could check on?

src/generator/AutoRest.Ruby/Model/MethodRb.cs
```diff
@@ -495,14 +495,7 @@ public virtual string ResponseGeneration()
...
             builder.AppendLine("response = {0}_async({1}).value!", Name, MethodParameterInvocation);
-            if (ReturnType.Body != null)
-            {
-                builder.AppendLine("response.body unless response.nil?");
-            }
-            else
-            {
-                builder.AppendLine("nil");
-            }
+            builder.AppendLine("response.body unless response.nil?");
```
veronicagg: @sarangan12 in the previous code, if body wasn't null, then we would return response.body (unless response was null), otherwise we'd return nil, since body was null. With the updated code I see that we're updating the resulting source code, does it change the resulting value? if response.body is null, what do we return? and if it's not, it looks like we're returning the same thing as before. What am I missing? Is there an example that would help see this in action?

src/generator/AutoRest.Ruby.Azure/Templates/AzureMethodTemplate.cshtml
```diff
@@ -67,6 +67,11 @@ end
 @WrapComment("# ", string.Format("@return [{0}] operation results.", Model.OperationReturnTypeString))@:
 @:#
 }
+else
+{
+@WrapComment("# ", string.Format("@return {0} operation results.", Model.Name))@:
```
veronicagg: @sarangan12 could you provide an example of what's being returned? Model.Name seems to be the same as the operation name here, does this match the return type?

src/generator/AutoRest.Ruby.Azure/Templates/AzureMethodTemplate.cshtml
```diff
@@ -92,6 +97,10 @@ def @(Model.Name)_async(@(Model.MethodParameterDeclaration))
     {
       @:@(Model.DeserializePollingResponse("parsed_response", Model.ReturnType.Body))
     }
+    else
+    {
+      @:parsed_response
```
veronicagg: @sarangan12 what's in parsed_response? what ends up in "deserialize_method" and being passed to "get_long_running_operation_results"?

sarangan12:
I am closing this pull request and will be raising a PR against resp api specs to accomodate this request

PR #1649, opened 2016-11-30
Fix parameter validation codegen for flattened types.

jhendrixMSFT:
The codegen for parameter validation wasn't updated to account for type
flattening which introduced regressions in the SDK.  Track if a type was
flattened and if it was generate the appropriate validation code.

### DIFF
--- a/src/generator/AutoRest.Go/Extensions.cs
+++ b/src/generator/AutoRest.Go/Extensions.cs
@@ -611,8 +611,10 @@ namespace AutoRest.Go
                 }
                 else if (property.ShouldBeFlattened())
                 {
-                    // embed as an anonymous struct
+                    // embed as an anonymous struct.  note that the ordering of this clause is
+                    // important, i.e. we don't want to flatten primary types like dictionaries.
                     indented.AppendFormat("*{0} {1}\n", property.Type.Name, property.JsonTag());
+                    property.Extensions[SwaggerExtensions.FlattenOriginalTypeName] = compositeType.Name;
                 }
                 else
                 {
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
@mcardosos Here's the fix for the parameter validation codegen.

mcardosos:
@fearthecowboy Hello! :D
Do you know if we can fix the Jenkins build with [this](https://github.com/Azure/autorest/pull/1635)?

fearthecowboy:
@mcardosos yeah, make that change in your branch too.

jhendrixMSFT:
@mcardosos @fearthecowboy I checked in that change earlier in the week (see https://github.com/Azure/autorest/pull/1632).  This looks to be something different.

jhendrixMSFT:
@fearthecowboy Could you please take a look at this when you have some time?

fearthecowboy:

fearthecowboy:
Seems to pass now!

jhendrixMSFT:
@mcardosos Any comments before I merge?

src/generator/AutoRest.Go/Extensions.cs
```diff
@@ -810,27 +812,37 @@ public static string Validate(this IEnumerable<Parameter> parameters, HttpMethod
...
-                        }  
+                        }
                     }   
                 }
             }
 
             x.AddRange(from prop in ((CompositeType)p.Type).Properties
                        where prop.IsReadOnly
-                       select GetConstraint($"{name}.{prop.Name}", ReadOnlyConstraint, "true"));
+                       select GetConstraint(string.Format("{0}.{1}", name, prop.WasFlattened() ? prop.Type.Name : prop.Name),
```
mcardosos: Can we replace `prop.WasFlattened() ? prop.Type.Name : prop.Name` with `propName`?

mcardosos: It is just this comment :)

jhendrixMSFT: Do you mean propName from above?  If so it's out of scope.

mcardosos: Then everything is OK :D

Issue #1523, opened 2016-10-20
Requirement to model errors as "default" makes it hard to document error cases


brjohnstmsft:
Currently, AutoRest expects that all responses that aren't "default" in the Swagger spec are success cases. However, this makes it difficult to document error cases.

For example, in order to have AutoRest generate the correct code, I need to do this:

``` JSON
"responses": {
  "200": {
    "description": "Description of a successful response.",
    "schema": {
      "$ref": "#/definitions/AdminKeyResult"
    }
  },
  "default": {
    "description": "HTTP 404: The subscription, resource group, or Search service could not be found. HTTP 409: The specified subscription is disabled.",
    "schema": {
      "$ref": "#/definitions/CloudError"
    }
  }
}
```

But what I want to do is something like this:

``` JSON
"responses": {
  "200": {
    "description": "Description of a successful response.",
    "schema": {
      "$ref": "#/definitions/AdminKeyResult"
    }
  },
  "404": {
    "description": "The subscription, resource group, or Search service could not be found.",
    "schema": {
      "$ref": "#/definitions/CloudError"
    }
  },
  "409": {
    "description": "The specified subscription is disabled.",
    "schema": {
      "$ref": "#/definitions/CloudError"
    }
  },
  "default": {
    "description": "An unexpected error occurred.",
    "schema": {
      "$ref": "#/definitions/CloudError"
    }
  }
}
```

Where perhaps AutoRest can tell a response case is an error by the presence of an error contract (`#/definitions/CloudError`).

This is important because as Azure reference documentation moves to `docs.microsoft.com`, we are moving towards auto-generated REST API reference docs that are sourced from the same Swagger specs that we use to generate SDKs. These two processes/toolchains (docs via DocFX/docs.microsoft.com and SDK code via AutoRest) need to work well together.

tbombach:
@sphibbs @johanste FYI for prioritization: this is coming up a lot as teams build out their documentation and different descriptions for different status codes. Also @amarzavery, since we've briefly discussed this.

Context: Historically, AutoRest has influenced how service teams write their Swagger in this area (the recommendation was to merge almost all negative status codes into `default`). This was because AutoRest needed to know when to throw in the operation method vs. when to return normally. I believe this decision was influenced by the language in the spec: http://swagger.io/specification/#responsesObject

>  The container maps a HTTP response code to the expected response
>
>  However, it is expected from the documentation to cover a successful operation response and any known errors.

[...]

PR #1654, opened 2016-12-05
Add treat default type as error

csteegz:
Addresses #1523 

This takes the opinion that there is one canonical error response, which is defined in the swagger as the default response. To reflect that, all responses that share the error response's model are treated as the default case, and ignored.

### DIFF
--- a/src/modeler/AutoRest.Swagger/OperationBuilder.cs
+++ b/src/modeler/AutoRest.Swagger/OperationBuilder.cs
@@ -8,14 +8,13 @@ using System.Globalization;
 using System.Linq;
 using System.Net;
 using System.Text;
-using AutoRest.Core.Model;
 using AutoRest.Core.Logging;
+using AutoRest.Core.Model;
 using AutoRest.Core.Utilities;
 using AutoRest.Swagger.Model;
 using AutoRest.Swagger.Properties;
-using ParameterLocation = AutoRest.Swagger.Model.ParameterLocation;
-
 using static AutoRest.Core.Utilities.DependencyInjection;
+using ParameterLocation = AutoRest.Swagger.Model.ParameterLocation;
 
 namespace AutoRest.Swagger
 {
@@ -51,14 +50,13 @@ namespace AutoRest.Swagger
         {
             EnsureUniqueMethodName(methodName, methodGroup);
 
-            var method = New<Method>(new 
-            {
+            var method = New<Method>(new {
                 HttpMethod = httpMethod,
                 Url = url,
                 Name = methodName,
                 SerializedName = _operation.OperationId
             });
-            
+
             method.RequestContentType = _effectiveConsumes.FirstOrDefault() ?? APP_JSON_MIME;
             string produce = _effectiveConsumes.FirstOrDefault(s => s.StartsWith(APP_JSON_MIME, StringComparison.OrdinalIgnoreCase));
             if (!string.IsNullOrEmpty(produce))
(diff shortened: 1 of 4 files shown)


azurecla:
Hi __@csteegz__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (coverste). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

src/modeler/AutoRest.Swagger/OperationBuilder.cs
```diff
@@ -51,14 +50,13 @@ public Method BuildMethod(HttpMethod httpMethod, string url, string methodName,
         {
             EnsureUniqueMethodName(methodName, methodGroup);
 
-            var method = New<Method>(new 
-            {
+            var method = New<Method>(new {
```
fearthecowboy: While personally, I'd prefer using the `1TBS`, the rest of the project uses the traditional ANSI-style braces, and I'd prefer to keep code changes in a single brace format. 

If you could tweak the braces to be consistent that'd be great.

src/modeler/AutoRest.Swagger/OperationBuilder.cs
```diff
@@ -96,16 +94,12 @@ public Method BuildMethod(HttpMethod httpMethod, string url, string methodName,
 
             var headerTypeName = string.Format(CultureInfo.InvariantCulture,
                 "{0}-{1}-Headers", methodGroup, methodName).Trim('-');
-            var headerType = New<CompositeType>(headerTypeName,new
-            {
+            var headerType = New<CompositeType>(headerTypeName, new {
```
fearthecowboy: Switch to ANSI-style Braces

src/modeler/AutoRest.Swagger/OperationBuilder.cs
```diff
@@ -96,16 +94,12 @@ public Method BuildMethod(HttpMethod httpMethod, string url, string methodName,
...
                 SerializedName = headerTypeName,
                 Documentation = string.Format(CultureInfo.InvariantCulture, "Defines headers for {0} operation.", methodName)
             });
-            responseHeaders.ForEach(h =>
-            {
-                
-                var property = New<Property>(new
-                {
+            responseHeaders.ForEach(h => {
```
fearthecowboy: Switch to ANSI-style Braces

src/modeler/AutoRest.Swagger/OperationBuilder.cs
```diff
@@ -142,8 +136,7 @@ public Method BuildMethod(HttpMethod httpMethod, string url, string methodName,
         private static IEnumerable<SwaggerParameter> DeduplicateParameters(IEnumerable<SwaggerParameter> parameters)
         {
             return parameters
-                .Select(s =>
-                {
+                .Select(s => {
```
fearthecowboy: Switch to ANSI-style Braces

src/modeler/AutoRest.Swagger/OperationBuilder.cs
```diff
@@ -241,8 +261,7 @@ private Response BuildMethodReturnType(List<Stack<IModelType>> types, IModelType
             }
 
             // BuildParameter up type inheritance tree
-            types.ForEach(typeStack =>
-            {
+            types.ForEach(typeStack => {
```
fearthecowboy: ANSI Braces Please.

src/modeler/AutoRest.Swagger/OperationBuilder.cs
```diff
@@ -222,10 +215,37 @@ private List<Stack<IModelType>> BuildResponses(Method method, CompositeType head
...
             return typesList;
         }
 
+        private void RemoveResponsesWithDefaultType(Method method, List<Stack<IModelType>> typesList)
+        {
+            var errorModelType = method.DefaultResponse.Body as CompositeType;
+            if (errorModelType == null || method.Responses == null) return;
+            var keys = method.Responses.Keys;
+            var removeResponses = new List<HttpStatusCode>();
+            foreach (var key in keys)
```
fearthecowboy: The response types for error conditions should be _related_  (ie, default uses `#/definitions/SomeError` and all the other error responses should use something that either returns that type, or has an allOf: "#/definitions/SomeError" in it's type declaration.)

That way we can have very specific types for errors, and makes it very clear what's an acceptable error.

Of course that means we have to do three other things:
- make sure that use of the `allOf` doesn't force type-inheritance, it's possible that we should be able to have two error types that bubble up to the  same swagger model type (that doesn't even have any properties), but still generate individual classes without a hierarchy.

- change the generator to throw the correct exception type, rather than living with the default class

- and add method documentation for known possible exceptions

olydis:
closing due to inactivity (there are issues tracking and referring to this PR)

Issue #1656, opened 2016-12-05
[Bug] "Authentication failed because the remote party has closed the transport stream exception" exception reading swagger


stankovski:
Some HTTPS endpoints that use TLS throw "Authentication failed because the remote party has closed the transport stream exception" when reading swagger documents.

Example endpoint: https://api.mobile.azure.com/preview/swagger.json

fearthecowboy:
This should not be happening in platforms that support netcore now.

PR #1657, opened 2016-12-05
Fix for #1656 - "Authentication failed because the remote party has closed the transport stream exception"

stankovski:


stankovski:
Looks like Tls11 and Tls12 enum variables are not available on Linux. Is there a compilation variable that can be checked to exclude these when building under Linux?

fearthecowboy:
@stankovski 

Couple of questions:

>> Looks like `Tls11` and `Tls12` enum variables are not available on Linux. Is there a compilation variable that can be checked to exclude these when building under Linux?

I am looking into the whole build process at this very minute, and looking to see what .netstandard we're settling on.  I'll add that to my list.

>> `SecurityProtocolType.Ssl3`

`SSL3` has been removed from the default in net46 and beyond (and `tls11`->`tls13` are on), specifically because `SSL3` has vulnerabilities that are not addressable (ie, POODLE).

It's not `SSL3` that you're trying to enable here is it? (IIRC, .net 4.5 doesn't have all the tls stuff enabled, but still had `SSL3` turned on ).

If I can solve the linux build issue--which would likely be moving to .net core and/or .net46, that should solve the missing `tls` variants _provided that netstandard did it right..._  (and close down the `SSL3` variant)

stankovski:
For our endpoint we just need `tls11` or `tls12`. I've added `ssl3` in case there is a service that is exposed behind `ssl3`. No concerns if you remove `ssl3`.

fearthecowboy:
Closing this -- now that we're on .net core I don't think this is a problem at all with new builds?

PR #1668, opened 2016-12-07
[Ruby] Adding resource group accessor to all the resources

sarangan12:
Related to https://github.com/Azure/azure-sdk-for-ruby/issues/340 

@veronicagg @vishrutshah @salameer Please review.

### DIFF
--- a/src/generator/AutoRest.Ruby.Azure/Model/MethodRba.cs
+++ b/src/generator/AutoRest.Ruby.Azure/Model/MethodRba.cs
@@ -270,5 +270,32 @@ namespace AutoRest.Ruby.Azure.Model
                 return base.OperationReturnTypeString;
             }
         }
+
+        /// <summary>
+        /// Creates deserialization logic for the given <paramref name="type"/>.
+        /// </summary>
+        /// <param name="type">Type for which deserialization logic being constructed.</param>
+        /// <param name="valueReference">Reference variable name.</param>
+        /// <param name="responseVariable">Response variable name.</param>
+        /// <returns>Deserialization logic for the given <paramref name="type"/> as string.</returns>
+        public override string GetDeserializationString(IModelType type, string valueReference = "result", string responseVariable = "parsed_response")
+        {
+            var builder = new IndentedStringBuilder("  ");
+
+            string deserializedString = base.GetDeserializationString(type, valueReference, responseVariable);
+            builder.AppendLine(deserializedString);
+
+            if (type is CompositeType)
+            {
+                var composite = type as CompositeType;
+                if (composite.BaseModelType != null && composite.BaseModelType.Name == "Resource")
+                {
+                    builder.AppendLine("{0}.resource_group = resource_group_name if defined?resource_group_name", valueReference);
+                    builder.AppendLine("{0}", valueReference);
+                }
+            }
+
+            return builder.ToString();
+        }
     }
 }
--- a/src/generator/AutoRest.Ruby/Model/MethodRb.cs
+++ b/src/generator/AutoRest.Ruby/Model/MethodRb.cs
@@ -581,7 +581,7 @@ namespace AutoRest.Ruby.Model
         /// <param name="responseVariable">Response variable name.</param>
         /// <returns>Deserialization logic for the given <paramref name="type"/> as string.</returns>
         /// <exception cref="ArgumentNullException">Thrown when a required parameter is null.</exception>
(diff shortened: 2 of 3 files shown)


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

src/generator/AutoRest.Ruby/Model/MethodRb.cs
```diff
@@ -606,6 +606,16 @@ public string GetDeserializationString(IModelType type, string valueReference =
                 builder.AppendLine("{1} = @client.deserialize(result_mapper, {0}, '{1}')", responseVariable, valueReference);
             }
 
+            if (type is CompositeType)
+            {
+                var composite = type as CompositeType;
+                if (composite.BaseModelType != null && composite.BaseModelType.Name == "Resource")
```
veronicagg: We talked offline about the hard-coded string "resource_group_name" and I'm not sure of a way of obtaining this as a variable (for it not to be hard-coded), it seems to be a convention for it to be named that way, so it's probably ok.

Though, should these lines be added to the code only if "resource_group_name" is one of the parameters of the method in which we're adding them? (as that's where the value is coming from)

src/generator/AutoRest.Ruby/Model/MethodRb.cs
```diff
@@ -606,6 +606,16 @@ public string GetDeserializationString(IModelType type, string valueReference =
                 builder.AppendLine("{1} = @client.deserialize(result_mapper, {0}, '{1}')", responseVariable, valueReference);
             }
 
+            if (type is CompositeType)
```
veronicagg: would this added code be only applicable to Azure related methods? If so, should this be part of AutoRest.Ruby.Azure, instead of AutoRest.Ruby?

veronicagg:
Also, please regenerate AutoRest samples with the change.

src/generator/AutoRest.Ruby/Model/MethodRb.cs
sarangan12: Agreed that the resource group is an azure concept. Thanks for pointing that. I have changed the code with the latest commit

src/generator/AutoRest.Ruby/Model/MethodRb.cs
sarangan12: I think we would be in a much bigger trouble if anyone changes this name. Also, this name is actually hardcoded already in Swagger parser. Also, remember the defined? has been added as a safeguard to avoid any issues.

sarangan12:
@veronicagg Regenerated samples using gulp regenerate:expected:samples command

src/generator/AutoRest.Ruby/Model/MethodRb.cs
veronicagg: Agreed, though with this question I was wondering if we could avoid writing these lines of code in the file if there's no resource_group_name coming as parameter of the method. For example the following operation would end up with this added line of code, but it really doesn't apply to the method. Operation: https://github.com/Azure/azure-rest-api-specs/blob/master/arm-compute/2016-03-30/swagger/compute.json#L294, which in generated code looks like this https://github.com/Azure/azure-sdk-for-ruby/blob/master/management/azure_mgmt_compute/lib/generated/azure_mgmt_compute/virtual_machine_extension_images.rb#L163 
I understand it would still work because the property is not defined, but it's really not applicable.

fearthecowboy:

src/generator/AutoRest.Ruby.Azure/Model/MethodRba.cs
```diff
@@ -270,5 +270,32 @@ public override string OperationReturnTypeString
...
+        /// </summary>
+        /// <param name="type">Type for which deserialization logic being constructed.</param>
+        /// <param name="valueReference">Reference variable name.</param>
+        /// <param name="responseVariable">Response variable name.</param>
+        /// <returns>Deserialization logic for the given <paramref name="type"/> as string.</returns>
+        public override string GetDeserializationString(IModelType type, string valueReference = "result", string responseVariable = "parsed_response")
+        {
+            var builder = new IndentedStringBuilder("  ");
+
+            string deserializedString = base.GetDeserializationString(type, valueReference, responseVariable);
```
sarangan12: base is used in C#. not super

sarangan12:
This PR is no longer required. Closing this now.

PR #1675, opened 2016-12-13
added support for x-ms-examples in the schema

amarzavery:


schema/swagger-extensions.json
```diff
@@ -1589,6 +1592,51 @@
...
+        "parameters": {
+          "type": "object",
+          "description": "A dictionary of parameter names and their sample values for a particular operation.",
+          "additionalProperties": {
+            "type": ["object", "number", "array", "integer", "string", "boolean", "null"]
+          }
+        },
+        "responses":{
+          "type": "object",
+          "description": "A dictionary of {possible HTTP status codes (as the dictionary key)} that can be recieved from that operation and {examples (as the dictionary value)} of response {body} and {headers (for long running operations)}.",
```
TimLovellSmith: typo, should be 'received'

schema/swagger-extensions.json
```diff
@@ -1589,6 +1592,51 @@
...
       },
       "uniqueItems": true
+    },
+    "xmsResponseExample": {
+      "type": "object",
+      "description": "Provides the format of specifying an example response in the x-ms-examples extension.",
+      "properties": {
+        "headers": {
+          "type": "object",
+          "description": "Examples of response headers like 'location', 'azure-asyncoperation', 'retry-after', etc.",
```
TimLovellSmith: >'location', 'azure-asyncoperation', 'retry-after', etc.", 

Examples should be in the canonical HTTP casing: 'Location', 'Azure-AsyncOperation', 'Retry-After'

schema/swagger-extensions.json
```diff
@@ -1589,6 +1592,51 @@
...
+          "additionalProperties": {
+            "type": "string"
+          }
+        },
+        "body":{
+          "type": ["object", "number", "array", "integer", "string", "boolean", "null"],
```
TimLovellSmith: Is specifying the type necessary? It seems to not actually be restricting the type in practice.

TimLovellSmith: Could perhaps use further description - If 'null' is specified or 'string' is specified, I wonder how they are interpreted? An empty body, or a body containing the text 'null' (which I have seen some servers return :p) ? Would a string value be assumed to be plain text of the body?

schema/swagger-extensions.json
```diff
@@ -1589,6 +1592,51 @@
...
+          "description": "A dictionary of parameter names and their sample values for a particular operation.",
+          "additionalProperties": {
+            "type": ["object", "number", "array", "integer", "string", "boolean", "null"]
+          }
+        },
+        "responses":{
+          "type": "object",
+          "description": "A dictionary of {possible HTTP status codes (as the dictionary key)} that can be recieved from that operation and {examples (as the dictionary value)} of response {body} and {headers (for long running operations)}.",
+          "additionalProperties": {
+            "$ref": "#/definitions/xmsResponseExample"
```
veronicagg: should the schema of the response be more explicit with patternProperties (for the status codes) and object for their values?

schema/swagger-extensions.json
veronicagg: should body be of type object?

schema/swagger-extensions.json
veronicagg: where's status code of the response currently being modeled?

schema/swagger-extensions.json
```diff
@@ -1589,6 +1592,51 @@
...
+    },
+    "xmsExampleObject": {
+      "type": "object",
+      "required": ["parameters", "responses"],
+      "description": "Describes the format of an example defined in the x-ms-examples extension.",
+      "properties": {
+        "parameters": {
+          "type": "object",
+          "description": "A dictionary of parameter names and their sample values for a particular operation.",
+          "additionalProperties": {
```
amarzavery: The schema is a static thing. I do not know what kind of parameters would one want to provide. I mean, we surely know some common one's like subscriptionId, api-version, resourceGroupName. But these common one's are not always present as well. AD Graph does not have resourceGroupName and subscriptionId. 

We can go either way:
1 - The current approach that is taken in the PR
2 - Model the commonly used parameters as optional properties and set additionalProperties to true (so people can add more parameters).

What do you think?

TimLovellSmith:
:shipit:

PR #1677, opened 2016-12-14
[Ruby] Making credentials nil by default

sarangan12:
Related to https://github.com/Azure/azure-sdk-for-ruby/issues/613 

As mentioned in the issue, even though there is no security/security definition provided in the swagger files, our autorest code actually generates the template with credentials. More than that, it makes credentials a non nullable parameter. This simply is not required. (If we do compare this situation with similar node sdk, the credentials is merely an optional parameter). 

So, there are 2 possible solutions to this problem:
1. Check for the presence of security related information in the swagger file and then decide if the credentials must be added. Though I feel this approach is correct, currently the security related information is not passsed from servicedefinition object to code model object. This would be a bigger code change and the possibility of someone using autorest to generate code for their service (which does not need any authentication) is very less. So, I am not taking this approach. 

2. The second approach is much simpler and aligns with the node approach is to make the credentials parameter nil by default. Also, if we do verify the places (in this file, in service azure client - ms-rest-azure and service client - ms-rest), there are lines specifically written to check if the credentials object is nil. Thus, there is no danger in making this parameter nil by default. 

@veronicagg @vishrutshah @salameer Please review.

### DIFF
--- a/Samples/azure-storage/Azure.CSharp/StorageAccountsOperations.cs
+++ b/Samples/azure-storage/Azure.CSharp/StorageAccountsOperations.cs
@@ -264,11 +264,8 @@ namespace Petstore
         public async Task<AzureOperationResponse<StorageAccount>> CreateWithHttpMessagesAsync(string resourceGroupName, string accountName, StorageAccountCreateParameters parameters, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken))
         {
             // Send Request
-            AzureOperationResponse<StorageAccount> _response = await BeginCreateWithHttpMessagesAsync(
-                resourceGroupName, accountName, parameters, customHeaders, cancellationToken);
-            return await Client.GetPutOrPatchOperationResultAsync(_response,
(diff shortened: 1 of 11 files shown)


azurecla:
    <p>
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

veronicagg:
@sarangan12 since the original issue mentions MsRest:ServiceClient credentials, should we be making credentials optional here: https://github.com/Azure/azure-sdk-for-ruby/blob/master/runtime/ms_rest/lib/ms_rest/service_client.rb#L30 ? And if so, do we still require the proposed change in ServiceClientTemplate.cshtml?

sarangan12:
@veronicagg Yes. We still need the changes as the code here says credentials is a mandatory parameter (the nil check). The ms_rest code can have credentials = nil. But, it is not absolutely required. I will raise a PR to make it optional in ms_rest.

sarangan12:
@veronicagg Regenerated samples

fearthecowboy:

PR #1681, opened 2016-12-16
Safer structural equality

olydis:
might be slower but
- is simpler
- fixes potential stack overflow exception on recursive types

### DIFF
--- a/src/core/AutoRest.Core/Model/IModelType.cs
+++ b/src/core/AutoRest.Core/Model/IModelType.cs
@@ -142,8 +142,17 @@ namespace AutoRest.Core.Model
         /// <returns>true if the specified object is functionally equal to this object; otherwise, false.</returns>
         public virtual bool StructurallyEquals(IModelType other)
         {
-            var ta = JsonConvert.SerializeObject(this);
-            var tb = JsonConvert.SerializeObject(other);
+            if (ReferenceEquals(other, null))
+            {
+                return false;
+            }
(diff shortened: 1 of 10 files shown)


olydis:
@amarzavery ping

src/core/AutoRest.Core/Model/IModelType.cs
```diff
@@ -142,12 +142,9 @@ public virtual void Disambiguate()
         /// <returns>true if the specified object is functionally equal to this object; otherwise, false.</returns>
         public virtual bool StructurallyEquals(IModelType other)
         {
-            if (ReferenceEquals(other, null))
-            {
-                return false;
-            }
-
-            return GetType() == other.GetType() && Name.Equals(other.Name);
+            var ta = JsonConvert.SerializeObject(this);
```
fearthecowboy: Ouch!

Some considerations:
- this uses the default serialization rules where circular references will throw an exception. We overcome this in the `CodeModel` stuff by generating `$id` and `$ref` fields to allow multiple references to the same object to resolve correctly.
- even if you used the serializer settings, two structurally identical objects may not be literally identical, if dictionaries are involved, and the references aren't generated in exactly the same order (since, the first reference shows the entire object, and the second will get a `$ref` to the `$id` of the first.)

fearthecowboy: And, I'd still keep the `ReferenceEquals ` check at the top....

PR #1684, opened 2016-12-16
Refactor Go generator to work with the new AutoRest code model.

jhendrixMSFT:
This is a basic port of the existing Go generator code however some of the
previous code had been rendered obsolete in the new model so it's been
removed.  Also cleaned up some dead code and refactored a bunch of
extension methods into member functions of their respective types.


jhendrixMSFT:
@fearthecowboy @mcardosos @marstr Here's the PR for the Go generator refactoring.  I think there's a lot of room for improvement however this gets us on master so we can start generating SDKs again.

src/generator/AutoRest.Go/Model/EnumTypeGo.cs
```diff
@@ -2,28 +2,39 @@
...
+        public EnumTypeGo(EnumType source) : this()
+        {
+            this.LoadFrom(source);
+        }
+
+        public string GetEmptyCheck(string valueReference, bool asEmpty)
+        {
+            return string.Format(asEmpty
+                                    ? "len(string({0})) == 0"
+                                    : "len(string({0})) > 0", valueReference);
```
marstr: Your call, and there may be something I'm missing, but would this be more readable?
`asEmpty ? "string({0}) == \"\"" : "string({0}) != \"\""`

marstr: Meh.. actually looking at it now, the fact that you have to escape the quotes kinda reduces any benefit of readability. I'm pretty neutral on which would be the better choice.

src/generator/AutoRest.Go/Model/EnumTypeGo.cs
```diff
@@ -2,28 +2,39 @@
...
+        }
+
+        public bool IsNamed()
+        {
+            return Name != "string" && Values.Count > 0;
```
marstr: Values.Any()

src/generator/AutoRest.Go/Model/MethodGo.cs
```diff
@@ -396,5 +411,99 @@ public string ValidationError()
...
             return $"validation.NewErrorWithValidationError(err, \"{PackageName}.{Owner}\",\"{ScopedName}\")";
         }
+
+        /// <summary>
+        /// Check if method has a return response.
+        /// </summary>
+        /// <returns></returns>
+        public bool HasReturnValue()
+        {
+            return ReturnValue() != null && ReturnValue().Body != null;
```
marstr: Could be shortened to the new syntax:
`return null != ReturnValue()?.Body;`

src/generator/AutoRest.Go/Model/MethodGo.cs
```diff
@@ -1,40 +1,48 @@
...
-namespace AutoRest.Go.TemplateModels
+namespace AutoRest.Go.Model
 {
-    public class MethodTemplateModel : Method
+    public class MethodGo : Method
     {
-        public readonly MethodScopeProvider MethodScope;
-        public readonly string Owner;
-        public readonly string PackageName;
+        private MethodScopeProvider _scope;
```
marstr: I believe this should be `readonly`

src/generator/AutoRest.Go/Model/MethodGroupGo.cs
```diff
@@ -0,0 +1,97 @@
...
+        public string Documentation;
+        public string PackageName;
+        public string BaseClient;
+
+        public string GlobalParameters;
+        public string HelperGlobalParameters;
+        public IEnumerable<string> Imports { get; private set; }
+
+        public MethodGroupGo(string name): base(name)
+        {
```
marstr: [nit] I personally like to add "//Intentionally Left Blank" to method bodies

src/generator/AutoRest.Go/Model/MethodGroupGo.cs
```diff
@@ -0,0 +1,97 @@
...
+using System.Linq;
+using AutoRest.Core.Model;
+using AutoRest.Core.Utilities;
+using AutoRest.Go;
+
+namespace AutoRest.Go.Model
+{
+    public class MethodGroupGo : MethodGroup
+    {
+        public string ClientName;
```
marstr: Should these be public get private set?

src/generator/AutoRest.Go/Model/ParameterGo.cs
```diff
@@ -0,0 +1,238 @@
...
+        /// Refactor -> Probably CodeModeltransformer, but even with 5 references, the other mkethods are not used anywhere
+        /// </summary>
+        /// <param name="parameters"></param>
+        /// <param name="location"></param>
+        /// <returns></returns>
+        public static IEnumerable<ParameterGo> ByLocation(this IEnumerable<ParameterGo> parameters, ParameterLocation location)
+        {
+            return parameters
+                .Where(p => p.Location == location)
+                .AsEnumerable();
```
marstr: I don't think this is necessary, wouldn't just passing back the result of the .Where() sufficient?

src/generator/AutoRest.Go/Model/ParameterGo.cs
```diff
@@ -0,0 +1,238 @@
...
+        /// Refactor -> CodeModelTransformer, still, 3 erefences, but no one uses the other methods.
+        /// </summary>
+        /// <param name="parameters"></param>
+        /// <param name="location"></param>
+        /// <returns></returns>
+        public static IEnumerable<ParameterGo> ByLocationAsRequired(this IEnumerable<ParameterGo> parameters, ParameterLocation location, bool isRequired)
+        {
+            return parameters
+                .Where(p => p.Location == location && p.IsRequired == isRequired)
+                .AsEnumerable();
```
marstr: Same as above

src/generator/AutoRest.Go/Model/ParameterGo.cs
```diff
@@ -0,0 +1,238 @@
...
+
+        /// <summary>
+        /// Return list of parameters as per their location in request.
+        /// </summary>
+        /// <param name="parameters"></param>
+        /// <returns></returns>
+        public static ParameterGo BodyParameter(this IEnumerable<ParameterGo> parameters)
+        {
+            var bodyParameters = parameters.ByLocation(ParameterLocation.Body);
+            return bodyParameters.Count() > 0
```
marstr: bodyParameters.Any()

src/generator/AutoRest.Go/Model/PrimaryTypeGo.cs
```diff
@@ -0,0 +1,147 @@
...
+            };
+        }
+
+        /// <summary>
+        /// Add imports for primary type.
+        /// </summary>
+        /// <param name="imports"></param>
+        public void AddImports(HashSet<string> imports)
+        {
+            if (Import != string.Empty)
```
marstr: I know we declare it and it is safe, but I'd still make this `string.IsNullOrEmpty()`

src/generator/AutoRest.Go/Model/PrimaryTypeGo.cs
```diff
@@ -0,0 +1,147 @@
...
+            if (this.PrimaryType(KnownPrimaryType.ByteArray))
+            {
+                return string.Format(asEmpty
+                                        ? "{0} == nil || len({0}) == 0"
+                                        : "{0} != nil && len({0}) > 0", valueReference);
+            }
+            else if (this.PrimaryType(KnownPrimaryType.String))
+            {
+                return string.Format(asEmpty
+                                        ? "len({0}) == 0"
```
marstr: Thoughts on this?
`"{0} == \"\"" : "{0} != \"\""`

src/generator/AutoRest.Go/TransformerGo.cs
```diff
@@ -0,0 +1,258 @@
...
+            TransformModelTypes(cmg);
+
+            return cmg;
+        }
+
+        private void TransformEnumTypes(CodeModelGo cmg)
+        {
+            // fix up any enum types that are missing a name.
+            // NOTE: this must be done before the next code block
+            foreach (var mt in cmg.ModelTypes)
```
marstr: I think the block here on line 48 and the block on 66 should have matching style, either in Linq or foreach blocks.

src/generator/AutoRest.Go/TransformerGo.cs
```diff
@@ -0,0 +1,258 @@
...
+                {
+                    mtm.IsResponseType = true;
+                    if (cmg.PagedTypes.ContainsKey(mtm))
+                    {
+                        mtm.NextLink = CodeNamerGo.PascalCaseWithoutChar(cmg.PagedTypes[mtm], '.');
+                        if (cmg.NextMethodUndefined.Contains(mtm)) {
+                            mtm.PreparerNeeded = true;
+                        } else {
+                            mtm.PreparerNeeded = false;
+                        }
```
marstr: The above could be shortened to:
`mtm.PreparerNeeded = cmg.NextMethodUndefined.Contains(mtm);`

src/generator/AutoRest.Go/TransformerGo.cs
```diff
@@ -0,0 +1,258 @@
...
+            exportedTypes.UnionWith(cmg.EnumTypes);
+            exportedTypes.UnionWith(cmg.Methods);
+            exportedTypes.UnionWith(cmg.ModelTypes);
+
+            var stutteringTypes = exportedTypes
+                                    .Where(exported =>
+                                        (exported is IModelType && (exported as IModelType).Name.FixedValue.StartsWith(cmg.Namespace, StringComparison.InvariantCultureIgnoreCase)) ||
+                                        (exported is Method && (exported as Method).Name.FixedValue.StartsWith(cmg.Namespace, StringComparison.InvariantCultureIgnoreCase)));
+
+            if (stutteringTypes.Count() > 0)
```
marstr: `stutteringTypes.Any()`

src/generator/AutoRest.Go/TransformerGo.cs
```diff
@@ -0,0 +1,258 @@
...
+            {
+                Logger.Instance.Log(Category.Warning, string.Format(CultureInfo.InvariantCulture, Resources.NamesStutter, stutteringTypes.Count()));
+                stutteringTypes.ForEach(exported =>
+                    {
+                        var name = exported is IModelType
+                                        ? (exported as IModelType).Name
+                                        : (exported as Method).Name;
```
marstr: Do we have to do this because of shadowing? I don't run across examples of this too often, but dang, it makes me feel like Golang has the right approach to inheritance :p

src/generator/AutoRest.Go/CodeNamerGo.cs
```diff
@@ -0,0 +1,491 @@
...
+        public new static CodeNamerGo Instance
+        {
+            get
+            {
+                return (CodeNamerGo)CodeNamer.Instance;
+            }
+        }
+
+        // Refactor -> generator
+        public static readonly List<string> AutorestImports = new List<string> { "github.com/Azure/go-autorest/autorest" };
```
fearthecowboy: >ic readonly List<string> [](start = 19, length = 24)

If the intent is to have an immutable set here, this should be declared as IEnumerable<string> :

public static IEnumerable<string> AutorestImports = new string[] { "github.com/Azure/go-autorest/autorest" };

or,  as a Property rather than a static field:

public static IEnumerable<string> AutorestImports {get {  yield return "github.com/Azure/go-autorest/autorest" } };

Either way, the way it's done, the List<string> itself is still mutable.

src/generator/AutoRest.Go/MethodScopeProvider.cs
```diff
@@ -6,6 +6,7 @@
 
 namespace AutoRest.Go
 {
+    /// Refactor -> Namer
     public class MethodScopeProvider
```
fearthecowboy: This sort of thing shouldn't be necessary anymore, disambiguation is done on the fly. We should probably review why this is still here and find a way to remove it.

src/generator/AutoRest.Go/Model/CompositeTypeGo.cs
```diff
@@ -0,0 +1,217 @@
...
+        // (null or empty if the model is not paged).
+        // Refactor -> This is used for too many things... Ideally it should be in the CodeModelTransformer, and later used in the namer 
+        public string NextLink;
+
+        // Refactor -> CodeModelTransformer
+        public bool PreparerNeeded = false;
+
+        public CompositeTypeGo()
+        {
+
```
fearthecowboy: Eeeew. no. :p

---
In reply to: [92900124](https://github.com/Azure/autorest/pull/1684#discussion_r92900124) [](ancestors = 92900124)

src/generator/AutoRest.Go/CodeNamerGo.cs
fearthecowboy: >public static [](start = 8, length = 13)

Plus we're trying to get away from public static *anything* in the namer classes. They should be `public virtual` properties, so that they may be overriden in subclasses when necessary (and `Namer.Instance ` is the static accessor to the current instance.

src/generator/AutoRest.Go/Model/CompositeTypeGo.cs
```diff
@@ -0,0 +1,217 @@
...
+            else
+            {
+                return type;
+            }
+        }
+
+        public string PreparerMethodName()
+        {
+            return Name + "Preparer";
+        }
```
fearthecowboy: Methods like these should be properties, not methods, and can be specified as :

public string PreparerMethod => $"{Name}Preparer";

src/generator/AutoRest.Go/Model/CompositeTypeGo.cs
```diff
@@ -0,0 +1,217 @@
...
+                else
+                {
+                    indented.AppendFormat("{0} *{1} {2}\n", property.Name, property.ModelType.Name, property.JsonTag());
+                }
+            }
+
+            return indented.ToString();
+        }
+
+        public bool IsWrapperType { get { return _wrapper; } }
```
fearthecowboy: You've got a lot of properties that can be simplified using the lambda syntax:

public bool IsWrapperType => _wrapper;

src/generator/AutoRest.Go/Model/CodeModelGo.cs
```diff
@@ -0,0 +1,150 @@
...
+                                            !string.IsNullOrEmpty(Settings.Instance.PackageVersion)
+                                                    ? Settings.Instance.PackageVersion
+                                                    : "0.0.0");
+
+        public override string Namespace
+        {
+            get
+            {
+                if (string.IsNullOrEmpty(base.Namespace))
+                    return base.Namespace;
```
fearthecowboy: Code Format: you have a number of if( .. ) statements that should have { } for the body. We don't like 

if( ... ) 
  something;

please use 

if( ... ) 
{ 
   something;
}

src/generator/AutoRest.Go/Model/CodeModelGo.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        public string ServiceName
+        {
+            get
+            {
+                if (!string.IsNullOrEmpty(Namespace))
+                {
+                    return CodeNamerGo.Instance.PascalCase(Namespace);
+                }
+                return string.Empty;
```
fearthecowboy: You could significantly simplify properties like this:

public string ServiceName => CodeNamer.Instance.PascalCase (Namespace ?? String.Empty);

src/generator/AutoRest.Go/Model/CodeModelGo.cs
```diff
@@ -0,0 +1,150 @@
...
+            }
+        }
+
+        public string GetDocumentation()
+        {
+            return string.Format("Package {0} implements the Azure ARM {1} service API version {2}.\n\n{3}", Namespace, ServiceName, ApiVersion,
```
fearthecowboy: You should probably consider moving to interpolated strings :

public string DocumentationText => $"Package {Namespace} implements the Azure ARM {ServiceName} service API version {ApiVersion}.\n\n{(Documentation??string.Empty) .UnrwapAnchorTags()}"

String.Format isn't very fast, subject to typos in parameter ordering, and is kinda wordy :)

src/generator/AutoRest.Go/Model/MethodGo.cs
```diff
@@ -396,5 +411,99 @@ public string ValidationError()
...
+                }
+            }
+            return false;
+        }
+
+        /// <summary>
+        /// Check if method has long running extension (x-ms-long-running-operation) enabled. 
+        /// </summary>
+        /// <returns></returns>
+        public bool IsLongRunningOperation()
```
mcardosos: Not for this PR, but maybe in the future... Looking at the other generators, this is the kind of func that would better live in the core, thoughts? @fearthecowboy

src/generator/AutoRest.Go/Model/PrimaryTypeGo.cs
```diff
@@ -0,0 +1,147 @@
...
+                        return "int64";
+
+                    case KnownPrimaryType.Stream:
+                        return "io.ReadCloser";
+
+                    case KnownPrimaryType.String:
+                        return "string";
+
+                    case KnownPrimaryType.TimeSpan:
+                        return "string";
```
mcardosos: I wonder why we are using `string` instead of `time.Duration`. Again, not to fix in this PR, but another moment

src/generator/AutoRest.Go/Model/PrimaryTypeGo.cs
```diff
@@ -0,0 +1,147 @@
...
+
+                    case KnownPrimaryType.Object:
+                        // TODO: is this the correct way to support object types?
+                        return "map[string]interface{}";
+
+                    case KnownPrimaryType.UnixTime:
+                        // TODO: add support
+                        return "string";
```
mcardosos: The same as `TimeSpan`, the correct type would be `int64`

src/generator/AutoRest.Go/Model/PrimaryTypeGo.cs
mcardosos: Oops, go has no support for https://en.wikipedia.org/wiki/ISO_8601#Durations

src/generator/AutoRest.Go/CodeNamerGo.cs
```diff
@@ -0,0 +1,491 @@
...
+
+        /// <summary>
+        /// Converts names the conflict with Go reserved terms by appending the passed appendValue.
+        /// </summary>
+        /// <param name="name">Name.</param>
+        /// <param name="appendValue">String to append.</param>
+        /// <returns>The transformed reserved name</returns>
+        protected override string GetEscapedReservedName(string name, string appendValue)
+        {
+            if (name == null)
```
mcardosos: Maybe `IsNullOrWhitespace` is better for this?

src/generator/AutoRest.Go/CodeNamerGo.cs
```diff
@@ -0,0 +1,491 @@
...
+            if (ReservedWords.Contains(name, StringComparer.Ordinal))
+            {
+                name += appendValue;
+            }
+
+            return name;
+        }
+
+        // Refactor -> Namer
+        public void ReserveNamespace(string ns)
```
mcardosos: Another note for future PRs... I don't think the Namespace needs to be added to reserved words. For example, [Container Service SDK](https://github.com/Azure/azure-sdk-for-go/tree/master/arm/containerservice) has a struct named ContainerService (somewhere else in the code we managed to be able to use the namespace for types).
The namespace is only useful in the namer to avoid golint stuttering.

src/generator/AutoRest.Go/CodeNamerGo.cs
```diff
@@ -0,0 +1,491 @@
...
+                    {
+                        return defaultValue.ToLowerInvariant();
+                    }
+                    else if (primaryType.KnownPrimaryType == KnownPrimaryType.ByteArray)
+                    {
+                        return "[]bytearray(\"" + defaultValue + "\")";
+                    }
+                    else
+                    {
+                        //TODO: handle imports for package types.
```
mcardosos: I think we are getting rid of package types with this PR :D

src/generator/AutoRest.Go/Model/ParameterGo.cs
```diff
@@ -0,0 +1,238 @@
...
+
+namespace AutoRest.Go.Model
+{
+    public class ParameterGo : Parameter
+    {
+        public const string ApiVersionName = "APIVersion";
+
+        public const string ApiVersionSerializedName = "api-version";
+
+        public const string SkipUrlEncoding = "x-ms-skip-url-encoding";
```
mcardosos: Somewhere in the core this already exists. It is used in L75

src/generator/AutoRest.Go/Model/PrimaryTypeGo.cs
mcardosos: Blergh, go does not support [ISO 8601 duration](https://en.wikipedia.org/wiki/ISO_8601#Durations)

src/generator/AutoRest.Go/TransformerGo.cs
```diff
@@ -0,0 +1,258 @@
...
+            _normalizedTypes = new Dictionary<IModelType, IModelType>();
+        }
+
+        public override CodeModelGo TransformCodeModel(CodeModel cm)
+        {
+            var cmg = cm as CodeModelGo;
+
+            SwaggerExtensions.ProcessGlobalParameters(cmg);
+            // Add the current package name as a reserved keyword
+            CodeNamerGo.Instance.ReserveNamespace(cm.Namespace);
```
mcardosos: Again, I'm not sure this is necessary.

jhendrixMSFT:
I've talked to @mcardosos offline and she will open github issues for the issues she pointed out in this review.
@fearthecowboy @marstr I've fixed the issues you guys pointed out, any other feedback or can we merge?

src/generator/AutoRest.Go/Model/PrimaryTypeGo.cs
marstr: Sounds like something we should investigate/build for the future. I assume we're not the only ones that would like to be able to convert into something natively usable in Go.

jhendrixMSFT:
@fearthecowboy We have sign-off can you please merge?  I don't have write access.

fearthecowboy:
Done!

AutoRest.sln
```diff
@@ -118,7 +118,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Go", "Go", "{1803891D-AB39-
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Core", "Core", "{AB891A22-A69D-46E1-8C75-DB69641D3BBD}"
 EndProject
-Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "AutoRest.Preview", "src\dev\AutoRest.Preview\AutoRest.Preview.xproj", "{B2BB2267-019B-43BB-9225-BCB125BCD6DC}"
```
olydis: hey, that was mean :D

PR #1694, opened 2016-12-22
x-ms-external for CSharp

olydis:
#642


src/core/AutoRest.Extensions.Azure/AzureExtensions.cs
```diff
@@ -32,7 +32,6 @@ public abstract class AzureExtensions : SwaggerExtensions
         public const string AzureResourceExtension = "x-ms-azure-resource";
         public const string ODataExtension = "x-ms-odata";
         public const string ClientRequestIdExtension = "x-ms-client-request-id";
-        public const string ExternalExtension = "x-ms-external";
```
fearthecowboy: Nothing else was using this?

PR #1695, opened 2016-12-22
updated the all-specs file

olydis:


fearthecowboy:
We use it to manually regression-test against different versions of autorest

PR #1697, opened 2016-12-28
Generate Server side code from specs

dsgouda:
Iteration 1 to generate server side code from specs
Still has compilation errors
@fearthecowboy


.gitignore
```diff
@@ -5,6 +5,9 @@ src/generator/AutoRest.NodeJS.Tests/AcceptanceTests/*.js
 ## Ignore user-specific files, temporary files, build results, etc.
 AutoRest/**/Templates/*.cs
 src/**/Templates/*.cs
+src/generator/**/Templates/Rest/Common/*.cs
```
fearthecowboy: wouldn't 
` src/**/Templates/**/*.cs ` work to do this globally?  we're gonna have a ton of ignore lines if we don't

src/generator/AutoRest.CSharp/Model/ControllerToModelBinderMapping.cs
```diff
@@ -0,0 +1,31 @@
+using AutoRest.Core.Model;
+using System.Collections.Generic;
+namespace AutoRest.CSharp.Model
+{
+
+    static class ControllerToModelBinderMapping
+    {
```
fearthecowboy: Urg. I really dislike static dictionaries for translation purposes. It's memory wasteful, a maintenance hazard, not particularly efficient, and the compiler can't do anything to inline it. 

I'd rather go with something a bit slimmer, and simpler:
``` csharp
public static string GetModelBinder(this ParameterLocation parameterLocation )
{
  switch( parameterLocation ) {
    case ParameterLocation.None:
      return string.Empty;
    case ParameterLocation.Path:
      return "FromRoute";
    case ParameterLocation.FormData:
      return "FromForm";
    default:
      return $"From{parameterLocation}";
  }
}
```

Tools/AutoRest.Settings.targets
```diff
@@ -130,6 +130,9 @@
     <Message Text="Compiling Razor templates for project $(MSBuildProjectDirectory)" />
     <!-- Compile razor template. -->
     <Exec Command="$(RazorCommand) &quot;$(MSBuildProjectDirectory)\Templates&quot; $(RootNamespace).Templates" />
+    <Exec Command="$(RazorCommand) &quot;$(MSBuildProjectDirectory)\Templates\Rest\Client&quot; $(RootNamespace).Templates.Client" />
```
dsgouda: Sounds good, I'll let this be here for now.

---
In reply to: [94625465](https://github.com/Azure/autorest/pull/1697#discussion_r94625465) [](ancestors = 94625465)

src/generator/AutoRest.CSharp.Unit.Tests/project.json
```diff
@@ -32,6 +32,8 @@
     "Microsoft.Extensions.Logging": "1.0.0",
     "Microsoft.Extensions.Logging.Console": "1.0.0",
     "Microsoft.Extensions.Logging.Filter": "1.0.0",
+    "Microsoft.AspNetCore": "1.1.0",
```
fearthecowboy: I think you need to upgrade the version of the packages: 
```
Microsoft.Extensions.Logging to 1.1.0 
Microsoft.Extensions.Logging.Console to 1.1.0
Microsoft.NETCore.Platforms to 1.1.0

```

Issue #1660, opened 2016-12-05
Node code gen does breaks enum constants


begoldsm:
In node code gen does not properly enclose constant string values in single quotes. For example, this swagger snippet:
```
{
            "name": "syncFlag",
            "in": "query",
            "required": false,
            "type": "string",
            "default": "DATA",
            "description": "...",
            "enum": [
              "DATA",
              "METADATA",
              "CLOSE"
            ],
            "x-ms-enum": {
            "name": "SyncFlag",
            "modelAsString": false
            }
          },
```
Results in the following line of code in node (note the emphasis):
var syncFlag = (options && options.syncFlag !== undefined) ? **options.syncFlag : DATA;**

When it should result in DATA being enclosed ('DATA'):
var syncFlag = (options && options.syncFlag !== undefined) ? options.syncFlag : **'DATA';**

begoldsm:
@amarzavery FYI this is the issue for the node codegen for constants.

olydis:
Huh isn't this a duplicate of #1597!?

PR #1701, opened 2016-12-29
escape default enum value regardless of ModelAsString

### LINKED ISSUE #1597: Node Generation: Default enum values not properly being turned into strings
If a swagger spec contains a default enum value, node is not properly turning this into a string. Here is a swagger snippet and the corresponding code generated by node:
```
{
            "name": "syncFlag",
            "in": "query",
            "required": false,
            "type": "string",
            "default": "CLOSE",
            "description": "...",
            "enum": [
              "DATA",
              "METADATA",
              "CLOSE"
            ],
            "x-ms-enum": {
            "name": "SyncFlag",
            "modelAsString": false
            }
          }
```

And here is the code generated in node for that (note the double stars for emphasis)
`var syncFlag = (options && options.syncFlag !== undefined) ? options.syncFlag : **CLOSE**;`

That code should actually have CLOSE in single quotes:
`var syncFlag = (options && options.syncFlag !== undefined) ? options.syncFlag : 'CLOSE';`

olydis:
fixes #1597 and #1660

### DIFF
--- a/src/generator/AutoRest.NodeJS/CodeNamerJs.cs
+++ b/src/generator/AutoRest.NodeJS/CodeNamerJs.cs
@@ -101,13 +101,13 @@ namespace AutoRest.NodeJS
                 throw new ArgumentNullException(nameof(type));
             }
 
-            if (true == (type as EnumType)?.ModelAsString)
+            if (type is EnumType)
             {
                 if (!defaultValue.IsNullOrEmpty())
                 {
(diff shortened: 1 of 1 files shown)


olydis:
@amarzavery Can you take a quick look, please? :-)
I've confirmed that @begoldsm's case is fixed (also, note that two basically identical issues were opened for this - or am I missing something?).

PR #1709, opened 2017-01-10
validation suppressions

olydis:
Just a simple way to suppress printing validation messages to the console.
Usage: Add a file called `<input Swagger file>.suppressions` with content like
```
- what: UniqueResourcePaths
  why: I don't care about that
  where: "#/Paths"
- what: NonAppJsonTypeWarning
  why: I don't care about that either
  where: "#/SomewhereElse"
```
(meant to be replaced by *the configuration file* as soon as it exists - maybe this will influence the format used in there)

### DIFF
--- a/src/core/AutoRest.Core/Logging/ConsoleLogListener.cs
+++ b/src/core/AutoRest.Core/Logging/ConsoleLogListener.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using AutoRest.Core.Validation;
+using System;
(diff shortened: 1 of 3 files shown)


olydis:
@amarzavery ping

fearthecowboy:
If this is urgent, Let's look at this this morning, and come up with a format that fits with our ideas in configuration. I really don't want to have to revisit something like this.

fearthecowboy:
Closing -- we're doing this in configuration

PR #1715, opened 2017-01-12
Changes to markdown file generation in AzureResourceSchema generator.

jhendrixMSFT:
The doc team requested the following changes so they can start consuming
the generated markdown help files.

- generate a separate file for each resource type
- use namespace/type instead of just namespace
- include optional properties in the template sample
- move type into its own column
- remove line-breaks from tables
- re-word some of the boiler-plate text
- don't bold text when listing a type's values


jhendrixMSFT:
@marstr PTAL

fearthecowboy:
Have all the `expected` files in `AutoRest.AzureResourceSchema.Tests` been regenerated ? 

Was there no changes to the output?

jhendrixMSFT:
The schemas weren't affected only the markdown docs so no need to update the expected content.

fearthecowboy:
Uh, right... but...

https://github.com/Azure/autorest/tree/master/src/generator/AutoRest.AzureResourceSchema.Tests/Expected/Batch/2015-12-01

There are .md files in the expected section... am I loopy or am I looking at the wrong thing?

jhendrixMSFT:
Oh I didn't realize these were checked in.  Hmm you're right it seems like they should have been updated (how are the tests passing...).  I'll dig deeper.

jhendrixMSFT:
OK tests have been updated and are being properly compared now (it had been disabled :().

src/generator/AutoRest.AzureResourceSchema.Tests/Expected/ApiManagement/2016-07-07/Microsoft.ApiManagement/service_apis_operations.md
```diff
@@ -0,0 +1,141 @@
...
+    "name": "string",
+    "method": "string",
+    "urlTemplate": "string",
+    "templateParameters": [
+      {
+        "name": "string",
+        "description": "string",
+        "type": "string",
+        "defaultValue": "string",
+        "required": boolean,
```
marstr: Shouldn't this be surrounded by quotes?
edit: Along with a bunch of other examples.

src/generator/AutoRest.AzureResourceSchema/CodeGeneratorArs.cs
```diff
@@ -29,12 +29,35 @@ public CodeGeneratorArs()
...
+                await Write(stringWriter.ToString(), resourceProvider + Path.DirectorySeparatorChar + resourceProvider + ".json", true);
 
                 stringWriter = new StringWriter();
-                ResourceMarkdownWriter.Write(stringWriter, resourceSchemas[resourceProvider]);
-                await Write(stringWriter.ToString(), resourceProvider + ".md", false);
+                var md = ResourceMarkdownGenerator.Generate(resourceSchemas[resourceProvider]);
+
+                foreach (var m in md)
+                {
+                    var content = m.Content.Replace("\"boolean\"", "boolean");
```
marstr: Ah, here is this difference. Why is this replace happening? I thought that Boolean values themselves i.e. `true` and `false` didn't have to be in quotes, but I assumed the value `boolean` would

src/generator/AutoRest.AzureResourceSchema/CodeGeneratorArs.cs
```diff
@@ -29,12 +29,35 @@ public CodeGeneratorArs()
...
+                var name = Path.GetFileNameWithoutExtension(fileName);
+                var ext = Path.GetExtension(fileName);
+
+                fileName = string.Format("{0}{1}{2}{3}{4}", subdir, Path.DirectorySeparatorChar, name, i, ext);
+                filePath = Path.Combine(Settings.Instance.OutputDirectory, subdir, fileName);
+
+                ++i;
+            }
+
+            return string.Format("{0}{1}{2}", subdir, Path.DirectorySeparatorChar, fileName);
```
marstr: I'd use `Path.Combine` instead of string.Format for both instances in the loop.

src/generator/AutoRest.AzureResourceSchema.Tests/Expected/ApiManagement/2016-07-07/Microsoft.ApiManagement/service_apis_operations.md
jhendrixMSFT: The doc team specifically asked for the removal of the quotes.

src/generator/AutoRest.AzureResourceSchema/CodeGeneratorArs.cs
jhendrixMSFT: This is the way the doc team wanted this part of the template to look (it does make it a bit easier to read).

jhendrixMSFT:
@fearthecowboy I think this is ready to be merged.

Issue #1716, opened 2017-01-12
Client param with enum emits unquoted default string


olydis:
```
  ApiVersionParameter:
    x-ms-client-name: version
    in: header
    type: string
    x-ms-parameter-location: client
    enum:
    - '2015-04-05'
    - '2014-02-14'
    default: '2015-04-05'
```
emits `this.Version = 2015-04-05;`

Without the `enum`, it emits `this.Version = "2015-04-05";` as expected

PR #1717, opened 2017-01-12
Better preprocessor and 3 bugfixes

olydis:
Bugfixes:
- had a throwing `catch` block in `SwaggerParser` that prevented us from seeing meaningful JSON parsing exceptions
- C# generator emitted `Stream.Null` without any usings
- #1716

### DIFF
--- a/src/generator/AutoRest.CSharp/CodeNamerCs.cs
+++ b/src/generator/AutoRest.CSharp/CodeNamerCs.cs
@@ -80,6 +80,10 @@ namespace AutoRest.CSharp
                 {
                     return "new " + type.Name + "()";
                 }
+                if (type is EnumType && (type as EnumType).ModelAsString)
+                {
+                    return Instance.QuoteValue(defaultValue);
+                }
                 if (primaryType != null)
                 {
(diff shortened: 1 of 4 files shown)


src/generator/AutoRest.CSharp/CodeNamerCs.cs
```diff
@@ -80,6 +80,10 @@ public override string EscapeDefaultValue(string defaultValue, IModelType type)
                 {
                     return "new " + type.Name + "()";
                 }
+                if (type is EnumType && (type as EnumType).ModelAsString)
```
fearthecowboy: Did you do a test run with the compare script? I'm nervous about what happens on all the generators...

olydis: something like another code location relying on the thing being unquoted and adding its own set of quotes? sounds likely, will do :-)

olydis: empty diff for me!

fearthecowboy: Awesome!

PR #1726, opened 2017-01-17
Generator should not flatten out parameters if they are polymorphic

fearthecowboy:


src/modeler/AutoRest.Swagger/OperationBuilder.cs
```diff
@@ -40,7 +40,6 @@ public OperationBuilder(Operation operation, SwaggerModeler swaggerModeler)
             {
                 throw new ArgumentNullException("swaggerModeler");
             }
-
```
olydis: 👍

PR #1731, opened 2017-01-18
[Go] Telemetry tiny detail

mcardosos:
To follow same convention as other SDKs telemetry
@vishrutshah @marstr @jhendrixMSFT


jhendrixMSFT:
As long as this won't negatively affect how we can mine telemetry data it looks good to me.

PR #1732, opened 2017-01-18
[Linter] Relaxing naming rules to accomodate suffix words in names

vishrutshah:


vishrutshah:
@sarangan12 Could you please take a quick look when you get a chance. Thanks!

PR #1756, opened 2017-01-27
Added templates for json rpc client and server

dsgouda:
- New swagger document for the vscode extension specifications
- New templates to generate json rpc server/client 
- Added CodeGenerationMode options 'jsonrpc-client' and 'jsonrpc-server' 
@fearthecowboy PTAL


olydis:
Nice! Do we want tests at this (inofficial) stage or will the actual VS code integration serve as that? :-D

dsgouda:
This is blocking on a dependent jsonrpc library that Garrett is working on, let's put this on the backburner for now

PR #1761, opened 2017-01-27
FileSystem related changes towards isolation and sanity

olydis:


olydis:
will come as part of configuration file support

PR #1762, opened 2017-01-28
Fixing Bug 1710

dsgouda:
@olydis PTAL


src/modeler/AutoRest.Swagger/OperationBuilder.cs
```diff
@@ -118,9 +118,18 @@ public Method BuildMethod(HttpMethod httpMethod, string url, string methodName,
                 headerType = null;
             }
 
+            
             // Response format
             List<Stack<IModelType>> typesList = BuildResponses(method, headerType);
 
+            // if each of the responses have explicitly set their nullable property to false, we will have to consider it
+            bool isNonNullableReturnType = true;
+            _operation.Responses.Keys.ForEach(key =>
```
olydis: can you use LINQ's `All` instead? Should be a side-effects free one liner.

src/generator/AutoRest.CSharp.Unit.Tests/Bug1710.cs
```diff
@@ -0,0 +1,69 @@
...
+    public class Bug1710 : BugTest
+    {
+
+        public Bug1710(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        /// <summary>
+        ///     https://github.com/Azure/autorest/issues/1754
+        ///     Verifies autorest succesfully generates code for empty path object
```
olydis: Wrong comment. (also, links to wrong issue)

src/modeler/AutoRest.Swagger/OperationBuilder.cs
```diff
@@ -148,9 +148,15 @@ public Method BuildMethod(HttpMethod httpMethod, string url, string methodName,
                 headerType = null;
             }
 
+            
             // Response format
             List<Stack<IModelType>> typesList = BuildResponses(method, headerType);
 
+            method.IsXNullableReturnType = (_operation.Responses.Keys.All(key =>
```
olydis: a little over-parenthesized (outer one is superfluous), would also go with expression-only lambda notation

dsgouda: Baby steps in learning Linq, hence the noise!

olydis: generally: does `IsXNullableReturnType` have to be set here? This is very early in the pipeline so response types *might* be changed by later steps and extensions! If possible: make `IsXNullableReturnType` a derived property that simply does precisely what you do here, but precisely at the time you query for it instead of ahead of time.

PR #1765, opened 2017-01-30
Fix generation of ARM markdown help files.

jhendrixMSFT:
Include API specific name property instead of the generic one (it provides
better doc text than the generic "name of the resource" one).
Merge AllOf/Properties into a flat list so that AllOf is included.
Removed EnsureUniqueFileName as it's not needed.
Fix unnecessary nesting output directories.


jhendrixMSFT:
@daschult for the changes in ResourceSchemaParser.cs

daschult:
Looks fine to me, although these are changes mostly in the markdown area, which I don't know a lot about. It should be fine though.

jhendrixMSFT:
@marstr can you take a quick look at the code changes?

src/generator/AutoRest.AzureResourceSchema/ResourceMarkdownGenerator.cs
```diff
@@ -436,5 +450,46 @@ private JToken JsonSchemaToJTokenImpl(string propName, JsonSchema jsonSchema, St
...
+        /// <param name="items">The items to merge into the dictionary.</param>
+        private void FlattenPropertiesImpl(Dictionary<string, JsonSchema> merged, IEnumerable<JsonSchema> items)
+        {
+            foreach (var item in items)
+            {
+                if (item.AllOf != null)
+                {
+                    FlattenPropertiesImpl(merged, item.AllOf);
+                }
+                else if (item.Properties != null)
```
marstr: It's not clear to me that each item will have one of AllOf, Properties, or Ref to flatten. Are there any cases where multiple of these clauses would need to be hit?

src/generator/AutoRest.AzureResourceSchema/ResourceSchemaParser.cs
```diff
@@ -81,8 +81,28 @@ public static class ResourceSchemaParser
...
                     resourceDefinition.JsonType = "object";
-
                     resourceDefinition.ResourceType = resourceType;
+
+                    // get the resource name parameter, e.g. {fooName}
+                    var resNameParam = methodUrlPathAfterProvider.Substring(methodUrlPathAfterProvider.LastIndexOf('/') + 1);
+                    if (IsPathVariable(resNameParam))
+                    {
+                        // strip the enclosing braces
+                        resNameParam = resNameParam.Substring(1, resNameParam.Length - 2);
```
marstr: [nit] Perhaps a more readable version:

``` C#
resNameParam = resNameParam.Trim(new[]{'{', '}'});
```

src/generator/AutoRest.AzureResourceSchema/CodeGeneratorArs.cs
```diff
@@ -29,35 +29,17 @@ public CodeGeneratorArs()
...
+                await Write(stringWriter.ToString(), resourceProvider + ".json", true);
 
                 stringWriter = new StringWriter();
                 var md = ResourceMarkdownGenerator.Generate(resourceSchemas[resourceProvider]);
 
                 foreach (var m in md)
                 {
                     var content = m.Content.Replace("\"boolean\"", "boolean");
-                    await Write(content, EnsureUniqueFileName(resourceProvider, m.Type + ".md"), false);
+                    await Write(content, m.Type + ".md", false);
```
jhendrixMSFT: Yep m.Type is guaranteed to be unique.

src/generator/AutoRest.AzureResourceSchema/ResourceSchemaParser.cs
jhendrixMSFT: Good suggestion I'll make this change.

jhendrixMSFT:
@fearthecowboy Ready for merging.

Issue #1763, opened 2017-01-28
Underscore in property name results in shadowing in generated C# constructor


saikatguha:
**Swagger** model fragment:
```
  "definitions": {
    "Model": {
      "type": "object",
      "properties": {
        "_id": {
          "format": "uuid",
          "type": "string"
        }
      }
    }
```

Generated **C#** model fragment:
```
    public partial class Model
    {
        public Document(System.Guid? _id = default(System.Guid?))
        {
            _id = _id; // <----- Oops! Shadowed
        }

        /// <summary>
        /// </summary>
        [Newtonsoft.Json.JsonProperty(PropertyName = "_id")]
        public System.Guid? _id { get; set; }
    }
```
Results in:
```Warning	CS1717	Assignment made to same variable; did you mean to assign something else?```

Expected qualified property access:
```
            this._id = _id; // <----- Fixed.
```

**Comment by fearthecowboy** [comment]
I recall this being fixed before, but it's possible it was just parameters, not properties.

@dsgouda , can you take a look and make sure it works for properties too?

**Comment by olydis** [comment]
What version of AutoRest are you using? I can't reproduce the bad behavior. Your example prodices precisely:
```C#
public partial class Model
{
    /// <summary>
    /// Initializes a new instance of the Model class.
    /// </summary>
    public Model() { }

    /// <summary>
    /// Initializes a new instance of the Model class.
    /// </summary>
    public Model(System.Guid? _id = default(System.Guid?))
    {
        this._id = _id;
    }

    /// <summary>
    /// </summary>
    [Newtonsoft.Json.JsonProperty(PropertyName = "_id")]
    public System.Guid? _id { get; set; }

}
```

saikatguha:
I was using the latest release (0.17.3)
Looks like it's been fixed since. Thanks!

PR #1766, opened 2017-01-30
Fixed bug 1763 and added unit tests

dsgouda:
fixing https://github.com/Azure/autorest/issues/1763
@fearthecowboy PTAL

### DIFF
--- a/src/generator/AutoRest.CSharp.Unit.Tests/Bug1763.cs
+++ b/src/generator/AutoRest.CSharp.Unit.Tests/Bug1763.cs
@@ -19,7 +19,7 @@ namespace AutoRest.CSharp.Unit.Tests
 
         /// <summary>
         ///     https://github.com/Azure/autorest/issues/1763
-        ///     Verifies autorest succesfully generates code for empty path object
+        ///     Verifies autorest succesfully generates code for models with properties starting with '_'
         /// </summary>
         [Fact]
         public async Task VerifyUnderscoredClassMembers()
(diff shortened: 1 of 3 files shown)


src/generator/AutoRest.CSharp.Unit.Tests/Bug1763.cs
```diff
@@ -0,0 +1,68 @@
...
+{
+    public class Bug1763 : BugTest
+    {
+        public Bug1763(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        /// <summary>
+        ///     https://github.com/Azure/autorest/issues/1763
+        ///     Verifies autorest succesfully generates code for empty path object
```
olydis: Wrong comment.

dsgouda:
@azuresdkci Test this please

src/generator/AutoRest.CSharp.Unit.Tests/Bug1763.cs
```diff
@@ -0,0 +1,79 @@
...
+                // Don't proceed unless we have zero Errors.
+                Assert.Empty(errors);
+
+                // Should also succeed.
+                Assert.True(result.Succeeded);
+
+                // try to load the assembly
+                var asm = Assembly.Load(result.Output.GetBuffer());
+                Assert.NotNull(asm);
+                var petModel = asm.ExportedTypes.First((type) =>
```
olydis: weeeeell I would use short notation here: `asm.ExportedTypes.First(type => type.FullName == "Test.Models.Pet")`
Return-only functions are 90% syntax noise ;-)

PR #1768, opened 2017-01-31
Fixed and improved UniqueResourcePaths linter rule

olydis:
Before:
Raised false alarms (broken regex) but was also too relaxed in other regards.

Now:
Strict, better linter message.

### DIFF
--- a/src/modeler/AutoRest.Swagger/Validation/UniqueResourcePaths.cs
+++ b/src/modeler/AutoRest.Swagger/Validation/UniqueResourcePaths.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System;
 using AutoRest.Core.Logging;
 using AutoRest.Core.Properties;
 using AutoRest.Core.Validation;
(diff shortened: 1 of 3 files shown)


src/modeler/AutoRest.Swagger/Validation/UniqueResourcePaths.cs
```diff
@@ -13,18 +14,22 @@ namespace AutoRest.Swagger.Validation
...
-                    { return apiRegExp.Match(path).Success ? apiRegExp.Match(path).Groups?[0]?.Value.ToString().ToLowerInvariant() : null; })
-                                      .Where(p=>p!=null).Distinct().Count() <= 1;
+        public override bool IsValid(Dictionary<string, Dictionary<string, Operation>> paths, RuleContext context, out object[] formatParameters)
+        {
+            var resources = paths.Keys
+                .SelectMany(path => resPathPattern.Matches(path)
+                    .OfType<Match>()
+                    .Select(match => match.Groups["resPath"].Value.ToString()))
+                .Distinct()
+                .ToList();
```
fearthecowboy: `.ToArray()` is a bit nicer than `.ToList()` 

`ToList()` creates a mutable container. If you don't need the mutability, go for `ToArray()`

olydis: whoops, sure thing!

PR #1773, opened 2017-02-01
regex valitation for dictionary values

olydis:


olydis:
@fearthecowboy empty diff 👍

PR #1783, opened 2017-02-02
[Linter] Adding rule to warn about boolean type properties

veronicagg:
Rule: Boolean types are not recommended under properties unless it is the only option. Booleans are not descriptive and make them hard to use. Instead use string enums with allowed set of values defined.


veronicagg:
@amarzavery , @dsgouda , @vishrutshah please take a look when you get a chance.

src/modeler/AutoRest.Swagger/Validation/BooleanPropertyNotRecommended.cs
```diff
@@ -0,0 +1,58 @@
...
+                        if (property.Value.Type.ToString().ToLower().Equals("boolean"))
+                        {
+                            booleanProperties.Add(key + "/" + property.Key);
+
+                        }
+                    }
+                }
+            }
+            formatParameters = new[] { string.Join(", ", booleanProperties.ToArray()) };
+            return (booleanProperties.Count == 0);
```
dsgouda: If I understand correctly you are returning a true (or valid) if there are no boolean properties, I believe the comments don't reflect it

src/modeler/AutoRest.Swagger/Model/ServiceDefinition.cs
```diff
@@ -94,7 +94,7 @@ public ServiceDefinition()
 
         /// <summary>
         /// Key is the object serviceTypeName and the value is swagger definition.
-        /// </summary>
```
dsgouda: nit: deleted line

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
```diff
@@ -73,6 +73,14 @@ public void AvoidMsdnReferencesValidation()
         }
 
         [Fact]
+        public void BooleanPropertiesValidation()
```
dsgouda: Per @amarzavery we should be adding "positive" tests too

src/modeler/AutoRest.Swagger/Validation/BooleanPropertyNotRecommended.cs
```diff
@@ -0,0 +1,58 @@
...
+        public override bool IsValid(Dictionary<string, Schema> definitions, RuleContext context, out object[] formatParameters)
+        {
+            formatParameters = null;
+            List<string> booleanProperties = new List<string>();
+            foreach (string key in definitions.Keys)
+            {
+                Schema definitionSchema = definitions.GetValueOrNull(key);
+                if (definitionSchema.Properties != null)
+                {
+                    foreach (var property in definitionSchema.Properties)
```
dsgouda: Can probably be simplified using Linq's .Select() construct, will leave it to your discretion

src/modeler/AutoRest.Swagger/Validation/BooleanPropertyNotRecommended.cs
veronicagg: Yup, thanks! I'll make another PR for the fix as this one got merged.

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
veronicagg: I wasn't sure which positive test to add in this case, since many of the other tests already don't have boolean properties included. For example, "clean-complex-spec.json" does not have boolean properties and therefore doesn't return this warning. If you think that's not enough, just let me know and I can add something else.

PR #1784, opened 2017-02-02
Cache version strings after first generation.

marstr:
Also, rewrite CodeNamerGo method to be more readable and have fewer
allocations.

### DIFF
--- a/src/generator/AutoRest.Go/Templates/VersionTemplate.cshtml
+++ b/src/generator/AutoRest.Go/Templates/VersionTemplate.cshtml
@@ -21,13 +21,16 @@ const (
     major = "@(Model.Version[0])"
     minor = "@(Model.Version[1])"
     patch = "@(Model.Version[2])"
-    // Always begin a "tag" with a dash (as per http://semver.org)
-    tag   = "-beta"
+    tag   = "@(Model.Version[3])"
 
     userAgentFormat = "Azure-SDK-For-Go/%s arm-%s/%s"
 )
 
-var userAgent string
+// cached results of UserAgent and Version to prevent repeated operations.
+var (
+    userAgent string
+    version string
+)
 
 @EmptyLine
 // UserAgent returns the UserAgent string to use when sending http.Requests.
@@ -39,8 +42,6 @@ func UserAgent() string {
 }
 
 @EmptyLine
-var version string
-
 // Version returns the semantic version (see http://semver.org) of the client.
 func Version() string {
     if version == "" {
--- a/src/generator/AutoRest.Go/Templates/VersionTemplate.cshtml
+++ b/src/generator/AutoRest.Go/Templates/VersionTemplate.cshtml
@@ -13,7 +13,9 @@ package @Model.Namespace
 
 @EmptyLine
 import (
+    "bytes"
     "fmt"
+    "strings"
 )
 @EmptyLine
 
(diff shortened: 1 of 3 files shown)


azurecla:
Hi __@marstr__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (marstr). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

src/generator/AutoRest.Go/Templates/VersionTemplate.cshtml
```diff
@@ -24,18 +24,32 @@ const (
     // Always begin a "tag" with a dash (as per http://semver.org)
     tag   = "-beta"
 
-    semVerFormat = "%s.%s.%s%s"
     userAgentFormat = "Azure-SDK-For-Go/%s arm-%s/%s"
 )
 
+var userAgent string
```
mcardosos: I think this looks more similar to other declarations, but still it is just a nit
``` go
var (
 userAgent string
 version string
)
```

src/generator/AutoRest.Go/Templates/VersionTemplate.cshtml
```diff
@@ -24,18 +24,32 @@ const (
     // Always begin a "tag" with a dash (as per http://semver.org)
     tag   = "-beta"
```
mcardosos: What about getting rid of the dash here and move the comment to line 49?

src/generator/AutoRest.Go/Templates/VersionTemplate.cshtml
marstr: Fair enough. Personally, what I really wanted was the VB.NET version of static where you can define fields inside of methods so that they aren't in global scope, but are persistent between calls to that method.

src/generator/AutoRest.Go/Templates/VersionTemplate.cshtml
marstr: Ah, I'll actually just go ahead and delete the comment all together. However you are right, I may as well delete the dash in the beginning here.

mcardosos:
Oops, the go tests are not passing in Travis.

mcardosos:
Add `bytes` and `strings` in the import section in the template

src/generator/AutoRest.Go/TransformerGo.cs
```diff
@@ -136,7 +136,7 @@ private void TransformModelTypes(CodeModelGo cmg)
                     name = $"{name}Type";
                 }
 
-                if (CodeNamerGo.Instance.UserDefinedNames.Contains(name))
+                if (CodeNamerGo.UserDefinedNames.Contains(name))
```
mcardosos: I think this has to do with @fearthecowboy refactor

marstr: Meh, it doesn't really matter I guess. Either way only a single copy should be generated. What are your thoughts, @jhendrixMSFT?

jhendrixMSFT: In this specific case you're right it doesn't really matter.  However, there was a push to eliminate static fields and methods as they cannot be virtual (although it's likely these will never need to be virtual).  It feels like unnecessary churn IMO unless there's a real motivation for it that I'm missing.

marstr: If there was a push to have thing that way, I'm happy to change it. Though, it should be noted that I was changing it because I specifically don't think virtual makes much sense. :p

olydis:
@azuresdkci Test this please

olydis:
@azuresdkci Retest this please

fearthecowboy:

marstr:
Thanks folks!

PR #1785, opened 2017-02-03
Configuration file

olydis:
10% configuration file support (very basic, not tested, not fully according to proposal)
90% decoupling and deduplication that naturally goes along with it (see LOC delta!), also introducing a more explicit pipeline of events, since no longer everything can have an effect on anything


olydis:
running full diff over night

PR #1788, opened 2017-02-04
Adding list operations validation rules

dsgouda:
Added two new validation rules for list operations
- If a get or post operation is xmspageable or returns an array of some kind, it should be named as _list
- If a get or post operation is xmspageable or returns an array of some kind and the path are of either of the following formats:
1. Subscriptions/{subid}/ResourceGroups/{rgname}/providers/ProviderType/{providerName}/ResourceName
2. Subscriptions/{subid}/ResourceGroups/{rgname}/
3. Subscriptions/{subid}
then the allowable patterns for the operation id are 
1. Resource_ListByParent
2. Resource_ListByResourceGroup
3. Resource_List/Resource_ListSubscriptionId

@amarzavery PTAL


src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
```diff
@@ -194,6 +194,24 @@ public void InvalidFormatValidation()
...
+        {
+            var messages = ValidateSwagger(Path.Combine("Swagger", "Validation", "list-operation-naming.json"));
+            messages.AssertOnlyValidationMessage(typeof(ListOperationNamingWarning));
+        }
+
+        [Fact]
+        public void ListByOperationsValidation()
+        {
+            var messages = ValidateSwagger(Path.Combine("Swagger", "Validation", "listby-operations.json"));
+            Assert.Equal(messages.Count(), 3);
```
olydis: Swap parameters. Expected number always goes first, leading to more readable error messages. Currently:
```
Assert.Equal() Failure
Expected: 4
Actual:   3
```

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace AutoRest.Swagger.Model.Utilities
+{
+    public static class ValidationUtilities
+    {
+        private static readonly string XmsPageable = "x-ms-pageable";
+
+        // determine if the
```
olydis: whoops

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -0,0 +1,32 @@
...
+    public static class ValidationUtilities
+    {
+        private static readonly string XmsPageable = "x-ms-pageable";
+
+        // determine if the 
+        public static bool IsXmsPageableOrArrayResponseOperation(Operation op, ServiceDefinition entity)
+        {
+            if (op.Extensions.GetValue<object>(XmsPageable) != null) return true;
+
+            if (!(op.Responses?.ContainsKey("200")??false)) return false;
```
olydis: simpify a bit, think that should be equivalent to `op.Responses?.ContainsKey("200") != true`

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -0,0 +1,32 @@
...
+
+            if (!((op.Responses["200"]?.Schema?.Reference?.Equals(string.Empty)) ?? true))
```
olydis: `op.Responses["200"]?.Schema?.Reference?.Equals(string.Empty) == false`?

src/modeler/AutoRest.Swagger/Model/ServiceDefinition.cs
```diff
@@ -79,6 +79,8 @@ public ServiceDefinition()
         /// Key is actual path and the value is serializationProperty of http operations and operation objects.
         /// </summary>
         [Rule(typeof(UniqueResourcePaths))]
+        [Rule(typeof(ListOperationNamingWarning))]
+        [Rule(typeof(ListByOperationsValidation))]
```
olydis: general question: shouldn't ALL of these also apply to `CustomPaths`??? Check with the decision makers

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -0,0 +1,32 @@
+﻿using AutoRest.Core.Utilities;
```
olydis: MS header

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
dsgouda: CSharp-y question here,  if we don't do ```(op.Responses?.ContainsKey("200")??false``` here, what would be the default boolean value of this expression in case Responses object is null

src/modeler/AutoRest.Swagger/Validation/ListOperationNamingWarning.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    public class ListOperationNamingWarning : TypedRule<Dictionary<string,Dictionary<string, Operation>>>
+    {
+        private readonly string XmsPageable = "x-ms-pageable";
+        private readonly Regex regex = new Regex(@".+_List(.*)$", RegexOptions.IgnoreCase);
+        public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string,Dictionary<string, Operation>> entity, RuleContext context)
+        {
+            // get all operation objects that are either of get or post type
+            var serviceDefinition = (ServiceDefinition)context.Root;
+            var listingOperations = entity.Values.SelectMany(opDict=>opDict.Where(pair => pair.Key.Equals("get") || pair.Key.Equals("post")));
```
olydis: are the dict keys known to be lower case?

src/modeler/AutoRest.Swagger/Validation/ListOperationNamingWarning.cs
```diff
@@ -0,0 +1,53 @@
...
+            foreach (var opPair in listingOperations)
+            {
+                // if the operation id is already of the type _list we can skip this check
+                if (regex.IsMatch(opPair.Value.OperationId))
+                { continue; }
+
+                if(AutoRest.Swagger.Model.Utilities.ValidationUtilities.IsXmsPageableOrArrayResponseOperation(opPair.Value, serviceDefinition))
```
olydis: seeing this I wonder why the `ValidationUtilities` are not under `AutoRest.Swagger.Model.Utilities.ValidationUtilities`

src/modeler/AutoRest.Swagger/Model/ServiceDefinition.cs
dsgouda: Good point!

src/modeler/AutoRest.Swagger/Validation/ListOperationNamingWarning.cs
dsgouda: I'd be surprised if they aren't, will add an ignorecase just to be sure

src/modeler/AutoRest.Swagger/Validation/ListByOperationsValidation.cs
```diff
@@ -0,0 +1,143 @@
...
+                var caps = match.Groups[1].Captures;
+                int index = caps.Count - 1;
+                while (index >= 0 && PathParameterRegex.IsMatch(caps[index].Value))
+                {
+                    index--;
+                }
+                // if due to some reason the parent we find is of the form "foo.bar", let's recommend the
+                // operation id be named as "_listbyfoobar"
+                if (index > -1)
+                { opList.Add(string.Format(ListByTemplate, methodGrp, caps[index].Value.Substring(1)).Replace(".", string.Empty).ToLower()); }
```
olydis: not sure about that brace style, somewhat introduces a third one into AutoRest!

dsgouda: Fair enough, I didn't want to open a new line for such a small operation 😄

src/modeler/AutoRest.Swagger/Validation/ListByOperationsValidation.cs
```diff
@@ -0,0 +1,143 @@
...
+            var opList = new List<string>();
+            var methodGrp = GetMethodGroup(path);
+            if (methodGrp == string.Empty)
+            { return opList; }
+
+            // construct _listbyparent
+            var match = ParentRegex.Match(path.Substring(0, path.IndexOf(methodGrp)-1));
+            if (match.Success)
+            {
+                var caps = match.Groups[1].Captures;
```
olydis: think all this can be simplified to
```c#
var cap = match.Groups[1].Captures.Reverse().FirstOrDefault(cap => PathParameterRegex.IsMatch(cap));
if (cap != null)
{
    opList.Add(string.Format(ListByTemplate, methodGrp, cap.Value.Substring(1)).Replace(".", string.Empty).ToLower());
}
```

src/modeler/AutoRest.Swagger/Validation/ListByOperationsValidation.cs
```diff
@@ -0,0 +1,143 @@
...
+using AutoRest.Core.Validation;
+using AutoRest.Swagger.Model;
+using System.Linq;
+
+namespace AutoRest.Swagger.Validation
+{
+    public class ListByOperationsValidation : TypedRule<Dictionary<string, Dictionary<string, Operation>>>
+    {
+        private readonly Regex ListByRegex = new Regex(@".+_listby.+$", RegexOptions.IgnoreCase);
+        private readonly Regex UrlRegex = new Regex(@"(\/[^\/]+)+$", RegexOptions.IgnoreCase);
```
olydis: use named regex groups like `(?<foo>bar)`! Then you can go with `match.Groups["foo"]` rather than `match.Groups[MAGIC_NUMBER]` (note that the magic number can change when the regex is changed so suddenly hell breaks loose tracking all the required changes down)

src/modeler/AutoRest.Swagger/Validation/ListByOperationsValidation.cs
dsgouda: This did cross my mind, but ```CaptureCollection``` doesn't seem to be of ```<IEnuermable>``` type which means we cannot weave the Linq magic, unless I have missed something obvious here

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -0,0 +1,52 @@
...
+        private static readonly string XmsPageable = "x-ms-pageable";
+
+        // determine if the operation returns an array type or is of xmspageable type
+        public static bool IsXmsPageableOrArrayResponseOperation(Operation op, ServiceDefinition entity)
+        {
+            // if xmspageable type, return true
+            if (op.Extensions.GetValue<object>(XmsPageable) != null) return true;
+
+            // if a success response is not defined, we have nothing to check, return false
+            if ((op.Responses?.ContainsKey("200") ?? false) !=true) return false;
```
olydis: equal to `op.Responses?.ContainsKey("200") != true`

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            // if we have a non-null response schema, and the schema is of type array, return true
+            if (((op.Responses["200"]?.Schema?.Reference?.Equals(string.Empty)) ?? true)!=true)
```
olydis: `op.Responses["200"]?.Schema?.Reference?.Equals(string.Empty) == false`

src/modeler/AutoRest.Swagger/Validation/ListByOperationsValidation.cs
```diff
@@ -0,0 +1,163 @@
...
+        private readonly string SubscriptionId = "SubscriptionId";
+
+        public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string,Dictionary<string, Operation>> entity, RuleContext context)
+        {
+            // get all operation objects that are either of get or post type
+            var serviceDefinition = (ServiceDefinition)context.Root;
+            
+            foreach (var pathObj in entity)
+            {
+                var listOpIds = pathObj.Value.Select(pair => {
```
olydis: looks like one could first filter and then select, something like
```
pathObj.Value
    // operation should be a get or post
    .Where(pair => pair.Key.ToLower().Equals("get") || pair.Key.ToLower().Equals("post"))
    // operation id should be of the form *_list(by)
    .Where(pair => ListByRegex.IsMatch(pair.Value.OperationId) || pair.Value.OperationId.ToLower().EndsWith("_list"))
    // operation should return an array type or is an xms pageable type
    .Where(pair => ValidationUtilities.IsXmsPageableOrArrayResponseOperation(pair.Value, serviceDefinition))
    .Select(pair => pair.Value.OperationId)
```

src/modeler/AutoRest.Swagger/Validation/ListByOperationsValidation.cs
```diff
@@ -0,0 +1,163 @@
...
+                // find if there are any operations that violate the rule
+                var errOpIds = listOpIds.Where(opId => !opNames.Contains(opId.ToLower()));
+                
+                // no violations found, skip
+                if (IsNullOrEmpty(errOpIds))
+                {
+                    continue;
+                }
+                // aggregate suggested op names in a single readable string for the formatter
+                var suggestedNames = opNames.Aggregate((curr, next)=> curr+", "+ next);
```
olydis: `string.Join(", ", opNames)`

src/modeler/AutoRest.Swagger/Validation/ListByOperationsValidation.cs
```diff
@@ -0,0 +1,163 @@
...
+                foreach (var errOpId in errOpIds)
+                {
+                    var stringParams = new List<string>() { errOpId };
```
olydis: uhm, not used

src/modeler/AutoRest.Swagger/Validation/ListByOperationsValidation.cs
```diff
@@ -0,0 +1,163 @@
...
+            opList.Add(string.Format(ListTemplate, methodGrp).ToLower());
+
+            // construct _listbyresourcegroup
+            if ((RgRegex.IsMatch(path)))
+            {
+                opList.Add(string.Format(ListByTemplate, methodGrp, ResourceGroup).ToLower());
+            }
+            else
+            {
+                return opList;
```
olydis: so `_listbyparent` cannot apply if `_listbyresourcegroup` doesn't?

src/modeler/AutoRest.Swagger/Validation/ListByOperationsValidation.cs
dsgouda: 👍

src/modeler/AutoRest.Swagger/Validation/ListByOperationsValidation.cs
dsgouda: 👍

src/modeler/AutoRest.Swagger/Validation/ListByOperationsValidation.cs
```diff
@@ -0,0 +1,163 @@
...
+            return opList;
+        }
+
+        private string GetMethodGroup(string path)
+        {
+            var match = UrlRegex.Match(path);
+            var caps = match.Groups["urltoken"].Captures.OfType<Capture>().Reverse();
+            var capture = caps.FirstOrDefault(cap => !PathParameterRegex.IsMatch(cap.Value));
+            
+            return (capture!=null) ? capture.Value.Substring(1) : string.Empty;
```
olydis: make this `capture?.Value.Substring(1)` and change the check `GetValidListOpNames` to check for `null`. Don't use an empty string as a "special" value

src/modeler/AutoRest.Swagger/Validation/ListByOperationsValidation.cs
dsgouda: Yes, it is a cascading pattern, if the url path does not follow the ```/resourcegroup/{resourcegroupname}``` pattern then it does not follow the pattern ```/resourcegroup/{resourcegroupname}/providers/{namespace}/{type}/{type_name}```

olydis: note that providers can exist without a resource group

olydis: think that's called "tenant-wide", see for example https://github.com/jselfms/azure-rest-api-specs/blob/214bba23bbd38522b9b7d32a89d8e9f2508eff8c/arm-billing/2017-02-27-preview/swagger/billing.json, but not sure the linter rule applies there anyways, so cascading might be the right thing

dsgouda: Yup, confirmed here that we should be cascading this rule

PR #1792, opened 2017-02-07
make sure {{EmptyLine}} is on own line

mcardosos:
fixes razor bug:
- sometimes, no "\r\n" is emitted with plain text lines (without apparent reason!) so the `{{EmptyLine}}` ends up together with something else
- subsequently, the entire line is emitted as nothing but a newline, swallowing the content that was next to `{{EmptyLine}}`


olydis:
adding `PR-Do-Not-Merge` as we stumbled upon an even bigger, more fundamental problem with Razor that needs a serious fix!

PR #1797, opened 2017-02-07
Adding rule to validate body parameters' top level properties

veronicagg:
Rule to validate that top level properties in body of "GET/PUT/PATCH" should be one of name, type, id, location, properties, tags, plan, sku, etag, managedBy, identity; no other top level properties are allowed.


veronicagg:
@amarzavery , @dsgouda, @vishrutshah, please take a look when you get a chance. Thanks!

src/modeler/AutoRest.Swagger/Validation/BodyTopLevelProperties.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// id, location, properties, tags, plan, sku, etag, managedBy, identity
+        /// </summary>
+        /// <param name="paths"></param>
+        /// <returns></returns>
+        public override bool IsValid( Dictionary<string, Operation> path, RuleContext context, out object[] formatParameters)
+        {
+            List<string> notAllowedProperties = new List<string>();
+            foreach (var operation in path.Keys)
+            {
+                if ((operation.ToLower().Equals("get") ||
```
dsgouda: A suggestion here, we can probably leverage an already implemented function here https://github.com/Azure/autorest/blob/master/src/modeler/AutoRest.Swagger/Validation/TrackedResourceValidation.cs#L159 (we may have to move it to some common utilities namespace)

src/modeler/AutoRest.Swagger/Validation/BodyTopLevelProperties.cs
```diff
@@ -0,0 +1,94 @@
...
+                    operation.ToLower().Equals("put") ||
+                    operation.ToLower().Equals("patch")) && path[operation].Parameters != null)
+                {
+                    foreach (var param in path[operation].Parameters)
+                    {
+                        if (param.In == ParameterLocation.Body)
+                        {
+                            if (param.Schema.Reference != null)
+                            {
+                                string defName = param.Schema.Reference.Split('/').Last();
```
dsgouda: nit: you can use https://github.com/Azure/autorest/blob/e2af872f328678a6d6bddb44fc15cb5efc5b7b0c/src/modeler/AutoRest.Swagger/Extensions.cs#L54 instead

veronicagg: @dsgouda thanks for the suggestion, I've updated the code to use this.

src/modeler/AutoRest.Swagger/Validation/BodyTopLevelProperties.cs
veronicagg: @dsgouda thanks for the suggestion. I've decided to leave it as is, as the method in TrackedResourceValidation is taking ServiceDefinition as input and gets the filtered set of operations from all paths in the model, and in this case I'm filtering methods for each path passed as parameter instead, so I didn't see as much benefit in refactoring.

src/modeler/AutoRest.Swagger/Validation/BodyTopLevelProperties.cs
```diff
@@ -38,11 +38,11 @@ public override bool IsValid( Dictionary<string, Operation> path, RuleContext co
                         {
                             if (param.Schema.Reference != null)
                             {
-                                string defName = param.Schema.Reference.Split('/').Last();
+                                string defName = Extensions.StripDefinitionPath(param.Schema.Reference);
                                 var definition = ((ServiceDefinition)context.Root).Definitions[defName];
                                 if (definition != null && definition.AllOf != null)
                                 {
-                                    if (definition.AllOf.Select(s => s.Reference).Where(reference => resourceRefRegEx.IsMatch(reference)) !=null)
+                                    if (definition.AllOf.Select(s => s.Reference).Where(reference => resourceRefRegEx.IsMatch(reference)) !=null && definition.Properties != null)
```
vishrutshah: Question:  would it be better to check for `definition.Properties != null` before making linq query to avoid unnecessary computation?

src/modeler/AutoRest.Swagger/Validation/BodyTopLevelProperties.cs
```diff
@@ -57,7 +57,7 @@ public override bool IsValid( Dictionary<string, Operation> path, RuleContext co
                             }
                             if (param.Schema.AllOf != null)
                             {
-                                if (param.Schema.AllOf.Select(s => s.Reference).Where(reference => resourceRefRegEx.IsMatch(reference)) != null)
+                                if (param.Schema.AllOf.Select(s => s.Reference).Where(reference => resourceRefRegEx.IsMatch(reference)) != null && param.Schema.Properties != null)
```
vishrutshah: same question as previous one

src/modeler/AutoRest.Swagger/Validation/BodyTopLevelProperties.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// <summary>
+        /// This rule passes if the body parameter contains top level properties only from the allowed set: name, type,
+        /// id, location, properties, tags, plan, sku, etag, managedBy, identity
+        /// </summary>
+        /// <param name="paths"></param>
+        /// <returns></returns>
+        public override bool IsValid( Dictionary<string, Operation> path, RuleContext context, out object[] formatParameters)
+        {
+            List<string> notAllowedProperties = new List<string>();
+            foreach (var operation in path.Keys)
```
vishrutshah: minor: I'd prefer `foreach (Operation operation in path.keys)` instead of using `var` in general

src/modeler/AutoRest.Swagger/Validation/BodyTopLevelProperties.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                if ((operation.ToLower().Equals("get") ||
+                    operation.ToLower().Equals("put") ||
+                    operation.ToLower().Equals("patch")) && path[operation].Parameters != null)
+                {
+                    foreach (var param in path[operation].Parameters)
+                    {
+                        if (param.In == ParameterLocation.Body)
+                        {
+                            if (param.Schema.Reference != null)
```
vishrutshah: should we be using null coal-lasing operator for safety? `param?.Schema?.Reference != null`

src/modeler/AutoRest.Swagger/Validation/BodyTopLevelProperties.cs
```diff
@@ -0,0 +1,94 @@
...
+                            {
+                                string defName = Extensions.StripDefinitionPath(param.Schema.Reference);
+                                var definition = ((ServiceDefinition)context.Root).Definitions[defName];
+                                if (definition != null && definition.AllOf != null)
+                                {
+                                    if (definition.AllOf.Select(s => s.Reference).Where(reference => resourceRefRegEx.IsMatch(reference)) !=null && definition.Properties != null)
```
vishrutshah: similar question on checking `definition.Properties != null` before as we have `&&` operator

src/modeler/AutoRest.Swagger/Validation/BodyTopLevelProperties.cs
```diff
@@ -0,0 +1,94 @@
...
+                {
+                    foreach (var param in path[operation].Parameters)
+                    {
+                        if (param.In == ParameterLocation.Body)
+                        {
+                            if (param.Schema.Reference != null)
+                            {
+                                string defName = Extensions.StripDefinitionPath(param.Schema.Reference);
+                                var definition = ((ServiceDefinition)context.Root).Definitions[defName];
+                                if (definition != null && definition.AllOf != null)
```
vishrutshah: minor: `definitions?.AllOf != null`

veronicagg:
@ravbhatnagar please take a look if you get a chance.

src/modeler/AutoRest.Swagger/Validation/BodyTopLevelProperties.cs
veronicagg: Yup, will do, thanks!

src/modeler/AutoRest.Swagger/Validation/BodyTopLevelProperties.cs
veronicagg: Sure, thanks!

veronicagg:
@vishrutshah Here's some output from the rule:
"WARNING: BodyTopLevelProperties - Top level properties should be one of name, type, id, location, properties, tags, plan, sku, etag, managedBy, identity. Extra properties found: "StorageAccountCreateParameters/kind, StorageAccounts_Create:extraprop1, StorageAccounts_Update:extraprop"."

PR #1801, opened 2017-02-08
Validation rules for providers path urls

dsgouda:
@olydis @amarzavery PTAL


olydis:
Have you tested this rule against the specs repo? Playing around with a few paths I thing that this rule might be too strict, i.e. complains on paths that (I think) are actually fine:
image
(note: the first two paths do not match your regex, while the last two do.)

src/modeler/AutoRest.Swagger/Validation/ProvidersPathValidation.cs
```diff
@@ -13,10 +13,10 @@
 
 namespace AutoRest.Swagger.Validation 
 {
-    public class ProviderPathsValidation : TypedRule<Dictionary<string, Dictionary<string, Operation>>>
+    public class ProvidersPathValidation : TypedRule<Dictionary<string, Dictionary<string, Operation>>>
     {
-        private readonly Regex FullRegex = new Regex(@"\/Subscriptions\/\{.+\}\/ResourceGroups\/\{.+\}\/providers\/[^\/]+(\/[^\/]+\/\{[^\}]+\})*(\/[\/^])?([\?]?[^\/]+)*$", RegexOptions.IgnoreCase);
-        private readonly Regex ProviderRegex = new Regex(@"Subscriptions\/\{.+\}\/ResourceGroups\/\{.+\}\/providers\/.+$", RegexOptions.IgnoreCase);
+        private Regex FullRegex = new Regex(@"\/Subscriptions\/\{.+\}\/ResourceGroups\/\{.+\}\/providers\/[^\/]+(\/[^\/]+\/\{[^\}]+\})*(\/[^\/]+)?([\?]?[^\/]+)*$", RegexOptions.IgnoreCase);
```
olydis: why did you remove the `readonly`?

olydis: regex should be equivalent to `\/Subscriptions\/\{.+\}\/ResourceGroups\/\{.+\}\/providers\/[^\/]+(\/[^\/]+\/\{[^\}]+\})*(\/[^\/]+)?$` (shorter)

dsgouda: I'm trying to catch the pattern that x-ms-paths follow (i.e. /path?foo={fooid})
I know it's a little overboard, but don't see the harm

olydis: thought you wanted a simpler regex earlier ;-) also note that your regex also matches `.../{foo}bar` so it does not really catch the pattern! Note also that your test case is failing on Jenkins

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
```diff
@@ -427,6 +435,17 @@ public void LongRunningResponseDefined()
...
             messages.AssertOnlyValidationMessage(typeof(LongRunningResponseValidation), 0);
         }
+
+        /// <summary>
+        /// Verifies that a providers path is of proper format
+        /// </summary>
+        [Fact]
+        public void ProvidersPathValidJson()
+        {
+            var messages = ValidateSwagger(Path.Combine("Resource", "Swagger", "Validation", "positive", "providers-path-valid.json"));
```
olydis: `Core.Utilities.Extensions.CodeBaseDirectory, ` missing

dsgouda: I understand I am being inconsistent here but not all paths in this file have ```Core.Utilities.Extensions.CodeBaseDirectory``` just wondering why that is so

PR #1803, opened 2017-02-09
[Go] Added x-ms-parametrized-host support

mcardosos:
Adding support for `x-ms-parametrized-host`. Needs https://github.com/Azure/go-autorest/pull/116 to work. It works in a similar way to how adding path parameters works.
Please take a look.

Example generating [datalake-store/filesysytem](https://github.com/Azure/azure-rest-api-specs/blob/master/arm-datalake-store/filesystem/2016-11-01/swagger/filesystem.json).
`client.go`
``` go
// Package filesystem implements the Azure ARM Filesystem service API version
// 2016-11-01.
//
// Creates an Azure Data Lake Store filesystem client.
package filesystem

import (
	"github.com/Azure/go-autorest/autorest"
)

const (
	// APIVersion is the version of the Filesystem
	APIVersion = "2016-11-01"
	// DefaultAdlsFileSystemDNSSuffix is the default value for adls file system dns suffix
	DefaultAdlsFileSystemDNSSuffix = "azuredatalakestore.net"
)

// ManagementClient is the base client for Filesystem.
type ManagementClient struct {
	autorest.Client
	APIVersion              string
	AdlsFileSystemDNSSuffix string
}

// New creates an instance of the ManagementClient client.
func New() ManagementClient {
	return NewWithoutDefaults(DefaultAdlsFileSystemDNSSuffix)
}

// NewWithoutDefaults lots of tantrums creates an instance of the
// ManagementClient client.
func NewWithoutDefaults(adlsFileSystemDNSSuffix string) ManagementClient {
	return ManagementClient{
		Client:                  autorest.NewClientWithUserAgent(UserAgent()),
		APIVersion:              APIVersion,
		AdlsFileSystemDNSSuffix: adlsFileSystemDNSSuffix,
	}
}
```
`someoperationgroup.go`
```go
// NewGroupClient creates an instance of the GroupClient client.
func NewGroupClient() GroupClient {
	return GroupClient{New()}
}

// Append appends to the specified file. NOTE: The target must not contain data
// added by ConcurrentAppend. ConcurrentAppend and Append cannot be used
// interchangeably; once a target file has been modified using either of these
// append options, the other append option cannot be used on the target file.
//
// accountName is the Azure Data Lake Store account to execute filesystem
// operations on. directFilePath is the Data Lake Store path (starting with
// '/') of the file to which to append. streamContents is the file contents to
// include when appending to the file. streamContents will be closed upon
// successful return. Callers should ensure closure when receiving an error.op
// is the constant value for the operation. appendParameter is the constant
// value for the operation. offset is the optional offset in the stream to
// begin the append operation. Default is to append at the end of the stream.
// syncFlag is optionally indicates what to do after completion of the append.
// DATA indicates more data is coming so no sync takes place, METADATA
// indicates a sync should be done to refresh metadata of the file only. CLOSE
// indicates that both the stream and metadata should be refreshed upon append
// completion.
func (client GroupClient) Append(accountName string, directFilePath string, streamContents io.ReadCloser, op string, appendParameter string, offset *int64, syncFlag SyncFlag) (result autorest.Response, err error) {
	req, err := client.AppendPreparer(accountName, directFilePath, streamContents, op, appendParameter, offset, syncFlag)
	if err != nil {
		return result, autorest.NewErrorWithError(err, "filesystem.GroupClient", "Append", nil, "Failure preparing request")
	}

	resp, err := client.AppendSender(req)
	if err != nil {
		result.Response = resp
		return result, autorest.NewErrorWithError(err, "filesystem.GroupClient", "Append", resp, "Failure sending request")
	}

	result, err = client.AppendResponder(resp)
	if err != nil {
		err = autorest.NewErrorWithError(err, "filesystem.GroupClient", "Append", resp, "Failure responding to request")
	}

	return
}

// AppendPreparer prepares the Append request.
func (client GroupClient) AppendPreparer(accountName string, directFilePath string, streamContents io.ReadCloser, op string, appendParameter string, offset *int64, syncFlag SyncFlag) (*http.Request, error) {
	urlParameters := map[string]interface{}{
		"accountName":             accountName,
		"adlsFileSystemDnsSuffix": client.AdlsFileSystemDNSSuffix,
	}

	pathParameters := map[string]interface{}{
		"directFilePath": autorest.Encode("path", directFilePath),
	}

	queryParameters := map[string]interface{}{
		"api-version": client.APIVersion,
		"append":      autorest.Encode("query", appendParameter),
		"op":          autorest.Encode("query", op),
	}
	if offset != nil {
		queryParameters["offset"] = autorest.Encode("query", *offset)
	}
	if len(string(syncFlag)) > 0 {
		queryParameters["syncFlag"] = autorest.Encode("query", syncFlag)
	}

	preparer := autorest.CreatePreparer(
		autorest.AsPost(),
		autorest.WithCustomBaseURL("https://{accountName}.{adlsFileSystemDnsSuffix}", urlParameters),
		autorest.WithPathParameters("/webhdfs/v1/{directFilePath}", pathParameters),
		autorest.WithFile(streamContents),
		autorest.WithQueryParameters(queryParameters))
	return preparer.Prepare(&http.Request{})
}

// AppendSender sends the Append request. The method will close the
// http.Response Body if it receives an error.
func (client GroupClient) AppendSender(req *http.Request) (*http.Response, error) {
	return autorest.SendWithSender(client, req)
}

// AppendResponder handles the response to the Append request. The method always
// closes the http.Response Body.
func (client GroupClient) AppendResponder(resp *http.Response) (result autorest.Response, err error) {
	err = autorest.Respond(
		resp,
		client.ByInspecting(),
		azure.WithErrorUnlessStatusCode(http.StatusOK),
		autorest.ByClosing())
	result.Response = resp
	return
}
```


mcardosos:
Ahhh, Travis is failing because of Java tests

src/generator/AutoRest.Go/Model/ParameterGo.cs
```diff
@@ -186,9 +186,30 @@ public static IEnumerable<ParameterGo> HeaderParameters(this IEnumerable<Paramet
...
         }
 
+         public static IEnumerable<ParameterGo> URLParameters(this IEnumerable<ParameterGo> parameters)
+        {
+            var urlParams = new List<ParameterGo>();
+            foreach (ParameterGo p in parameters.ByLocation(ParameterLocation.Path))
+            {
+                if (p.Method.CodeModel.BaseUrl.Contains(p.SerializedName))
+                {
+                    urlParams.Add(p);
```
marstr: [syntax sugar opportunity]
Instead of allocating `urlParams` you could rewrite this function to look like:

``` C#
public static IEnumerable<ParameterGo> URLParameters(this IEnumerable<ParameterGo> parameters)
{
    foreach (ParameterGo p in parameters.ByLocation(ParameterLocation.Path))
    {
        if (p.Method.CodeModel.BaseUrl.Contains(p.SerializedName))
        {
            yield return p;
        }
    }
}
```

Alternatively, you could embrace LINQ:

``` C#
public static IEnumerable<ParameterGo> URLParameters(this IEnumerable<ParameterGo> parameters)
{
    return parameters.ByLocation(ParameterLocation.Path).Where(p => p.Method.CodeModel.BaseUrl.Contains(p.SerializedName));
}
```

Needless to say, these suggestions aren't even nits, just showing off cool things about C# :)

src/generator/AutoRest.Go/Model/ParameterGo.cs
```diff
@@ -186,9 +186,30 @@ public static IEnumerable<ParameterGo> HeaderParameters(this IEnumerable<Paramet
...
-            return parameters.ByLocation(ParameterLocation.Path);
+            var pathParams = new List<ParameterGo>();
+            foreach (ParameterGo p in parameters.ByLocation(ParameterLocation.Path))
+            {
+                if (!p.Method.CodeModel.BaseUrl.Contains(p.SerializedName))
+                {
+                    pathParams.Add(p);
+                }
+            }
+            return pathParams;
```
marstr: The comment above would apply here as well.

src/generator/AutoRest.Go/Templates/ServiceClientTemplate.cshtml
```diff
@@ -28,49 +28,92 @@ import (
...
+    @WrapComment("// ", string.Format("NewWithoutDefaults creates an instance of the {0} client.", Model.BaseClient))
+    func NewWithoutDefaults(@(Model.AllGlobalParameters)) @(Model.BaseClient) {
+        return @(Model.BaseClient){
+            Client: autorest.NewClientWithUserAgent(UserAgent()),
+            APIVersion: APIVersion,
+            @foreach(var p in Model.Properties)
+            {
+                if (!p.SerializedName.Value.IsApiVersion())
+                {
+                    @:@(string.Format("{0}: {1},", p.Name, p.Name.Value.ToSentence()))
```
marstr: Without digging in futher, is there any chance that this would need to be a pointer?

mcardosos: Ño. I think there is no need to. Client properties are always required. AFAIK, I think they will always be strings.

marstr: I figured that was the case, just wanted to make sure you'd thought about it. :)

src/generator/AutoRest.Go/Model/ParameterGo.cs
mcardosos: Cool!

Issue #1710, opened 2017-01-10
Suport (non-)nullability of return types


olydis:
- review the current default behavior of C#'s  code gen (it generates primitive return types as nullable, other generators don't)
- allow `x-nullable` on responses

brjohnstmsft:
:+1:

mieliespoor:
Yes this is a must. We can't use this at the moment because of this issue - we would need to make too many code changes.

dsgouda:
Fixed here https://github.com/Azure/autorest/commit/bc2c58df08ee5ebabe254d99a36eb712f4a33df3

PR #1804, opened 2017-02-09
Fixing bug 1710 and adding unit tests

dsgouda:
@olydis PTAL
https://github.com/Azure/autorest/issues/1710


src/modeler/AutoRest.Swagger/OperationBuilder.cs
```diff
@@ -249,6 +249,10 @@ private List<Stack<IModelType>> BuildResponses(Method method, CompositeType head
                             methodName,
                             response.Key));
                     }
+                    foreach (var key in _operation.Responses[response.Key].Extensions.Keys)
```
olydis: I'd rather copy `_operation.Responses[response.Key].Extensions` to `method.Responses[response.Key.ToHttpStatusCode()].Extensions` directly (so yes, make the property settable, it's mutable anyways).

dsgouda:
@olydis the code should consider every response type and check if it has set x-nullable to false and only then remove the nullability property (this does not take into consideration the "default" response which is generally the error object) I have modified the test yaml to reflect these cases just to be sure

dsgouda:
@azuresdkci Test this please

src/generator/AutoRest.CSharp/Templates/Rest/Client/MethodTemplate.cshtml
```diff
@@ -353,12 +353,12 @@ else
         @:{
             if (Model.DefaultResponse.Body.IsPrimaryType(KnownPrimaryType.Stream))
             {
-            @:@(Model.DefaultResponse.Body.AsNullableType()) _errorBody = await _httpResponse.Content.ReadAsStreamAsync().ConfigureAwait(false);
+            @:@(Model.DefaultResponse.Body.AsNullableType(Model.IsXNullableReturnType)) _errorBody = await _httpResponse.Content.ReadAsStreamAsync().ConfigureAwait(false);
```
olydis: Careful: this seems to be the error body, which is orthogonal to the "return type" body! So this does not belong here.

src/generator/AutoRest.CSharp/Templates/Rest/Client/MethodTemplate.cshtml
```diff
@@ -353,12 +353,12 @@ else
...
             }
             else
             {
             @:_responseContent = await _httpResponse.Content.ReadAsStringAsync().ConfigureAwait(false);
-            @:@(Model.DefaultResponse.Body.AsNullableType()) _errorBody =  Microsoft.Rest.Serialization.SafeJsonConvert.DeserializeObject<@(Model.DefaultResponse.Body.AsNullableType())>(_responseContent, @(Model.GetDeserializationSettingsReference(Model.DefaultResponse.Body)));
+            @:@(Model.DefaultResponse.Body.AsNullableType()) _errorBody =  Microsoft.Rest.Serialization.SafeJsonConvert.DeserializeObject<@(Model.DefaultResponse.Body.AsNullableType(Model.IsXNullableReturnType))>(_responseContent, @(Model.GetDeserializationSettingsReference(Model.DefaultResponse.Body)));
```
olydis: same as above

src/generator/AutoRest.CSharp/Model/MethodCs.cs
```diff
@@ -261,11 +261,11 @@ public virtual string ReturnTypeString
             {
                 if (ReturnType.Body != null)
                 {
-                    return ReturnType.Body.AsNullableType(HttpMethod != HttpMethod.Head);
+                    return ReturnType.Body.AsNullableType(HttpMethod != HttpMethod.Head && IsXNullableReturnType);
                 }
                 if (ReturnType.Headers != null)
                 {
-                    return ReturnType.Headers.AsNullableType(HttpMethod != HttpMethod.Head);
+                    return ReturnType.Headers.AsNullableType(HttpMethod != HttpMethod.Head && IsXNullableReturnType);
```
olydis: nullability does not apply to headers type (that is always an object)

src/generator/AutoRest.CSharp/Model/MethodCs.cs
```diff
@@ -167,12 +167,12 @@ public virtual string TaskExtensionReturnTypeString
...
                 {
                     return string.Format(CultureInfo.InvariantCulture,
-                        "System.Threading.Tasks.Task<{0}>", ReturnType.Body.AsNullableType(HttpMethod != HttpMethod.Head));
+                        "System.Threading.Tasks.Task<{0}>", ReturnType.Body.AsNullableType(HttpMethod != HttpMethod.Head && IsXNullableReturnType));
                 }
                 else if (ReturnType.Headers != null)
                 {
                     return string.Format(CultureInfo.InvariantCulture,
-                        "System.Threading.Tasks.Task<{0}>", ReturnType.Headers.AsNullableType(HttpMethod != HttpMethod.Head));
+                        "System.Threading.Tasks.Task<{0}>", ReturnType.Headers.AsNullableType(HttpMethod != HttpMethod.Head && IsXNullableReturnType));
```
olydis: nullability does not apply to headers type (that is always an object)

src/generator/AutoRest.CSharp/Model/MethodCs.cs
```diff
@@ -141,15 +141,15 @@ public virtual string OperationResponseReturnTypeString
...
                 }
                 if (ReturnType.Body != null)
                 {
-                    return $"Microsoft.Rest.HttpOperationResponse<{ReturnType.Body.AsNullableType(HttpMethod != HttpMethod.Head)}>";
+                    return $"Microsoft.Rest.HttpOperationResponse<{ReturnType.Body.AsNullableType(HttpMethod != HttpMethod.Head && IsXNullableReturnType)}>";
                 }
                 if (ReturnType.Headers != null)
                 {
-                    return $"Microsoft.Rest.HttpOperationHeaderResponse<{ReturnType.Headers.AsNullableType(HttpMethod != HttpMethod.Head)}>";
+                    return $"Microsoft.Rest.HttpOperationHeaderResponse<{ReturnType.Headers.AsNullableType(HttpMethod != HttpMethod.Head && IsXNullableReturnType)}>";
```
olydis: nullability does not apply to headers type (that is always an object)

src/generator/AutoRest.CSharp/Templates/Rest/Client/MethodTemplate.cshtml
dsgouda: Yup, that's a good catch

src/generator/AutoRest.CSharp/Templates/Rest/Client/MethodTemplate.cshtml
olydis: @dsgouda still wrong

src/generator/AutoRest.CSharp/Templates/Rest/Client/MethodTemplate.cshtml
```diff
@@ -479,7 +479,7 @@ else
         }
         else
         {
-        @:_result.Body = Microsoft.Rest.Serialization.SafeJsonConvert.DeserializeObject<@(Model.DefaultResponse.Body.AsNullableType())>(_defaultResponseContent, @(Model.GetDeserializationSettingsReference(Model.DefaultResponse.Body)));
+        @:_result.Body = Microsoft.Rest.Serialization.SafeJsonConvert.DeserializeObject<@(Model.DefaultResponse.Body.AsNullableType(Model. IsXNullableReturnType))>(_defaultResponseContent, @(Model.GetDeserializationSettingsReference(Model.DefaultResponse.Body)));
```
olydis: also not affected by nullability

dsgouda: What if the default response is just a non-nullable primary type?

src/core/AutoRest.Core/Model/Method.cs
```diff
@@ -268,5 +268,8 @@ public bool StructurallyEquals(Method other)
         [JsonIgnore]
         public virtual HashSet<string> LocallyUsedNames { get; } = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
+        [JsonIgnore]
+        public virtual bool IsXNullableReturnType { get { return (Responses.Keys.Any(key => Responses[key].Extensions?.Get<bool>("x-nullable") ?? true)); } }
```
olydis: In https://github.com/Azure/autorest/pull/1804#issuecomment-278780237 it sounds like you check if *all* responses response types are marked as nullable, but this checks whether *one* response is. That's why I asked for more tests, we would almost have missed this. Please don't only add tests that test for the "good" behavior.

src/generator/AutoRest.CSharp/Templates/Rest/Client/MethodTemplate.cshtml
olydis: valid point, but does `Model.IsXNullableReturnType` reflect whether the default response is supposed to be nullable? thought it's about the succeeding responses.

dsgouda: On second thoughts, seems like you might be right, we do not consider ixnullable for default response types

PR #1814, opened 2017-02-10
fixed the tests, CI and razor line break issue

olydis:


src/generator/AutoRest.CSharp.Unit.Tests/SerializationTests.cs
```diff
@@ -10,7 +10,7 @@ public class SerializationTests
         [Fact]
         public void ParseSecurityDefinitionType()
         {
-            var filePath = @"Resource\SerializationTests\SerializationTests.json";
+            var filePath = Path.Combine("Resource", "SerializationTests", "SerializationTests.json");
```
fearthecowboy: 👍 Nice Catch.

olydis:
@fearthecowboy should be good to merge! :-) Note that worker 2 and 3 don't seem to be updated, yet.

PR #1843, opened 2017-02-16
Adding Validation rule that verifies if a tracked resource has corresponding "patch" operation

dsgouda:
Validation rule that verifies if a tracked resource has  corresponding "patch" operation
@olydis @amarzavery  PTAL

P.S.: Disregard the commit history, a lot of it is local commits I had to make to save my work


olydis:
@dsgouda is this PR doing what it's name suggests???

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -48,5 +69,20 @@ public static bool IsXmsPageableOrArrayResponseOperation(Operation op, ServiceDe
             return false;
         }
 
+        public static List<Operation> GetOperationsByRequestMethod(string id, ServiceDefinition serviceDefinition)
```
olydis: make return `IEnumerable`, body should also be a one-liner

src/modeler/AutoRest.Swagger/Validation/TrackedResourceValidation.cs
```diff
@@ -138,7 +138,7 @@ private bool HandleSchema(Schema schema, Dictionary<string, Schema> definitions)
             return true;
         }
 
-        private bool listByXCheck(List<Operation> getOperations, Regex regEx, string definitionKey, Dictionary<string, Schema> definitions)
+        private bool listByXCheck(IEnumerable<Operation> getOperations, Regex regEx, string definitionKey, Dictionary<string, Schema> definitions)
```
olydis: pascal case

dsgouda: I don't mind making these changes but then I would have to update the whole file (and its dependencies if any). Given that it would be beyond the scope of this PR is that a good idea? Just wondering....

olydis: don't know if I get you, how does changing the name of this *private* method cause changes "beyond the scope of this PR"?

dsgouda: There are other methods in the file with similar naming convention, it would make sense to update all of them too, hence my doubts

dsgouda:
@azuresdkci Test this please

PR #1853, opened 2017-02-17
typescript AutoRest host (WIP)

olydis:


src/next-gen/autorest/index.ts
```diff
@@ -2,7 +2,25 @@
...
 
-// this file should get 'required' by the boostrapper 
+// this file should get 'required' by the boostrapper
 
-console.log("AutoRest!");
+import { spawn, execFile, ChildProcess } from "child_process";
+import * as path from "path";
+import * as yargs from "yargs";
 
+function awaitable(child: ChildProcess): Promise<number> {
```
fearthecowboy: Nice.

Should we be collecting this type of thing into a common library?

olydis: 👍

PR #1864, opened 2017-02-21
Validating collection objects have "value" property

dsgouda:
Validation rule to ensure that every collection object has a "value" property that is the list of items it contains
Added positive and negative tests
@vishrutshah @veronicagg PTAL

### DIFF
--- a/src/core/AutoRest.Core/Properties/Resources.resx
+++ b/src/core/AutoRest.Core/Properties/Resources.resx
@@ -370,6 +370,6 @@ Modelers:
     <value>'PUT' operation must use method name 'Create'.</value>
   </data>
   <data name="CollectionObjectPropertiesNamingMessage" xml:space="preserve">
-    <value>Collection object {0} returned by list operation {1} has has no property named 'value'</value>
+    <value>Collection object {0} returned by list operation {1} has no property named 'value'</value>
   </data>
 </root>
--- a/src/core/AutoRest.Core/Properties/Resources.resx
+++ b/src/core/AutoRest.Core/Properties/Resources.resx
@@ -369,4 +369,7 @@ Modelers:
   <data name="PutOperationNameNotValid" xml:space="preserve">
     <value>'PUT' operation must use method name 'Create'.</value>
   </data>
+  <data name="CollectionObjectPropertiesNamingMessage" xml:space="preserve">
+    <value>Collection object {0} returned by list operation {1} has has no property named 'value'</value>
+  </data>
 </root>
--- a/src/core/AutoRest.Core/Properties/Resources.resx
+++ b/src/core/AutoRest.Core/Properties/Resources.resx
@@ -370,6 +370,6 @@ Modelers:
     <value>'PUT' operation must use method name 'Create'.</value>
   </data>
   <data name="CollectionObjectPropertiesNamingMessage" xml:space="preserve">
-    <value>Collection object {0} returned by list operation {1} has no property named 'value'</value>
+    <value>Collection object {0} returned by list operation {1} has no property named 'value'.</value>
   </data>
 </root>
--- a/src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/collection-objects-naming.json
+++ b/src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/collection-objects-naming.json
@@ -9,6 +9,12 @@
   "schemes": [
     "https"
   ],
+  "produces": [
+    "application/json"
+  ],
+  "consumes": [
+    "application/json"
+  ],
   "basePath": "/",
   "paths": {
     "/foo": {
@@ -26,16 +32,7 @@
               "$ref": "#/definitions/Models"
             }
           }
-        },
-        "produces": [
-          "application/json"
-        ],
-        "consumes": [
-          "application/json"
-        ],
-        "schemes": [
-          "https"
-        ]
+        }
       }
     },
     "/operations": {
--- a/src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/collection-objects-naming.json
+++ b/src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/collection-objects-naming.json
@@ -1,4 +1,4 @@
-{
+﻿{
   "swagger": "2.0",
   "info": {
     "title": "Operations schemes has non-http/https type",
@@ -13,7 +13,7 @@
   "paths": {
     "/foo": {
       "post": {
-        "operationId": "Noun_Verb",
+        "operationId": "Models_list",
         "summary": "Foo path",
         "description": "Foo operation",
         "responses": {
@@ -64,7 +64,7 @@
   "x-ms-paths": {
     "/foo?{id}": {
       "get": {
-        "operationId": "foo_getByID",
+        "operationId": "foo_listByID",
         "summary": "Foo get path",
         "description": "Foo get operation",
         "responses": {
@@ -117,7 +117,7 @@
     "Models": {
       "description": "the models to return",
       "properties": {
-        "value": {
+        "mymodels": {
           "type": "array",
           "items": {
             "$ref": "#/definitions/Model"
--- a/src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/positive/collection-objects-naming-valid.json
+++ b/src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/positive/collection-objects-naming-valid.json
@@ -9,6 +9,12 @@
   "schemes": [
     "https"
   ],
+  "produces": [
+    "application/json"
+  ],
+  "consumes": [
+    "application/json"
+  ],
   "basePath": "/",
   "paths": {
     "/foo": {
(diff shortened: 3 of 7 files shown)


src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/collection-objects-naming.json
```diff
@@ -0,0 +1,150 @@
...
+            "description": "Unexpected error"
+          },
+          "200": {
+            "description": "OK",
+            "schema": {
+              "$ref": "#/definitions/Models"
+            }
+          }
+        },
+        "produces": [
```
vishrutshah: minor: I'd push all the produces and consumes to top level as we have one SHOULD rule that'll throw warning otherwise :)

In general, I'd try to avoid all other kind of errors/warning except the one i am testing this for.

src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/positive/collection-objects-naming-valid.json
```diff
@@ -0,0 +1,150 @@
...
+            "description": "Unexpected error"
+          },
+          "200": {
+            "description": "OK",
+            "schema": {
+              "$ref": "#/definitions/Models"
+            }
+          }
+        },
+        "produces": [
```
vishrutshah: minor: similar to previous one for all consumes and produces pushing up.

src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
```diff
@@ -0,0 +1,65 @@
...
+            }
+        }
+
+        /// <summary>
+        /// The template message for this Rule. 
+        /// </summary>
+        /// <remarks>
+        /// This may contain placeholders '{0}' for parameterized messages.
+        /// </remarks>
+        public override string MessageTemplate => "Collection object {0} returned by list operation {1} has has no property named 'value'";
```
vishrutshah: it's be good to get this from the resources files. 
Example: https://github.com/Azure/autorest/pull/1797/files#diff-6e873d5c1e662d9aeba912007ae331f2R82

src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
```diff
@@ -0,0 +1,65 @@
...
+using AutoRest.Core.Validation;
+using AutoRest.Swagger.Model;
+using AutoRest.Swagger.Model.Utilities;
+using System.Linq;
+
+namespace AutoRest.Swagger.Validation
+{
+    public class CollectionObjectPropertiesNamingValidation : TypedRule<Dictionary<string, Dictionary<string, Operation>>>
+    {
+        private readonly Regex ListRegex = new Regex(@".+_List([^_]*)$", RegexOptions.IgnoreCase);
```
vishrutshah: please add Id and violation category for this rule

src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// </remarks>
+        public override string MessageTemplate => "Collection object {0} returned by list operation {1} has has no property named 'value'";
+
+        /// <summary>
+        /// The severity of this message (ie, debug/info/warning/error/fatal, etc)
+        /// </summary>
+        public override Category Severity => Category.Warning;
+    }
+}
+
```
vishrutshah: minor: extra lines in EOF

src/modeler/AutoRest.Swagger/Model/ServiceDefinition.cs
```diff
@@ -83,6 +83,7 @@ public ServiceDefinition()
         [Rule(typeof(UniqueResourcePaths))]
         [Rule(typeof(ListOperationNamingWarning))]
         [Rule(typeof(ListByOperationsValidation))]
+        [Rule(typeof(CollectionObjectPropertiesNamingValidation))]
```
vishrutshah: curious: why is this rule getting applied at two places? one here

src/modeler/AutoRest.Swagger/Model/ServiceDefinition.cs
```diff
@@ -96,6 +97,7 @@ public ServiceDefinition()
         /// </summary>
         [JsonProperty("x-ms-paths")]
         [Rule(typeof(ListOperationNamingWarning))]
+        [Rule(typeof(CollectionObjectPropertiesNamingValidation))]
```
vishrutshah: curious: why is this rule getting applied at two places? another here

src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
```diff
@@ -0,0 +1,65 @@
...
+namespace AutoRest.Swagger.Validation
+{
+    public class CollectionObjectPropertiesNamingValidation : TypedRule<Dictionary<string, Dictionary<string, Operation>>>
+    {
+        private readonly Regex ListRegex = new Regex(@".+_List([^_]*)$", RegexOptions.IgnoreCase);
+        public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string, Dictionary<string, Operation>> entity, RuleContext context)
+        {
+            // get all operation objects that are either of get or post type
+            var serviceDefinition = (ServiceDefinition)context.Root;
+            var listOperations = entity.Values.SelectMany(opDict => opDict.Where(pair => pair.Key.ToLower().Equals("get") || pair.Key.ToLower().Equals("post")));
```
vishrutshah: just curoius why did we choose `.ToLower().Equals()` over `.EqualsIgnoreCase()`

src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
```diff
@@ -0,0 +1,65 @@
...
+
+namespace AutoRest.Swagger.Validation
+{
+    public class CollectionObjectPropertiesNamingValidation : TypedRule<Dictionary<string, Dictionary<string, Operation>>>
+    {
+        private readonly Regex ListRegex = new Regex(@".+_List([^_]*)$", RegexOptions.IgnoreCase);
+        public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string, Dictionary<string, Operation>> entity, RuleContext context)
+        {
+            // get all operation objects that are either of get or post type
+            var serviceDefinition = (ServiceDefinition)context.Root;
```
vishrutshah: minor: in general I'd prefer concrete type over `var` because that helps in understanding the type :)

src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
```diff
@@ -0,0 +1,65 @@
...
+                
+                var collType = opPair.Value.Responses["200"]?.Schema?.Reference?.StripDefinitionPath();
+                // if no response type defined skip
+                if (collType == null)
+                {
+                    continue;
+                }
+
+                var collTypeDef = serviceDefinition.Definitions[collType];
+                // if collection object has 2 properties or less (xmspageable objects can have the nextlink prop)
```
vishrutshah: minor: I'd follow `x-ms-pageable` exact name of extension so that it becomes easier to search the usage even in comments :)

src/modeler/AutoRest.Swagger/Model/ServiceDefinition.cs
dsgouda: Since we have 2 "paths" sections, one for the regular paths and one for our "x-ms-paths" both of which could theoretically violate the rule (as indicated in the test jsons)

src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
dsgouda: string does not seem to have "EqualsIgnoreCase" method

---
In reply to: [102373085](https://github.com/Azure/autorest/pull/1864#discussion_r102373085) [](ancestors = 102373085)

vishrutshah: You are right. That is one of our extension method: https://github.com/Azure/autorest/blob/5cce029452e94eb0362bf0ea9b1bb43893aca33a/src/core/AutoRest.Core/Utilities/Extensions.cs#L467

src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -369,4 +369,7 @@ Modelers:
   <data name="PutOperationNameNotValid" xml:space="preserve">
     <value>'PUT' operation must use method name 'Create'.</value>
   </data>
+  <data name="CollectionObjectPropertiesNamingMessage" xml:space="preserve">
+    <value>Collection object {0} returned by list operation {1} has has no property named 'value'</value>
```
veronicagg: typo: double "has"

src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
```diff
@@ -0,0 +1,72 @@
...
+                // if no response type defined skip
+                if (collType == null)
+                {
+                    continue;
+                }
+
+                var collTypeDef = serviceDefinition.Definitions[collType];
+                // if collection object has 2 properties or less (x-ms-pageable objects can have the nextlink prop)
+                // and the object does not have a property named "value", show the warning
+                if ((collTypeDef.Properties?.Count <= 2) && collTypeDef.Properties.All(prop => !prop.Key.ToLower().Equals("value")))
```
veronicagg: should we also be checking that the type of "value" is an "array"?

src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -370,6 +370,6 @@ Modelers:
     <value>'PUT' operation must use method name 'Create'.</value>
   </data>
   <data name="CollectionObjectPropertiesNamingMessage" xml:space="preserve">
-    <value>Collection object {0} returned by list operation {1} has has no property named 'value'</value>
+    <value>Collection object {0} returned by list operation {1} has no property named 'value'</value>
```
vishrutshah: minor: `.` at end of the message.

src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// </summary>
+        /// <remarks>
+        /// This may contain placeholders '{0}' for parameterized messages.
+        /// </remarks>
+        public override string MessageTemplate => Resources.CollectionObjectPropertiesNamingMessage;
+        
+        /// <summary>
+        /// The severity of this message (ie, debug/info/warning/error/fatal, etc)
+        /// </summary>
+        public override Category Severity => Category.Fatal;
```
vishrutshah: `Critical` mentioned in the excel sheet are `Error` not `Fatal` as per my knowledge.

dsgouda: Are you sure about this, I talked to @ravbhatnagar about this and he indicated ```Critical=Fatal```

vishrutshah: What I think we discussed was whether you mark it as critical/Error, swagger having this validation fail cannot be merged in, hence we do not need 2 possible category.

ravbhatnagar: @vishrutshah @dsgouda - Where I think this will help is in the reporting when number of criticals vs number of errors will help differentiate the extent of violations. It also helps when we try and get existing RPs fix the violations. We can try put the case for just fixing the critical ones if there is push back.

src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
```diff
@@ -49,7 +49,7 @@ public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<
                 var collTypeDef = serviceDefinition.Definitions[collType];
                 // if collection object has 2 properties or less (x-ms-pageable objects can have the nextlink prop)
                 // and the object does not have a property named "value", show the warning
```
veronicagg: nit: comment update to indicate a property of type array named "value"

dsgouda:
@azuresdkci Test this please

src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
dsgouda: So if I understand this correctly, we "should" mark this as Fatal ?

PR #1865, opened 2017-02-22
Validating rule to ensure tracked resource has GET operation

dsgouda:
- added positive/negative tests
- renamed a few files and methods for consistency
@vishrutshah @veronicagg PTAL


src/modeler/AutoRest.Swagger/Validation/TrackedResourceGetOperationValidation.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Collections.Generic;
+using AutoRest.Swagger.Model;
+using System.Text.RegularExpressions;
+using System.Linq;
+
+namespace AutoRest.Swagger.Validation
+{
+    public class TrackedResourceGetOperationValidation : TypedRule<Dictionary<string, Schema>>
+    {
+        private readonly Regex resNames = new Regex(@"(RESOURCE|TRACKEDRESOURCE)$", RegexOptions.IgnoreCase);
```
vishrutshah: please add Rule Id and violation category.

src/modeler/AutoRest.Swagger/Validation/TrackedResourceGetOperationValidation.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// The template message for this Rule. 
+        /// </summary>
+        /// <remarks>
+        /// This may contain placeholders '{0}' for parameterized messages.
+        /// </remarks>
+        public override string MessageTemplate => "Tracked resource {0} must have a get operation.";
```
vishrutshah: use resources file to get this string that'd help us searching all the messages at one place.

src/modeler/AutoRest.Swagger/Validation/TrackedResourceGetOperationValidation.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// The severity of this message (ie, debug/info/warning/error/fatal, etc)
+        /// </summary>
+        public override Category Severity => Category.Warning;
```
vishrutshah: as we now want to correctly fail, feel free to make this as Error if it is error, otherwise it's cool

src/modeler/AutoRest.Swagger/Validation/TrackedResourceGetOperationValidation.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        // Verifies if a tracked resource has a corresponding get operation
+        public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string, Schema> definitions, RuleContext context)
```
vishrutshah: Why are you not using IsValid method or am i missing something ?

dsgouda: If I understand correctly, isValid would return a bool for the entire dictionary of definitions, GetValidationMessages will return an IEnumerable of all violations in "definitions"

---
In reply to: [102374098](https://github.com/Azure/autorest/pull/1865#discussion_r102374098) [](ancestors = 102374098)

src/modeler/AutoRest.Swagger/Validation/TrackedResourceGetOperationValidation.cs
dsgouda: That probably is a good idea, but I see we're using Resources file from AutoRest.Core all over AutoRest swagger modeler tests, not sure if that's the right thing to do since these two are rather unrelated. Having said that, I'm on the fence about it.

src/modeler/AutoRest.Swagger/Validation/TrackedResourceGetOperationValidation.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            IEnumerable<Operation> getOperations = ValidationUtilities.GetOperationsByRequestMethod("get", (ServiceDefinition)context.Root);
+            foreach (KeyValuePair<string, Schema> definition in definitions)
+            {
+                if (resNames.IsMatch(definition.Key) || ValidationUtilities.IsTrackedResource(definition.Value, definitions))
+                {
+                    if (!getOperations.Any(op => (op.Responses["200"].Schema?.Reference?.StripDefinitionPath()) == definition.Key))
```
veronicagg: do we know there'll be "200" response for all get operations?

src/modeler/AutoRest.Swagger/Validation/TrackedResourceGetOperationValidation.cs
veronicagg: @dsgouda how does the console output look for this rule? does it output one violation per issue by doing it this way?

src/modeler/AutoRest.Swagger/Validation/TrackedResourceGetOperationValidation.cs
dsgouda: I'm going with the assumption that successful get operations will return the expected response model. Is there a case when this is not true?

---
In reply to: [102594553](https://github.com/Azure/autorest/pull/1865#discussion_r102594553) [](ancestors = 102594553)

src/modeler/AutoRest.Swagger/Validation/TrackedResourceGetOperationValidation.cs
dsgouda: This is similar to a number of validation rules already implemented. Please refer [https://github.com/Azure/autorest/blob/22f6f44aa8ca6560b581f8748c5fed958a73aa83/src/modeler/AutoRest.Swagger/Validation/ModelTypeIncomplete.cs#L19](https://github.com/Azure/autorest/blob/22f6f44aa8ca6560b581f8748c5fed958a73aa83/src/modeler/AutoRest.Swagger/Validation/ModelTypeIncomplete.cs#L19) This does return a validation error message per violation.

---
In reply to: [102595319](https://github.com/Azure/autorest/pull/1865#discussion_r102595319) [](ancestors = 102595319)

dsgouda:
@veronicagg This is similar to a number of validation rules already implemented. Please refer https://github.com/Azure/autorest/blob/22f6f44aa8ca6560b581f8748c5fed958a73aa83/src/modeler/AutoRest.Swagger/Validation/ModelTypeIncomplete.cs#L19 This does return a validation error message per violation.

Issue #1869, opened 2017-02-22
[BUG] Extensions in path cause generator to fail


stankovski:
```"/array/empty": {
	"x-extension-under-path": true,
	"get": {
		"operationId": "array_getEmpty",
		"description": "Get empty array value []",
		"tags": [
			"Array Operations"
		],
```
Above swagger causes generator to fail.

olydis:
fixed as of #1870

PR #1870, opened 2017-02-22
Fix for "Extensions in path cause generator to fail #1869"

stankovski:
Fix for #1869

### DIFF
--- a/src/modeler/AutoRest.Swagger/JsonConverters/PathLevelParameterConverter.cs
+++ b/src/modeler/AutoRest.Swagger/JsonConverters/PathLevelParameterConverter.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using AutoRest.Swagger.Model;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using AutoRest.Core.Logging;
 
 namespace AutoRest.Swagger.JsonConverters
 {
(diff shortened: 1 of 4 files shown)


src/modeler/AutoRest.Swagger/JsonConverters/PathLevelParameterConverter.cs
```diff
@@ -64,8 +64,23 @@ public override bool CanConvert(System.Type objectType)
...
+            foreach (JProperty operation in jo.Children())
+            {
+                try
+                {
+                    result[operation.Name] = JsonConvert.DeserializeObject<Operation>(operation.Value.ToString(),
+                        GetSettings(serializer));
+                }
+                catch (JsonException)
+                {
+                    // Skip
```
olydis: so we will silently fail for any *actual* deserialization issues in an operation, swallowing the operation without a bleep? I'd be very restrictive here, not even work with `try catch` but rather check the `operation.Name` or something for `x-` and otherwise try to deserialize without mercy.

stankovski: I'll make a change to check for the name and skip deserialization for extensions. I'll also use logger to report the error if it happens.

PR #1875, opened 2017-02-24
Refining/Fixing exisitng validation rules

dsgouda:
- Improved messaging for tracked resource get/patch operation validations
- Get/patch operation validation is not applied to the top level resource
- Get/patch operation validation is not applied to tracked resources which are not response models
@vishrutshah @amarzavery PTAL

### DIFF
--- a/src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
+++ b/src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
@@ -86,7 +86,7 @@ namespace AutoRest.Swagger.Model.Utilities
                                         pathPair => pathPair.Value.Select(
                                             pathObj => pathObj.Value.Responses?.ContainsKey("200") == true ? pathObj.Value.Responses["200"]?.Schema?.Reference?.StripDefinitionPath() : string.Empty)));
 
-            respDefinitions = respDefinitions.Where(def => !string.IsNullOrEmpty(def)).Distinct();
+            respDefinitions = respDefinitions.Where(def => !string.IsNullOrWhiteSpace(def)).Distinct();
 
             return respDefinitions;
         }
--- a/src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
+++ b/src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
@@ -73,5 +73,23 @@ namespace AutoRest.Swagger.Model.Utilities
         {
             return serviceDefinition.Paths.Values.Select(pathObj => pathObj.Where(pair=> pair.Key.ToLower().Equals(id.ToLower()))).SelectMany(pathPair => pathPair.Select(opPair => opPair.Value));
         }
+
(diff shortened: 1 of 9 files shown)


src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -73,5 +73,23 @@ public static IEnumerable<Operation> GetOperationsByRequestMethod(string id, Ser
...
+            var respDefinitions = serviceDefinition.Paths.SelectMany(
+                                    pathPair => pathPair.Value.Select(
+                                        pathObj => pathObj.Value.Responses?.ContainsKey("200") == true ? pathObj.Value.Responses["200"]?.Schema?.Reference?.StripDefinitionPath() : string.Empty));
+
+            respDefinitions = respDefinitions.Concat(
+                                    serviceDefinition.CustomPaths.SelectMany(
+                                        pathPair => pathPair.Value.Select(
+                                            pathObj => pathObj.Value.Responses?.ContainsKey("200") == true ? pathObj.Value.Responses["200"]?.Schema?.Reference?.StripDefinitionPath() : string.Empty)));
+
+            respDefinitions = respDefinitions.Where(def => !string.IsNullOrEmpty(def)).Distinct();
```
vishrutshah: I'd use `IsNullOrWhiteSpace` instead to avoid white spaces as well.

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -73,5 +73,23 @@ public static IEnumerable<Operation> GetOperationsByRequestMethod(string id, Ser
         {
             return serviceDefinition.Paths.Values.Select(pathObj => pathObj.Where(pair=> pair.Key.ToLower().Equals(id.ToLower()))).SelectMany(pathPair => pathPair.Select(opPair => opPair.Value));
         }
+
+        public static IEnumerable<string> GetResponseModelDefinitions(ServiceDefinition serviceDefinition)
```
vishrutshah: Please add comments stating that it only gets definitions for 200 responses not other status codes.

src/modeler/AutoRest.Swagger/Validation/TrackedResourcePatchOperationValidation.cs
```diff
@@ -41,12 +41,14 @@ public class TrackedResourcePatchOperationValidation : TypedRule<Dictionary<stri
         // Verifies if a tracked resource has a corresponding patch operation
         public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string, Schema> definitions, RuleContext context)
         {
-            IEnumerable<Operation> patchOperations = ValidationUtilities.GetOperationsByRequestMethod("patch", (ServiceDefinition)context.Root);
+            var serviceDefinition = (ServiceDefinition)context.Root;
```
vishrutshah: minor: in general i'd go with concrete type instead of `var`.

PR #1878, opened 2017-02-24
Validating properties bag for top level ARM resources

dsgouda:
@veronicagg @amarzavery  PTAL
@ravbhatnagar FYI


dsgouda:
@azuresdkci Test this please

src/modeler/AutoRest.Swagger/Model/ServiceDefinition.cs
```diff
@@ -105,6 +105,7 @@ public ServiceDefinition()
         /// <summary>
         /// Key is the object serviceTypeName and the value is swagger definition.
         /// </summary>
+        [Rule(typeof(ArmResourcePropertiesBag))]
```
veronicagg: should the rule apply only to definitions or could this properties be defined inline?

src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -378,4 +378,7 @@ Modelers:
   <data name="TrackedResourcePatchOperationMissing" xml:space="preserve">
     <value>Tracked resource '{0}' must have patch operation that at least supports the update of tags.</value>
   </data>
+  <data name="ArmPropertiesBagValidationMessage" xml:space="preserve">
+    <value>Properties bag [{0}] should not be repeated for ARM top level properties</value>
```
veronicagg: Not sure about the clarity of this message. You may want to check with @ravbhatnagar on what he'd like to see.

src/modeler/AutoRest.Swagger/Model/ServiceDefinition.cs
dsgouda: We discourage inline definitions and there is a rule that flags such violations, so this should do for us.

src/core/AutoRest.Core/Properties/Resources.resx
dsgouda: Agreed, @ravbhatnagar the message above would look something like "Properties bag [a, b, c, d] should not be repeated for ARM top level properties" where a,b,c,d are the properties found to be violating the rule. Any suggestions here?

ravbhatnagar: "Top level property names should not be repeated inside the properties bag. Properties {} conflict with ARM top level properties. Please rename these."

PR #1883, opened 2017-02-25
[Linter] Adding rule - property names should be camelCase

veronicagg:
@dsgouda, @amarzavery @vishrutshah please take a look when you get a chance.
FYI @ravbhatnagar


src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -378,4 +378,7 @@ Modelers:
   <data name="TrackedResourcePatchOperationMissing" xml:space="preserve">
     <value>Tracked resource '{0}' must have patch operation that at least supports the update of tags.</value>
   </data>
+  <data name="PropertyNameShouldBeCamelCase" xml:space="preserve">
+    <value>Property names should follow camelCase style: {0}.</value>
```
vishrutshah: If the category is RPCViolation, should "should" be "must"?

dsgouda: nit: For better clarity, can we make the message look like "Property name {0} should follow camelCase style: {1}." where 0 is the violating property and 1 is the suggested correct format?

src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/positive/property-names-casing-valid.json
```diff
@@ -0,0 +1,121 @@
...
+        "summary": "Foo path",
+        "description": "Foo operation",
+        "parameters": [
+          {
+            "name": "parameters",
+            "in": "body",
+            "required": true,
+            "schema": {
+              "type": "object",
+              "properties": {
```
dsgouda: I wonder why this is not flagged as a violation, we should not put inline definitions like so

src/modeler/AutoRest.Swagger/Model/ServiceDefinition.cs
```diff
@@ -90,6 +90,7 @@ public ServiceDefinition()
         [CollectionRule(typeof(BodyTopLevelProperties))]
         [CollectionRule(typeof(HttpVerbValidation))]
         [CollectionRule(typeof(DeleteMustHaveEmptyRequestBody))]
+        [CollectionRule(typeof(PropertiesNamesCamelCase))]
```
dsgouda: Shouldn't we traverse all definitions and apply this rule over all model definitions in the doc

src/modeler/AutoRest.Swagger/Validation/PropertiesNamesCamelCase.cs
```diff
@@ -0,0 +1,88 @@
...
+        public override bool IsValid(Dictionary<string, Operation> path, RuleContext context, out object[] formatParameters)
+        {
+            List<string> badlyNamedProperties = new List<string>();
+            foreach (string operation in path.Keys)
+            {
+                if (path[operation]?.Parameters != null)
+                {
+                    foreach (SwaggerParameter param in path[operation].Parameters)
+                    {
+                        if (param.In == ParameterLocation.Body)
```
dsgouda: I wonder..... does this rule apply only to body parameters? What about query parameters or path parameters, they are passed as json properties over the wire too... We might want to verify this.

src/modeler/AutoRest.Swagger/Validation/PropertiesNamesCamelCase.cs
```diff
@@ -0,0 +1,88 @@
...
+                        if (param?.Schema?.Reference != null)
+                        {
+                            string defName = Extensions.StripDefinitionPath(param.Schema.Reference);
+                            var definition = ((ServiceDefinition)context.Root).Definitions[defName];
+                            if (definition?.Properties != null && definition.Properties != null)
+                            {
+                                var selectedProps = definition.Properties.Select(s => s.Key).Where(propertyName => !propNameRegEx.IsMatch(propertyName));
+                                if (selectedProps != null)
+                                {
+                                    foreach (string prop in selectedProps)
```
dsgouda: It'd be better if we use the GetValidationMessages signature instead of returning a bool here, this could also simplify the for loop here into a simple Linq query

src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/positive/property-names-casing-valid.json
veronicagg: Hmm, I didn't know this was the case. @amarzavery could you confirm I shouldn't be checking for inline properties? and just check definitions?

src/modeler/AutoRest.Swagger/Model/ServiceDefinition.cs
veronicagg: If the previous question is confirmed, and properties are only defined in "definitions", then sure, I can make the tool apply to definitions only.

src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/positive/property-names-casing-valid.json
amarzavery: @dsgouda was saying that the linter should have flagged for Anonymous properties. He was surprised at it not being flagged.

src/modeler/AutoRest.Swagger/Validation/PropertiesNamesCamelCase.cs
```diff
@@ -0,0 +1,88 @@
...
+                            {
+                                badlyNamedProperties.AddRange(param.Schema.Properties.Select(s => s.Key).Where(propertyName => !propNameRegEx.IsMatch(propertyName)));
+                            }
+                        }
+                        if (param?.Schema?.Reference != null)
+                        {
+                            string defName = Extensions.StripDefinitionPath(param.Schema.Reference);
+                            var definition = ((ServiceDefinition)context.Root).Definitions[defName];
+                            if (definition?.Properties != null && definition.Properties != null)
+                            {
```
veronicagg: thanks  :)!

src/core/AutoRest.Core/Properties/Resources.resx
veronicagg: sure, addressing both comments in next commit.

src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/positive/property-names-casing-valid.json
veronicagg: ok, thanks.

src/modeler/AutoRest.Swagger/Model/ServiceDefinition.cs
veronicagg: left it as is, so I can check the body parameter properties too, if they exist.

src/modeler/AutoRest.Swagger/Validation/PropertiesNamesCamelCase.cs
veronicagg: I've changed it to use GetValidationMessages signature, let me know if it addresses your comment. thanks!

src/modeler/AutoRest.Swagger/Validation/PropertiesNamesCamelCase.cs
```diff
@@ -28,25 +30,24 @@ public class PropertiesNamesCamelCase : TypedRule<Dictionary<string, Operation>>
...
-        /// <summary>
-        /// Validates whether property names are camelCase.
-        /// </summary>
-        /// <param name="path">Operation path to validate</param>
-        /// <returns>true if there are no propeties violating camelCase style, false otherwise.</returns>
-        public override bool IsValid(Dictionary<string, Operation> path, RuleContext context, out object[] formatParameters)
+        ///// <summary>
+        ///// Validates whether property names are camelCase.
+        ///// </summary>
+        public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string, Operation> path, RuleContext context)
```
dsgouda: So, when you apply this validation rule to the paths only, it will check model properties that are being referenced in appropriate paths; what about nested references? There can also be cases where service teams like to declare forward-looking models that are not being referenced yet, we should be flagging those properties. I think this rule should be applied to the definitions dictionary instead of paths

veronicagg: @dsgouda I agree that definitions that aren't used are not validated, but we're flagging not used definitions via our semantic validation as there shouldn't be unused/not referenced definitions included. So I would leave this rule as is.

veronicagg:
@dsgouda thanks for the feedback. I've made updates to validate definitions that were missing before. Please take a look when you get a chance. thanks!

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -91,6 +92,44 @@ public static IEnumerable<string> GetResponseModelDefinitions(ServiceDefinition
...
+                sb[0] = sb[0].ToString().ToLower()[0];
+            }
+            bool firstUpper = true;
+            for (int i = 1; i < name.Length; i++)
+            {
+                if (char.IsUpper(sb[i]) && firstUpper)
+                {
+                    firstUpper = false;
+                }
+                else if (char.IsUpper(sb[i]))
```
dsgouda: What happens if the name is of the form "MyBloatedPropertyName"

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -91,6 +92,44 @@ public static IEnumerable<string> GetResponseModelDefinitions(ServiceDefinition
...
+            Regex propNameRegEx = new Regex(@"^[a-z0-9]+([A-Z][a-z0-9]+)+|^[a-z0-9]+$|^[a-z0-9]+[A-Z]$");
+            return (propNameRegEx.IsMatch(name));
+        }
+
+        /// <summary>
+        /// Returns a camel case styled string based on the string passed as parameter.
+        /// </summary>
+        /// <param name="name">String to convert to camel case style</param>
+        /// <returns>A string that conforms with camel case style based on the string passed as parameter.</returns>
+        public static string toCamelCase(string name)
```
dsgouda: nit: function names should start with a capitalized letter

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
veronicagg: It'll return myBloatedpropertyname as a suggestion. I thought I'd return something that complies with the camel case style we're imposing, even though it may not be the ideal suggestion, as the user would anyway review it and update it to a name that makes sense to them.

dsgouda: This worries me, proper camel casing should do "myBloatedPropertyName" i.e., subsequent capitalized letters should be okay http://stackoverflow.com/questions/5543490/json-naming-convention

PR #1885, opened 2017-02-27
Adding basic test infrastructure

fearthecowboy:
adds ability to use gulp test/typescript to run typescript mocha unit tests.

### DIFF
--- a/src/local_modules/common.iced
+++ b/src/local_modules/common.iced
@@ -2,6 +2,7 @@ fs = require('fs')
 
 concurrency = 0 
 queue = []
+global.completed = []
 
 module.exports =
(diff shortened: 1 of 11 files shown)


src/local_modules/common.iced
```diff
@@ -53,7 +54,25 @@ module.exports =
       prev.dep.unshift name
     
     # add the new task.
-    gulp.task name, deps, fn
+    # gulp.task name, deps, fn
+
+    if fn.length # see if the task function has arguments (betcha never saw that before!)
```
olydis: once ran into having `f.length` instead of `f().length` with TSC not complaining, was mind-blown... and then forgot about it :-(

Issue #781, opened 2016-03-02
Support default value in C# models


stankovski:
Currently default value is only honored in method arguments. Need to add support to models.

emgerner-msft:
+1 for this feature.

fearthecowboy:
Howdy!

In our planning for driving towards a stable '1.0' release, I'm marking this issue as 'deferred' :zzz: and we're going to review it during the post-1.0 planning cycle. 

_It's not to say that we're not going to work on it, or that this isn't not important, but at the moment, we're picking and choosing the stuff we **must** do before 1.0._ :horse_racing:  :horse_racing:  :horse_racing: 

We'll make sure we pick this back up at that point. :tada:

fearthecowboy:
Hmmm. I think this may need to be included in 1.0.

brjohnstmsft:
If you add this ability, will it be opt-out?

The reason I ask is that for default values, it doesn't really make sense to set them on the client, since the REST API should provide them. The types themselves are nullable (because only optional properties can have default values), so leaving them out of the request payload saves bandwidth.

What I'd find more useful is if AutoRest appended something to the description of the property that describes what the default value is.

fearthecowboy:
I imagine it will have to be something you actively want turned on.

I do have to think about this one just a bit more before I actually reactivate this bug for 1.0. 

Adding it to the description in any event would be a good idea.

olydis:
@fearthecowboy Do we consider this fixed as of #2190?

olydis:
Will close due to https://github.com/Azure/autorest/pull/2190. I think this solution will do until generator wave 2.0.

PR #1891, opened 2017-03-01
Removing default empty constructors, added tests

dsgouda:
This is a partial fix for bug https://github.com/Azure/autorest/issues/781
- Removing default constructor and leaving it up to service teams to implement it which is possible since class declarations are partial
- This is to mitigate cases where default values need to be set on models themselves, there are inherent problems to doing this within the default empty constructor itself since these model default values can cause serious conflicts during PATCH operations
- Added a test case that ensures that we do not auto generate an empty default constructor for a given model

@fearthecowboy @olydis PTAL


dsgouda:
@azuresdkci Test this please

PR #1899, opened 2017-03-02
[Linter] Making small change to camel case conversion

veronicagg:
@dsgouda I think this change could help for the scenario you mentioned in your review comment for #1883 .
If you agree, let me know and we can merge this. I wouldn't change more on this method, since the use is to make only a suggestion for the user.


dsgouda:
LGTM pending CI build pass

PR #1905, opened 2017-03-03
TS pipeline & working JsonRpc plugin & ext. ref. resolving

olydis:


src/autorest-core/lib/approved-imports/jsonrpc.ts
```diff
@@ -0,0 +1,6 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export * from 'vscode-jsonrpc';
```
fearthecowboy: How is cancellation done over json-rpc? Do we have to put a method in for that?

src/autorest-core/lib/approved-imports/multi-promise.ts
```diff
@@ -0,0 +1,173 @@
...
+ *   - type system enforces order of events and single end of things
+ *   - minimum amount of "past events" stored due to single linked list strucutre:
+ *     - whoever holds a MultiPromise prevents garbage collection
+ *     - consumption of an event effectively decrements ref-count on associated node
+ *   - leverages error infrastrucure of Promise
+ * - easy to "marshal" from/to any other paradigm (events, streams, regular promises, ...)
+ */
+
+
+export type MultiPromiseItem<T> = { readonly current: T, readonly next: MultiPromise<T> };
```
fearthecowboy: Do we have tests for this stuff?

olydis: not yet, but it's also not (anymore) used in tons of places (since I couldn't decide on what AAAL is actually gonna expose/what you can subscribe to as a consumer). if this is gonna be the "channel" between autorest modules/plugins/users I'll definitely add a load of tests!

olydis: thinking of "customizable pipeline" I thought that modules (like the modeler, the generator, the validator, ...) have those guys as (named) input and output channels, connected merely by the config file (while currently this is hard coded into some function). But not sure if that's the way to go

src/autorest-core/lib/approved-imports/jsonrpc.ts
olydis: their implementation seems to have some sort of cancel message https://github.com/Microsoft/vscode-languageserver-node/blob/master/jsonrpc/src/main.ts#L48 but I'm not sure if that's automatically built into every channel or something... will dig

Issue #1906, opened 2017-03-03
False positive validations


fearthecowboy:
Given 

``` batch
autorest -input https://github.com/AzureAutomationTeam/azure-rest-api-specs/blob/ac72df3b75765089756673c3bff43a2e5cafc044/arm-automation/compositeAutomation.json -CodeGenerator none -Modeler CompositeSwagger
```

Generates the following false-positive errors (`subscriptionId` and `api-verison`  are both global parameters.): 

```

WARNING: Parameters "subscriptionId" and "api-version" are not allowed in the operations section
        Path: https://github.com/AzureAutomationTeam/azure-rest-api-specs/blob/ac72df3b75765089756673c3bff43a2e5cafc044/arm-automation/compositeAutomation.json#/paths/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Automation/automationAccounts/{automationAccountName}/patch/parameters[5]
WARNING: Parameters "subscriptionId" and "api-version" are not allowed in the operations section
        Path: https://github.com/AzureAutomationTeam/azure-rest-api-specs/blob/ac72df3b75765089756673c3bff43a2e5cafc044/arm-automation/compositeAutomation.json#/paths/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Automation/automationAccounts/{automationAccountName}/put/parameters[5]
WARNING: Parameters "subscriptionId" and "api-version" are not allowed in the operations section
        Path: https://github.com/AzureAutomationTeam/azure-rest-api-specs/blob/ac72df3b75765089756673c3bff43a2e5cafc044/arm-automation/compositeAutomation.json#/paths/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Automation/automationAccounts/{automationAccountName}/delete/parameters[4]
WARNING: Parameters "subscriptionId" and "api-version" are not allowed in the operations section
        Path: https://github.com/AzureAutomationTeam/azure-rest-api-specs/blob/ac72df3b75765089756673c3bff43a2e5cafc044/arm-automation/compositeAutomation.json#/paths/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Automation/automationAccounts/{automationAccountName}/get/parameters[4]
WARNING: Parameters "subscriptionId" and "api-version" are not allowed in the operations section
[...]

PR #1907, opened 2017-03-04
 	Fixing validation rule for subscription id and api-version

dsgouda:
If a swagger spec does not have any reference to subscriptionId or api-version in any of its paths skipping this rule
@amarzavery @fearthecowboy ran against the spec here https://github.com/AzureAutomationTeam/azure-rest-api-specs/blob/ac72df3b75765089756673c3bff43a2e5cafc044/arm-automation/compositeAutomation.json successfully
PTAL

Addresses issue https://github.com/Azure/autorest/issues/1906


dsgouda:
@azuresdkci Test this please

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -13,8 +13,14 @@ namespace AutoRest.Swagger.Model.Utilities
...
     public static class ValidationUtilities
     {
-        private static readonly string XmsPageable = "x-ms-pageable";
+        private const string XmsPageable = "x-ms-pageable";
         private static readonly Regex TrackedResRegEx = new Regex(@".+/Resource$", RegexOptions.IgnoreCase);
+        private const string AzureHostSignature = "management.azure.com";
+
+        // if x-ms-parameterized-host extension is used or if the host is anything of the form management.azure, assume this to be an ARMServiceDefinition
+        // TODO: refine this definition for azure data plain also check x-ms-parameterized-host in detail to ensure this is ARM
+        public static bool IsARMServiceDefinition(ServiceDefinition serviceDefinition) =>
```
olydis: what's the goal of this?

olydis: note that virtually all validation rules are meant for ARM, so this smells like a solo effort for something that should be handled in general? note that autorest no longer runs validation rules by default anyways

dsgouda: Agreed but we wanted to isolate this validation for ARM specs only, it shouldn't throw validation errors for something like PetStore
The effort could be useful in a couple of other validation rule implementations where we also want to enforce url path patterns but for ARM specs only.
I also feel in the long run this differentiation could help us create something like a base validator that can validate the swagger spec irrespective of whether it is ARM or not and a separate ARM-specific validator with all the complex details

olydis: long run yes, but indeed this is currently double effort (now, only who asks for validation will get validation, so no, PetStore will no longer get flagged) - also I think validation categories (ARM, non-ARM) should be triggerable from outside, not by looking at the Swagger!!!

olydis: what I really wanna say is: don't start anything like this before syncing with everyone - I suspect most rules are ARM specific (not also that we already classify the validation rules, so it'll be super easy to filter them *centralized* at some point, *not* from within the validation rules as you do here)

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
```diff
@@ -24,22 +24,19 @@ public class ServiceDefinitionParameters : TypedRule<Dictionary<string, SwaggerP
         public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string, SwaggerParameter> ParametersMap, RuleContext context)
         {
             var serviceDefinition = (ServiceDefinition)context.Root;
-            if (ValidationUtilities.IsARMServiceDefinition(serviceDefinition))
+            // Check if subscriptionId is used but not defined in global parameters
+            bool isSubscriptionIdReferenced = serviceDefinition.Paths.Keys.Any(key => key.ToLower().Contains("{" + SubscriptionId.ToLower() + "}"));
+            if (isSubscriptionIdReferenced && (ParametersMap?.Values.Any(parameter => parameter.Name.ToLower().Equals(SubscriptionId.ToLower()))) == false)
```
olydis: use `EqualsIgnoreCase` instead of own ignore case logic

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
```diff
@@ -5,23 +5,38 @@
...
+        {
+            var serviceDefinition = (ServiceDefinition)context.Root;
+            // Check if subscriptionId is used but not defined in global parameters
+            bool isSubscriptionIdReferenced = serviceDefinition.Paths.Keys.Any(key => key.ToLower().Contains("{" + SubscriptionId.ToLower() + "}"));
+            if (isSubscriptionIdReferenced && (ParametersMap?.Values.Any(parameter => parameter.Name.ToLower().Equals(SubscriptionId.ToLower()))) == false)
+            {
+                yield return new ValidationMessage(new FileObjectPath(context.File, context.Path), this, SubscriptionId);
+            }
+            // For ARM specs, api version is almost always required, call it out if it isn't defined in the global params
+            if (ParametersMap?.Values.Any(parameter => parameter.Name.ToLower().Equals(ApiVersion)) == false)
```
olydis: `== false` makes little sense here (and in the lambda above) - either change it to `!` or introduce `?.` stuff. I think the latter is what you want, looking at the failing CI ;-)

dsgouda: d'oh!

dsgouda: string type for Name property in SwaggerParameter object does not have EqualsIgnoreCase method

olydis: it's an extension method defined in `Extensions.cs`

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
```diff
@@ -5,23 +5,38 @@
...
+        {
+            var serviceDefinition = (ServiceDefinition)context.Root;
+            // Check if subscriptionId is used but not defined in global parameters
+            bool isSubscriptionIdReferenced = serviceDefinition.Paths.Keys.Any(key => key.ToLower().Contains("{" + SubscriptionId.ToLower() + "}"));
+            if (isSubscriptionIdReferenced && (ParametersMap?.Values.Any(parameter => parameter.Name?.ToLower().Equals(SubscriptionId) == true)) == false)
+            {
+                yield return new ValidationMessage(new FileObjectPath(context.File, context.Path), this, SubscriptionId);
+            }
+            // For ARM specs, api version is almost always required, call it out if it isn't defined in the global params
+            if (ParametersMap?.Values.Any(parameter => parameter.Name?.ToLower().Equals(ApiVersion) == true) == false)
```
veronicagg: should we look if api-version parameter being referenced by any of the operations too? I would expect we don't want an non-referenced parameter in the spec.

src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/positive/operations-valid-parameters.json
```diff
@@ -0,0 +1,86 @@
...
+  ],
+  "basePath": "/",
+  "produces": [
+    "application/json"
+  ],
+  "consumes": [
+    "application/json"
+  ],
+  "paths": {
+    "/foo": {
```
dsgouda: Ahh yes, I'll fix this!

src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/positive/service-def-valid-parameters.json
```diff
@@ -0,0 +1,86 @@
+﻿{
```
dsgouda: There are 2 rules in play here:
1. Operation parameters MUST not include the subscriptionId and ApiVersion parameters 
2. ServiceDefinition global parameters MUST have a definition for subscriptionId (as long as it is being referenced in any of the url params) and apiversion parameters

This json has a URL reference to subscriptionId and a corresponding definition in the global parameters section for both subscriptionId and apiversion

src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/positive/service-def-valid-parameters.json
```diff
@@ -0,0 +1,86 @@
...
+  "consumes": [
+    "application/json"
+  ],
+  "paths": {
+    "Subscription/{subscriptionId}/foo": {
+      "post": {
+        "operationId": "Noun_Verb",
+        "summary": "Foo path",
+        "description": "Foo operation",
+        "responses": {
```
dsgouda: Yes it should, I'll add it

src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/service-def-invalid-parameters.json
```diff
@@ -17,7 +17,7 @@
     "application/json"
   ],
   "paths": {
-    "/foo": {
+    "Subscription/{subscriptionId}/foo": {
```
dsgouda: That would be to test operational parameters, which is a different rule

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
```diff
@@ -5,23 +5,38 @@
...
         /// <param name="paths"></param>
         /// <returns></returns>
-        public override bool IsValid(Dictionary<string, SwaggerParameter> ParametersMap) =>        
-             true == (ParametersMap?.Values.Any(parameter => parameter.Name == SubscriptionId || parameter.Name == ApiVersion));
-        
+        public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string, SwaggerParameter> ParametersMap, RuleContext context)
+        {
+            var serviceDefinition = (ServiceDefinition)context.Root;
+            // Check if subscriptionId is used but not defined in global parameters
+            bool isSubscriptionIdReferenced = serviceDefinition.Paths.Keys.Any(key => key.ToLower().Contains("{" + SubscriptionId.ToLower() + "}"));
```
dsgouda: As previously stated that's a different rule where I check the operation parameters

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
```diff
@@ -5,23 +5,38 @@
...
+            if (isSubscriptionIdReferenced && (ParametersMap?.Values.Any(parameter => parameter.Name?.ToLower().Equals(SubscriptionId) == true)) == false)
```
dsgouda: if subscriptionid is referenced only then we throw this error, not sure what is wrong here

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
dsgouda: Same as above

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
veronicagg: Should the rule report a violation if the path for an operation does not have subscriptionID in it? I thought not all paths have subscriptionID on them (for example OperationsAPI I believe does not have it).

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
veronicagg: Since the comment on the condition says "For ARM specs, api version is almost always required" my thought was to check at least if the api-version parameter is used in any of the operations, if it's not used in any operations in the spec, I would think we don't want to report a violation.

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
dsgouda: The rule is not applied at the path but at the serviceDefinition level, so if any path has subscriptionId in it, it should be defined in the global parameters section (we ignore the ones that do not have subscriptionId in them)

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
dsgouda: Well that's the pain point currently, we don't have a definitive way of distinguishing between ARM and non-ARM specs, we should apply this for all specs for now and then dedicate some time categorizing these rules. Will update the comment to reflect this.

dsgouda:
@fearthecowboy @veronicagg PTAL

dsgouda:
@fearthecowboy can you approve this if it looks okay

Issue #1829, opened 2017-02-14
[Potential Bug] Parameter 'useDateTimeOffset' is not expected


vishrutshah:
travis-ci job: https://travis-ci.org/Azure/azure-rest-api-specs/jobs/200532009

Error File: https://travis-ci.org/Azure/azure-rest-api-specs/jobs/200532009#L2328
Error Info: https://travis-ci.org/Azure/azure-rest-api-specs/jobs/200532009#L2337

Doesn't looks like linter rule but needs an investigation.

veronicagg:
@olydis do you know what this error is pointing out?

olydis:
Looking at the code that generates it (`Settings.cs#488`), it's complete bogus. Feel free to remove that "validation"! :-)

olydis:
fixed as of #1909

PR #1909, opened 2017-03-04
skip val. by def., react. ARS tests, fix #1828 & #1829, legacy --help

### LINKED ISSUE #1828: [Potential Bug] Parameter 'internalConstructors' is not expected
travis-ci job: https://travis-ci.org/Azure/azure-rest-api-specs/jobs/200532009

Error File: https://travis-ci.org/Azure/azure-rest-api-specs/jobs/200532009#L2328
Error Info: https://travis-ci.org/Azure/azure-rest-api-specs/jobs/200532009#L2336

Doesn't looks like linter rule but needs an investigation.

olydis:
- skip validation unless `-CodeGenerator=None` (makes `-SkipValidation` unnecessary)
- fixed and reactivated ARS tests in gulp
- fixed #1828 and #1829 (nonsense validation of `CustomSettings`)
- make `--help` also print the legacy help (=> don't call "new" autorest core)

### DIFF
--- a/gulpfile.iced
+++ b/gulpfile.iced
@@ -32,7 +32,6 @@ Import
   tests:() ->
     source 'src/**/*[Tt]ests.csproj'
       .pipe except /AutoRest.Tests/ig #not used yet.
-      .pipe except /AutoRest.AzureResourceSchema.Tests/ig
       #.pipe except /AutoRest.Swagger.Tests/ig
     
   # assemblies that we sign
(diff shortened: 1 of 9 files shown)


olydis:
@amarzavery validation skipped by default now! `-SkipValidation` no longer exists since literally the only reference to that setting is now replaced by a check for `-CodeGenerator None`...

PR #1970, opened 2017-03-08
TS: plugins, composite Swagger, ext. ref. resolution

olydis:
before merge:
- no cheating with Amar's plugin


olydis:
got merged into https://github.com/Azure/autorest/pull/1978

PR #1974, opened 2017-03-09
random fix

olydis:
@fearthecowboy yeah not sure exactly how that fixed it but since the problem was in `@types/node` I removed dependencies to that - as I understand the node types are automatically activated in `tsc` through `"types" : ["node"]`... seems to work

### DIFF
--- a/src/autorest/tsconfig.json
+++ b/src/autorest/tsconfig.json
@@ -10,7 +10,6 @@
     "declaration": true,
     "stripInternal": true,
     "lib": [
-      "es5",
       "es6",
       "es2015.collection",
       "es2015.promise",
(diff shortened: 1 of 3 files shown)


src/autorest/tsconfig.json
```diff
@@ -10,7 +10,6 @@
     "declaration": true,
     "stripInternal": true,
     "lib": [
-      "es5",
```
olydis: that probably didn't fix anything, but is there any effect of this? shouldn't `es6` fully capture that? (not sure about the `es2015.` stuff)

fearthecowboy: Shouldn't hurt.

Issue #1979, opened 2017-03-10
SkuModelValidation too strict


olydis:
It complains that `Sku Model is not valid. A Sku model must have name property.` - however, there is a `name` property, and it has type `string` (I guess it fails since the type is `$ref`ed to instead of being inline).

Example where it fails:
https://github.com/pankajsn/azure-rest-api-specs/blob/09dc69af66af20ca869f4513600ba07dbf28ce8e/arm-logic/2016-06-01/swagger/logic.json#L3696

sarangan12:
Task completed

Issue #1939, opened 2017-03-07
SkuModelValidation: update rule


veronicagg:
ID: ?
Category: RPC Violation.
Severity: Warning

Here's what the rule does:
    /// Validates the SKU Model. A Sku model must have name property. 
    /// It can also have tier, size, family, capacity as optional properties.

@ravbhatnagar is tier optional or required?

sarangan12:
Task completed

PR #2000, opened 2017-03-13
Code fix to rule evaluating Sku Model

sarangan12:
Code fix for the rule evaluating the Sku Model. Related to the following issues:

1. https://github.com/Azure/autorest/issues/1939
2. https://github.com/Azure/autorest/issues/1979


src/modeler/AutoRest.Swagger/Validation/SkuModelValidation.cs
```diff
@@ -50,7 +60,7 @@ public override bool IsValid(Dictionary<string, Schema> definitions)
 
                     bool hasName = schema.Properties.Any(property =>
                         property.Key.EqualsIgnoreCase("name") &&
-                        property.Value.Type == Model.DataType.String);
+                        (property.Value.Type == Model.DataType.String || (property.Value.Type == null && evaluateIf(property.Value.Reference, definitions, Model.DataType.String))));
```
veronicagg: could the condition after the '||' call a method named something like "IsReferenceOfTypeString", if that's what's being validated here?

src/modeler/AutoRest.Swagger/Validation/SkuModelValidation.cs
```diff
@@ -62,5 +72,34 @@ public override bool IsValid(Dictionary<string, Schema> definitions)
...
+        /// <param name="dataType">Datatype value to evaluate</param>
+        /// <returns>true if the reference is of the provided data type. False otherwise.</returns>
+        private bool evaluateIf(string reference, Dictionary<string, Schema> definitions, Model.DataType dataType)
+        {
+            if (reference == null)
+            {
+                return false;
+            }
+
+            string definitionName = reference.Substring(reference.LastIndexOf('/') + 1).Trim();
```
veronicagg: should this use Extensions.StripDefinitionPath(reference) ?

src/modeler/AutoRest.Swagger/Validation/SkuModelValidation.cs
```diff
@@ -62,5 +72,34 @@ public override bool IsValid(Dictionary<string, Schema> definitions)
...
+            Schema schema = definitions.GetValueOrNull(definitionName);
+            if (schema == null)
+            {
+                return false;
+            }
+
+            if (schema.Type == dataType || (schema.Type == null && schema.Reference != null && evaluateIf(schema.Reference, definitions, dataType)))
```
veronicagg: should we allow another reference in this case? or should it the type be defined at this level? Since this is specific to sku model, I'm wondering if it'd make sense to have another reference.

src/modeler/AutoRest.Swagger/Validation/SkuModelValidation.cs
sarangan12: I have renamed the method to IsReferenceOfType. But not after the '||'. Consider a situation such as I actually have the type set to boolean and ref type set to string. In theory, it is possible.

src/modeler/AutoRest.Swagger/Validation/SkuModelValidation.cs
sarangan12: Thanks. I did not know of that method. Changed the code.

src/modeler/AutoRest.Swagger/Validation/SkuModelValidation.cs
sarangan12: I do not see why we should stop with first level of reference. In theory, you can define with 2 or more references and it would still be valid

sarangan12:
@veronicagg I have confirmed with Gaurav again. 'name' is the only required property.

Issue #1924, opened 2017-03-07
CollectionObjectPropertiesNamingValidation: update rule


veronicagg:
This rule should only apply for x-ms-pageable operations.

sarangan12:
Task completed

PR #2002, opened 2017-03-13
Changed CollectionObjectPropertiesNamingValidation rule to look for x-ms-pageable operations only

sarangan12:
Code fix for issue https://github.com/Azure/autorest/issues/1924

### DIFF
--- a/src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
+++ b/src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
@@ -35,14 +35,23 @@ namespace AutoRest.Swagger.Model.Utilities
             return false;
         }
 
-        // determine if the operation is xms pageable or returns an object of array type
-        public static bool IsXmsPageableOrArrayResponseOperation(Operation op, ServiceDefinition entity)
+        // determine if an operation is xms pageable operation
+        public static bool IsXmsPageableOperation(Operation op)
         {
             // if xmspageable type, return true
-            if (op.Extensions.GetValue<object>(XmsPageable) != null) return true;
+            if (op.Extensions.GetValue<object>(XmsPageable) != null)
+            {
+                return true;
+            }
 
+            return false;
+        }
+
+        // determine if an operation returns an object of array type
+        public static bool IsArrayResponseOperation(Operation op, ServiceDefinition entity)
+        {
             // if a success response is not defined, we have nothing to check, return false
-            if (op.Responses?.ContainsKey("200") !=true) return false;
+            if (op.Responses?.ContainsKey("200") != true) return false;
 
             // if we have a non-null response schema, and the schema is of type array, return true
             if (op.Responses["200"]?.Schema?.Reference?.Equals(string.Empty) == false)
@@ -61,10 +70,26 @@ namespace AutoRest.Swagger.Model.Utilities
                     return false;
                 }
 
-                if (entity.Definitions[modelLink.StripDefinitionPath()].Properties?.Values?.Any(type => type.Type == DataType.Array)??false)
+                if (entity.Definitions[modelLink.StripDefinitionPath()].Properties?.Values?.Any(type => type.Type == DataType.Array) ?? false)
                 {
                     return true;
                 }
+            }
+
+            return false;
+        }
+
+        // determine if the operation is xms pageable or returns an object of array type
+        public static bool IsXmsPageableOrArrayResponseOperation(Operation op, ServiceDefinition entity)
+        {
+            if (IsXmsPageableOperation(op))
+            {
+                return true;
+            }
+
+            if (IsArrayResponseOperation(op, entity))
+            {
+                return true;
             }
 
             return false;
--- a/src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
+++ b/src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
@@ -39,12 +39,7 @@ namespace AutoRest.Swagger.Model.Utilities
         public static bool IsXmsPageableOperation(Operation op)
         {
             // if xmspageable type, return true
-            if (op.Extensions.GetValue<object>(XmsPageable) != null)
-            {
-                return true;
-            }
-
-            return false;
+            return (op.Extensions.GetValue<object>(XmsPageable) != null);
         }
 
         // determine if an operation returns an object of array type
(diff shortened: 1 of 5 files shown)


src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
```diff
@@ -34,7 +35,7 @@ public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<
             foreach (var opPair in listOperations)
             {
                 // if the operation id is not of type _list* or does not return an array type, skip
-                if (!ListRegex.IsMatch(opPair.Value.OperationId) || !ValidationUtilities.IsXmsPageableOrArrayResponseOperation(opPair.Value, serviceDefinition))
+                if (!ListRegex.IsMatch(opPair.Value.OperationId) || !(opPair.Value.Extensions.GetValueOrNull(XmsPageable) != null))
```
veronicagg: since we already have a method defined in utilities around this, but the issue is that it does 2 things at the same time, could we split it, and call the one that we need here, this way "x-ms-pageable" string doesn't need to be defined here, and we try to centralize concept definitions.?

sarangan12: @veronicagg Comments addresses

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -35,14 +35,23 @@ public static bool IsTrackedResource(Schema schema, Dictionary<string, Schema> d
...
         }
 
-        // determine if the operation is xms pageable or returns an object of array type
-        public static bool IsXmsPageableOrArrayResponseOperation(Operation op, ServiceDefinition entity)
+        // determine if an operation is xms pageable operation
+        public static bool IsXmsPageableOperation(Operation op)
         {
             // if xmspageable type, return true
-            if (op.Extensions.GetValue<object>(XmsPageable) != null) return true;
+            if (op.Extensions.GetValue<object>(XmsPageable) != null)
```
veronicagg: should we just return the result of evaluating the expression?

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -70,6 +79,22 @@ public static bool IsXmsPageableOrArrayResponseOperation(Operation op, ServiceDe
             return false;
         }
 
+        // determine if the operation is xms pageable or returns an object of array type
+        public static bool IsXmsPageableOrArrayResponseOperation(Operation op, ServiceDefinition entity)
+        {
+            if (IsXmsPageableOperation(op))
```
veronicagg: should we return (IsXmsPageableOperation(op) || IsArrayResponseOperation(op, entity)) ?

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -35,14 +35,23 @@ public static bool IsTrackedResource(Schema schema, Dictionary<string, Schema> d
...
 
+        // determine if an operation returns an object of array type
+        public static bool IsArrayResponseOperation(Operation op, ServiceDefinition entity)
+        {
             // if a success response is not defined, we have nothing to check, return false
-            if (op.Responses?.ContainsKey("200") !=true) return false;
+            if (op.Responses?.ContainsKey("200") != true) return false;
 
             // if we have a non-null response schema, and the schema is of type array, return true
             if (op.Responses["200"]?.Schema?.Reference?.Equals(string.Empty) == false)
```
veronicagg: how about using "!" instead of ==false?

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -61,7 +70,7 @@ public static bool IsXmsPageableOrArrayResponseOperation(Operation op, ServiceDe
                     return false;
                 }
 
-                if (entity.Definitions[modelLink.StripDefinitionPath()].Properties?.Values?.Any(type => type.Type == DataType.Array)??false)
+                if (entity.Definitions[modelLink.StripDefinitionPath()].Properties?.Values?.Any(type => type.Type == DataType.Array) ?? false)
```
veronicagg: should we refactor code a bit and return the result of evaluating the conditions? return ((entity.Definitions[modelLink.StripDefinitionPath()].Properties?.Values?.Count ?? 2) < 2) && (entity.Definitions[modelLink.StripDefinitionPath()].AllOf == null) && ((entity.Definitions[modelLink.StripDefinitionPath()].Properties?.Values?.Any(type => type.Type == DataType.Array))

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -35,14 +35,23 @@ public static bool IsTrackedResource(Schema schema, Dictionary<string, Schema> d
...
+
+            return false;
+        }
 
+        // determine if an operation returns an object of array type
+        public static bool IsArrayResponseOperation(Operation op, ServiceDefinition entity)
+        {
             // if a success response is not defined, we have nothing to check, return false
-            if (op.Responses?.ContainsKey("200") !=true) return false;
+            if (op.Responses?.ContainsKey("200") != true) return false;
```
veronicagg: could we use "!" on the expression instead of "!= true"?

sarangan12: In this case because of the conversion from bool? to bool, I need to use a cast such as !(bool)(op.Responses?.ContainsKey("200")) which looks awkward that using != true. So, leaving it as such

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
sarangan12: same as above. leaving it as such

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
sarangan12: I believe you meant to type || between the conditions instead of &&.  I think merging these 3 long conditions will just make the code look awkward. The current split seems to be clean and readable. Leaving it as such

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs

Issue #1943, opened 2017-03-08
LongRunningResponseValidation: update message


veronicagg:
Message could be improved, by splitting, and providing an accurate error based on what's violated.

PR #2004, opened 2017-03-13
[Linter] Split LongRunningResponseValidation into respective method validation…

vishrutshah:
… for actionable messages

Fix for #1943


vishrutshah:
Example Output:

```
{
    "type": "Error",
    "code": "LongRunningResponseValidationPost",
    "message": "A 'POST' operation (Foo_Post) with x-ms-long-running-operation extension must have a valid terminal success status code 204.",
    "jsonref": "https://raw.githubusercontent.com/Azure/autorest/master/src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/long-running-invalid-response-post.json#/paths/~1foo/post/extensions/x-ms-long-running-operation",
    "json-path": "https://raw.githubusercontent.com/Azure/autorest/master/src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/long-running-invalid-response-post.json#/paths/foo/post/extensions/x-ms-long-running-operation",
    "id": "M2005",
    "validationCategory": "SDKViolation",
    "providerNamespace": null,
    "resourceType": null
  }
```

src/modeler/AutoRest.Swagger/Validation/Extensions/LongRunningExtensionRule.cs
```diff
@@ -100,7 +87,7 @@ private static bool IsPostResponseCodesValid(Dictionary<string, OperationRespons
         /// </summary>
         /// <param name="responses">Dictionary of responses for the operation.</param>
         /// <returns><c>true</c> if at least one valid terminal status code availabe, otherwise <c>false</c>.</returns>
-        private static bool IsDeleteResponseCodesValid(Dictionary<string, OperationResponse> responses)
+        protected static bool IsDeleteResponseCodesValid(Dictionary<string, OperationResponse> responses)
```
veronicagg: is the condition and code block in line 95 the same as in line 100? Should the method name be "AreDeleteResponseCodesValid" (it seems that there's an inconsistency in plural and singular)?

src/modeler/AutoRest.Swagger/Validation/Extensions/LongRunningResponseValidationPost.cs
```diff
@@ -0,0 +1,57 @@
...
+            return IsPostValid(context, out formatParameters);
+        }
+
+        /// <summary>
+        /// Validate that Put or Patch long running operations must have valid terminal status code 200 or 201.
+        /// </summary>
+        /// <param name="context">Rule context.</param>
+        /// <param name="formatParameters">list of parameters for message formatter.</param>
+        /// <returns><c>true</c> if at least one valid terminal status code availabe, otherwise <c>false</c>.</returns>
+        private static bool IsPostValid(RuleContext context, out object[] formatParameters)
```
veronicagg: it looks like this method is kind of repeated, except for the actual operation verb that's used for the check in line 41, not sure if there's another abstraction we could make, I thought I'd ask, but don't feel like you have to look into it too much.

src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -384,4 +384,13 @@ Modelers:
   <data name="ArmPropertiesBagValidationMessage" xml:space="preserve">
     <value>Top level property names should not be repeated inside the properties bag for ARM resource "{0}". Properties [{1}] conflict with ARM top level properties. Please rename these.</value>
   </data>
+  <data name="LongRunningResponseDeleteNotValid" xml:space="preserve">
+    <value>A 'DELETE' operation ({0}) with x-ms-long-running-operation extension must have a valid terminal success status code 200 or 204 or both.</value>
```
dsgouda: nit: For consistency's sake let's just say ```'DELETE' operation '{0}'....```

src/modeler/AutoRest.Swagger/Validation/Extensions/LongRunningExtensionRule.cs
```diff
@@ -12,52 +12,39 @@ public abstract class LongRunningExtensionRule : ExtensionRule
...
         /// </summary>
-        public abstract override Category Severity { get; }
+        public override Category Severity => Category.Error;
+
+        /// <summary>
+        /// Returns http verb for the given Rule context.
+        /// </summary>
+        /// <param name="context">Rule context.</param>
+        /// <returns>http verb as string</returns>
+        protected static string GetHttpVerb(RuleContext context)
```
dsgouda: Or simply say ```protected static string GetHttpVerb(RuleContext context) =>  context?.Parent?.Parent?.Key;```

src/modeler/AutoRest.Swagger/Validation/Extensions/LongRunningExtensionRule.cs
```diff
@@ -12,52 +12,39 @@ public abstract class LongRunningExtensionRule : ExtensionRule
...
+            return method ?? null;
+        }
 
         /// <summary>
         /// Validate that Put or Patch long running operations must have valid terminal status code 200 or 201.
         /// </summary>
         /// <param name="responses">Dictionary of responses for the operation.</param>
         /// <returns><c>true</c> if at least one valid terminal status code availabe, otherwise <c>false</c>.</returns>
-        private static bool IsPutPatchResponseCodesValid(Dictionary<string, OperationResponse> responses)
+        protected static bool IsPutPatchResponseCodesValid(Dictionary<string, OperationResponse> responses)
```
dsgouda: While we are at it let's get this straight
The whole function could be replaced by:
```private static bool IsPutPatchResponseCodesValid(Dictionary<string, OperationResponse> responses) => responses?.GetValueOrNull("200")!=null && responses?.GetValueOrNull("201")!=null;```
The method is unnecessarily verbose and quite inexplicably has the same if condition repeated 😮

File: src/modeler/AutoRest.Swagger/Validation/Extensions/LongRunningExtensionRule.cs [hunk 6]
```diff
@@ -80,7 +67,7 @@ private static bool IsPutPatchResponseCodesValid(Dictionary<string, OperationRes
         /// </summary>
         /// <param name="responses">Dictionary of responses for the operation.</param>
         /// <returns><c>true</c> if at least one valid terminal status code availabe, otherwise <c>false</c>.</returns>
-        private static bool IsPostResponseCodesValid(Dictionary<string, OperationResponse> responses)
+        protected static bool IsPostResponseCodesValid(Dictionary<string, OperationResponse> responses)
```
dsgouda: Again, condense this to ```protected static bool IsPostResponseCodesValid(Dictionary<string, OperationResponse> responses)=>!IsPutPatchResponseCodesValid(responses) && responses?.GetValueOrNull("204")!=null;```

src/modeler/AutoRest.Swagger/Validation/Extensions/LongRunningExtensionRule.cs
dsgouda: Exact same comment as for IsPutPatchResponseCodesValid method

src/modeler/AutoRest.Swagger/Validation/Extensions/LongRunningResponseValidationDelete.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        /// <summary>
+        /// An x-ms-long-running-operation extension passes this rule if the operation that this extension has a valid response defined.
+        /// </summary>
+        /// <param name="longRunning">long running extension object.</param>
+        /// <param name="context">Rule context.</param>
+        /// <param name="formatParameters">list of parameters for message formatter.</param>
+        /// <returns><c>true</c> if operation has valid response code modeled, otherwise <c>false</c>.</returns>
+        /// <remarks>This rule corresponds to M2005.</remarks>
+        public override bool IsValid(object longRunning, RuleContext context, out object[] formatParameters)
```
dsgouda: If I understand correctly you're applying this rule at the ```Operation``` level, why not apply it at the ```Paths``` level and use Linq to identify operations that violate this rule, that would be much cleaner IMO

vishrutshah: I am applying this rule at extension level. I agree with you on Linq query but specifically for this case, I have concerns in moving them to the paths level:

If i apply it to the path level then I think we'll introduce the overhead of finding out all operations that has `x-ms-long-running` extension i.e. Linq has to search through all the operation to find subset. Secondly, as you are already working for some optimization around performance to speedup the validation process this would be like just opposite in the sense to incuring more cost.  

Instead I'd prefer to keep it as deep as possible whenever we can. What do you think about this?

src/modeler/AutoRest.Swagger/Validation/Extensions/LongRunningResponseValidationDelete.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// <summary>
+        /// Validate that Put or Patch long running operations must have valid terminal status code 200 or 201.
+        /// </summary>
+        /// <param name="context">Rule context.</param>
+        /// <param name="formatParameters">list of parameters for message formatter.</param>
+        /// <returns><c>true</c> if at least one valid terminal status code availabe, otherwise <c>false</c>.</returns>
+        private static bool IsDeleteValid(RuleContext context, out object[] formatParameters)
+        {
+            string httpVerb = GetHttpVerb(context);
+            bool isPostOperation = httpVerb != null &&
```
vishrutshah: fixed

src/modeler/AutoRest.Swagger/Validation/Extensions/LongRunningExtensionRule.cs
vishrutshah: fixed

src/modeler/AutoRest.Swagger/Validation/Extensions/LongRunningExtensionRule.cs
vishrutshah: fixed

src/modeler/AutoRest.Swagger/Validation/Extensions/LongRunningExtensionRule.cs
vishrutshah: Fixed

src/modeler/AutoRest.Swagger/Validation/Extensions/LongRunningResponseValidationPost.cs
vishrutshah: Yes I can but personally feels like not spending too much time on this because that'd lead me into Func delegator into C# and seems like overkill. Let me know if you guys think otherwise.

dsgouda: Repeated code is never a good idea, unless I'm mistaken here we can simply parameterize the operation type and response codes to check in a common function (somewhere in ValidationUtilities) or perhaps create an abstract base class with this common method, no need for delegates and such.
Some effort now will save us a lot of time while debugging/implementing better logic later on

src/modeler/AutoRest.Swagger/Validation/Extensions/LongRunningResponseValidationPutPatch.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        /// <summary>
+        /// Validate that Put or Patch long running operations must have valid terminal status code 200 or 201.
+        /// </summary>
+        /// <param name="context">Rule context.</param>
+        /// <param name="formatParameters">list of parameters for message formatter.</param>
+        /// <returns><c>true</c> if at least one valid terminal status code availabe, otherwise <c>false</c>.</returns>
+        private static bool IsPutPatchValid(RuleContext context, out object[] formatParameters)
+        {
+            string httpVerb = GetHttpVerb(context);
```
dsgouda: Or we could say 
``` 
string httpVerb = GetHttpVerb(context)?.toLower();
if(httpVerb=="put" || httpVerb =="patch")
```

src/modeler/AutoRest.Swagger/Validation/Extensions/LongRunningResponseValidationDelete.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        ///// <summary>
+        ///// An x-ms-long-running-operation extension passes this rule if the operation that this extension has a valid response defined.
+        ///// </summary>
+        ///// <param name="longRunning">long running extension object.</param>
+        ///// <param name="context">Rule context.</param>
+        ///// <param name="formatParameters">list of parameters for message formatter.</param>
+        ///// <returns><c>true</c> if operation has valid response code modeled, otherwise <c>false</c>.</returns>
+        ///// <remarks>This rule corresponds to M2005.</remarks>
+        //public override bool IsValid(object longRunning, RuleContext context, out object[] formatParameters) => AreResponseCodesValid(context, out formatParameters);
```
veronicagg: commented out?

src/modeler/AutoRest.Swagger/Validation/Extensions/LongRunningResponseValidationPutPatch.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        ///// <summary>
+        ///// An x-ms-long-running-operation extension passes this rule if the operation that this extension has a valid response defined.
+        ///// </summary>
+        ///// <param name="longRunning">long running extension object.</param>
+        ///// <param name="context">Rule context.</param>
+        ///// <param name="formatParameters">list of parameters for message formatter.</param>
+        ///// <returns><c>true</c> if operation has valid response code modeled, otherwise <c>false</c>.</returns>
+        ///// <remarks>This rule corresponds to M2005.</remarks>
+        //public override bool IsValid(object longRunning, RuleContext context, out object[] formatParameters) => AreResponseCodesValid(context, out formatParameters);
```
veronicagg: commented out?

src/modeler/AutoRest.Swagger/Validation/Extensions/LongRunningResponseValidationDelete.cs
```diff
@@ -0,0 +1,44 @@
...
+        ///// <remarks>This rule corresponds to M2005.</remarks>
+        public override bool IsValid(object longRunning, RuleContext context, out object[] formatParameters)
+        {
+            if (!"delete".EqualsIgnoreCase(GetHttpVerb(context)))
+            {
+                formatParameters = new string[0];
+                return true;
+            }
+
+            bool IsValid = AreValidResponseCodes(context, allowedCodes);
```
vishrutshah: move it to line `41`

src/modeler/AutoRest.Swagger/Validation/Extensions/LongRunningResponseValidationDelete.cs
```diff
@@ -0,0 +1,44 @@
...
+        ///// An x-ms-long-running-operation extension passes this rule if the operation that this extension has a valid response defined.
+        ///// </summary>
+        ///// <param name="longRunning">long running extension object.</param>
+        ///// <param name="context">Rule context.</param>
+        ///// <param name="formatParameters">list of parameters for message formatter.</param>
+        ///// <returns><c>true</c> if operation has valid response code modeled, otherwise <c>false</c>.</returns>
+        ///// <remarks>This rule corresponds to M2005.</remarks>
+        public override bool IsValid(object longRunning, RuleContext context, out object[] formatParameters)
+        {
+            if (!"delete".EqualsIgnoreCase(GetHttpVerb(context)))
```
dsgouda: declare delete as a const static string instead

vishrutshah:
@azuresdkci Test this please

dsgouda:
🚢 it

PR #2009, opened 2017-03-14
enable service teams to have a composite swagger with 1 spec, when the next one is going to come soon

amarzavery:


amarzavery:
@veronicagg @fearthecowboy - Please take a look.

Issue #1948, opened 2017-03-08
XmsPathsMustOverloadPaths: update message


veronicagg:
Improve message to indicate path what's violating the rule
ID: ? 
Category: SDK Violation
Severity: Error

sarangan12:
Task Completed

PR #2011, opened 2017-03-14
Code changes for Issue 1948

sarangan12:
Code changes for Issue 1948

https://github.com/Azure/autorest/issues/1948 

### DIFF
--- a/src/core/AutoRest.Core/Properties/Resources.resx
+++ b/src/core/AutoRest.Core/Properties/Resources.resx
@@ -307,7 +307,7 @@ Modelers:
     <value>Errors found during Swagger document validation.</value>
   </data>
   <data name="XMSPathBaseNotInPaths" xml:space="preserve">
-    <value>Paths in x-ms-paths must overload a normal path in the paths section, i.e. a path in the x-ms-path must either be same as a path in the paths section or a path in the paths sections followed by additional parameters.</value>
+    <value>Paths in x-ms-paths must overload a normal path in the paths section, i.e. a path in the x-ms-paths must either be same as a path in the paths section or a path in the paths sections followed by additional parameters.</value>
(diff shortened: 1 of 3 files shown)


src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -307,7 +307,7 @@ Modelers:
     <value>Errors found during Swagger document validation.</value>
   </data>
   <data name="XMSPathBaseNotInPaths" xml:space="preserve">
-    <value>Paths in x-ms-paths must overload a normal path in the paths section.</value>
+    <value>Paths in x-ms-paths must overload a normal path in the paths section, i.e. a path in the x-ms-path must either be same as a path in the paths section or a path in the paths sections followed by additional parameters.</value>
```
veronicagg: typo: "x-ms-paths", s is missing in second instance.

veronicagg: are we providing the user which path is violating the rule? I believe that's why we wanted to update the message, could you paste an example output of something violating this rule?

sarangan12: @veronicagg  Here is the sample error message:

ERROR: Paths in x-ms-paths must overload a normal path in the paths section, i.e. a path in the x-ms-paths must either be same as a path in the paths section or a path in the paths sections followed by additional parameters.
        Path: file:///C:/workspace/azure-rest-api-specs/arm-apimanagement/2016-07-07/swagger/apimanagement.json#/x-ms-paths/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ApiManagement/service/{serviceName}/apis/{apiX}?export=true

PR #2016, opened 2017-03-15
Honeymoon of C# and TypeScript

olydis:
Improve communication and harmony between the two. Agenda:
- reliably rethrow exceptions across boundaries (no more unhandled promise rejections)

### DIFF
--- a/Tools/compare.ps1
+++ b/Tools/compare.ps1
@@ -97,6 +97,8 @@ $scrp = {
 
 @( "CSharp",
  "Azure.CSharp",
+ "Azure.CSharp.Fluent",
+ "AzureResourceSchema",
  "Ruby",
  "Azure.Ruby",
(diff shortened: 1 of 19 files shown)


amarzavery:
you didn't send me an invite of the marriage 📦

olydis:
@amarzavery The couple is very shy and celebrated the wedding in private. The benefits of marriage are still covered and may only be uncovered using `-FANCY` from the CLI. Instead of invoking the .NET CLI, the CLI args are converted into configuration that triggers appropriate plugins. 

In this mode, the C# side gives up all its rights to access the disk or have any impact on the course of life (the pipeline). This situation, the drastic shift of responsibilities, is new for both partners. The honeymoon will help the newly weds identify potential conflicts in this arrangement.

Issue #1928, opened 2017-03-07
InvalidConstraint: rule should be removed if covered by semantic validation


veronicagg:
If rule is covered by semantic validation, M6000, let's remove it from validation.

PR #2021, opened 2017-03-16
[linter] Removing InvalidConstraint as covered by semantic validator

vishrutshah:
Fixes https://github.com/Azure/autorest/issues/1928


vishrutshah:
Thanks for the review!!

Issue #1934, opened 2017-03-07
OperationsAPIImplementationValidation: update message


veronicagg:
Update output message with "Please expose operations API for this service "/providers/<yourProviderNameSpace>/operations""

PR #2026, opened 2017-03-17
[Linter] Correcting Operation API not implemeted rule and improving message

vishrutshah:
Fixes https://github.com/Azure/autorest/issues/1934

### DIFF
--- a/src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/positive/clean-complex-spec.json
+++ b/src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/positive/clean-complex-spec.json
@@ -19,16 +19,16 @@
     "application/json"
   ],
   "paths": {
-    "/operations": {
+    "/subscriptions/xxxx-xxxx-xxxx/providers/Microsoft.Cdn/operations": {
       "get": {
         "summary": "Lists all foo.",
-        "description":  "foo",
+        "description": "foo",
         "operationId": "Operations_List",
         "parameters": [
           {
             "name": "limit",
             "type": "string",
-            "description":  "foo"
+            "description": "foo"
           }
         ],
         "responses": {
--- a/src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/positive/clean-complex-spec.json
+++ b/src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/positive/clean-complex-spec.json
@@ -19,12 +19,15 @@
     "application/json"
   ],
   "paths": {
-    "/subscriptions/xxxx-xxxx-xxxx/providers/Microsoft.Cdn/operations": {
+    "/subscriptions/{subscriptionId}/providers/Microsoft.Cdn/operations": {
       "get": {
         "summary": "Lists all foo.",
         "description": "foo",
         "operationId": "Operations_List",
         "parameters": [
+          {
+            "$ref": "#/parameters/SubscriptionIdParameter"
+          },
           {
             "name": "limit",
             "type": "string",
--- a/src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
+++ b/src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
@@ -14,7 +14,8 @@ namespace AutoRest.Swagger.Model.Utilities
     {
         private const string XmsPageable = "x-ms-pageable";
         private static readonly Regex TrackedResRegEx = new Regex(@".+/Resource$", RegexOptions.IgnoreCase);
-        
+        private static readonly Regex resourceProviderPathPattern = new Regex(@"/providers/(?<resPath>[^{/]+)/", RegexOptions.IgnoreCase);
+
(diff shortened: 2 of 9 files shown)


src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -14,6 +14,7 @@ public static class ValidationUtilities
     {
         private static readonly string XmsPageable = "x-ms-pageable";
         private static readonly Regex TrackedResRegEx = new Regex(@".+/Resource$", RegexOptions.IgnoreCase);
+        private static readonly Regex resourceProviderPathPattern = new Regex(@"/providers/(?<resPath>[^{/]+)/", RegexOptions.IgnoreCase);
```
dsgouda: If I understand correctly, the string following``` providers/``` is your resource provider, everything else should be resourcetype and resourcetypename. In other words, a URL should have one resource provider only, Am I missing something here?

vishrutshah: You are right. The current implementation is just checking whether we have any path that ends with `/operations` and that can bring false positive as we'd want to flag out the once that ending with `providers/{name}/operations` and technically we can have more than one providers in the same swagger (although we do not want that and we have other rule that catches that but at the same time we do not want to miss out when we have more than one provider and only one of them have /providers/{name}/operations). 

We can make this simple to just catch for one provider and assume that as user fixes the swagger we'll surface other errors. It's upto us I am find with both approach :)

vishrutshah: Per offline talk with deepak, I've updated rule to check only for one RP in case of multiple RPs found in swagger.

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -170,5 +171,26 @@ public static bool IsReferenceOfType(string reference, Dictionary<string, Schema
...
+        /// <summary>
+        /// Returns array of resource providers
+        /// </summary>
+        /// <param name="paths">Dictionary of paths to look for</param>
+        /// <returns>Array of resource providers</returns>
+        public static string[] GetResourceProviders(Dictionary<string, Dictionary<string, Operation>> paths)
+        {
+            if (paths == null)
+            {
+                throw new System.ArgumentNullException(nameof(paths));
```
veronicagg: Is someone catching this exception at a higher level in the stack or is this method not being invoked when 'paths' is null, I'm wondering because we may have specs that don't have paths (like the ones we're starting to see with just model definitions) and I wouldn't want the rule to fail for this.

vishrutshah: that's good point. I've removed it so it'll not throw

src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/positive/clean-complex-spec.json
```diff
@@ -19,16 +19,16 @@
     "application/json"
   ],
   "paths": {
-    "/operations": {
+    "/subscriptions/xxxx-xxxx-xxxx/providers/Microsoft.Cdn/operations": {
```
dsgouda: nit: I'm not comfortable with ```xxx-xxx```, why not use ```{subscriptionId}```

src/modeler/AutoRest.Swagger/Validation/UniqueResourcePaths.cs
```diff
@@ -4,32 +4,24 @@
...
         {
-            var resources = paths.Keys
-                .SelectMany(path => resPathPattern.Matches(path)
-                    .OfType<Match>()
-                    .Select(match => match.Groups["resPath"].Value.ToString()))
-                .Distinct()
-                .ToArray();
-            formatParameters = new [] { string.Join(", ", resources) };
-            return resources.Length <= 1;
+            string[] resourceProviders = ValidationUtilities.GetResourceProviders(paths);
```
dsgouda: Good stuff! 👍

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -170,5 +171,21 @@ public static bool IsReferenceOfType(string reference, Dictionary<string, Schema
 
             return false;
         }
+
+        /// <summary>
+        /// Returns array of resource providers
+        /// </summary>
+        /// <param name="paths">Dictionary of paths to look for</param>
+        /// <returns>Array of resource providers</returns>
+        public static string[] GetResourceProviders(Dictionary<string, Dictionary<string, Operation>> paths)
```
dsgouda: nit: I prefer returning IEnumerables instead of arrays, allows more functionality

vishrutshah:
@azuresdkci Test this please

vishrutshah:
@azuresdkci Test this please

vishrutshah:
Thanks for the review!!

Issue #1938, opened 2017-03-07
ServiceDefinitionParameters: update rule


veronicagg:
Update: does it have host set to management.azure.com or x-ms-parameterized-host, if it is azure resource, if it's anywhere in the paths, then it should be globally defined, api-version should exist. Messages should be accurate to what's violating the rul.

ID: M2014
Category: SDK Violation
Severity: Error

vishrutshah:
Closing this rule based on https://github.com/Azure/autorest/issues/1940. We do not want distinction on rule based on host.

vishrutshah:
Based on offline sync with @dsgouda & @veronicagg , we are not tailoring any rule targeting towards Azure Management Plane. Hence we do not check on `management.azure.com` or `x-ms-parameterized-host` at all for now. Once we have infrastructure ready, we can update this rule to tailor towards specific planes.

Issue #1940, opened 2017-03-07
Implement rule categorization around azure/azure management/non-azure


veronicagg:
Add the following logic to the rule, if host set to management.azure.com or x-ms-parameterized-host, then only https is allowed. Update message to be accurate depending on what condition is not met. For non-azure services, it could be http or https.

ID: M2043
Category: SDK Violation
Severity: Error

dsgouda:
Talked to @fearthecowboy and @olydis and they agree that we should not be making any assumptions about whether a swagger spec is Azure management related or not, which means we cannot differentiate based on host. This rule boils down to checking whether the schemes are https or not.

veronicagg:
@dsgouda based on our offline conversation, we'll keep this rule as is for now, and we'll work on implementing a more structured differentiation of rules that apply to azure/azure management/non-azure specs, so that this one can be split in those buckets.

veronicagg:
We're extending categorization also for "rule applicable to an individual swagger file" versus "rule applicable to a composed swagger"
For example, OperationsAPI should applicable to a composed swagger, so if it's in one of them, the rule is not violated. 
We may have another case with api-version, in which we'd like all operations in one file to be only for one api-version, but a composed swagger could have multiple api-versions involved. So this type of rule should pass only for individual files.

dsgouda:
As step 0 @olydis will implement an AutoRest feature that will enable us to run validation over composite/single specs based on an input argument. This argument will then be passed to the Validator module based on which we can select what rules to run.

olydis:
Stub: https://github.com/Azure/autorest/pull/2164
Enhance with further categories as necessary ;-)

PR #2027, opened 2017-03-17
[Linter] Correcting rule to constrain for Azure management plane only

vishrutshah:
fixes https://github.com/Azure/autorest/issues/1938

### DIFF
--- a/src/core/AutoRest.Core/Properties/Resources.resx
+++ b/src/core/AutoRest.Core/Properties/Resources.resx
@@ -325,7 +325,7 @@ Modelers:
     <value>Parameters "subscriptionId" and "api-version" are not allowed in the operations section, define these in the global parameters section instead</value>
   </data>
   <data name="ServiceDefinitionParametersMissingMessage" xml:space="preserve">
-    <value>Parameter "{0}" is referenced but not defined in the global parameters section of Service Definition</value>
+    <value>Parameters "subscriptionId" and "api-version" must be defined in the parameters section in the swagger for the Azure Management Plane service.</value>
   </data>
   <data name="UniqueResourcePaths" xml:space="preserve">
     <value>Multiple resource providers are not allowed in a single spec. More than one the resource paths were found: '{0}'.</value>
--- a/src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
+++ b/src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
@@ -13,31 +13,20 @@ namespace AutoRest.Swagger.Validation
 {
     public class ServiceDefinitionParameters : TypedRule<Dictionary<string, SwaggerParameter>>
     {
-        private const string SubscriptionId = "subscriptionid";
-        private const string ApiVersion = "api-version";
+        private static readonly string SubscriptionId = "subscriptionId";
+        private static readonly string ApiVersion = "api-version";
+        private static readonly string AzureManagementHost = "management.azure.com";
+        private static readonly string XMsParameterizedHostExtensionName = "x-ms-parameterized-host";
 
         /// <summary>
-        /// This rule passes if the parameters section contains both subscriptionId and api-version parameters
+        /// Id of the Rule.
         /// </summary>
-        /// <param name="paths"></param>
-        /// <returns></returns>
-        public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string, SwaggerParameter> ParametersMap, RuleContext context)
-        {
-            var serviceDefinition = (ServiceDefinition)context.Root;
-            // Check if subscriptionId is used but not defined in global parameters
-            bool isSubscriptionIdReferenced = serviceDefinition.Paths.Keys.Any(key => key.ToLower().Contains("{" + SubscriptionId.ToLower() + "}"));
-            if (isSubscriptionIdReferenced && (ParametersMap?.Values.Any(parameter => parameter.Name?.ToLower().Equals(SubscriptionId) == true)) == false)
-            {
-                yield return new ValidationMessage(new FileObjectPath(context.File, context.Path), this, SubscriptionId);
-            }
-            // For ARM specs, api version is almost always required, call it out if it isn't defined in the global params
-            // We are not distinguishing between ARM and non-ARM specs currently, so let's apply this for all specs regardless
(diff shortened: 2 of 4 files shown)


dsgouda:
@olydis thoughts?

vishrutshah:
Closing the issue based on offline talk with @dsgouda https://github.com/Azure/autorest/issues/1940

vishrutshah:
@azuresdkci Test this please

vishrutshah:
@azuresdkci Test this please

vishrutshah:
@azuresdkci Test this please

src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -328,7 +328,7 @@ Modelers:
     <value>Parameters "subscriptionId" and "api-version" are not allowed in the operations section</value>
   </data>
   <data name="ServiceDefinitionParametersMissingMessage" xml:space="preserve">
-    <value>Parameters "subscriptionId" or "api-version" are referenced but not defined in the parameters section of Service Definition</value>
+    <value>Parameters "subscriptionId" and "api-version" must be defined in the parameters section in the swagger for the Azure Management Plane service.</value>
```
dsgouda: 1. Update this message
2. There is a different rule that flags a case where subscriptionId and ApiVersion are not defined in the global parameters section, you might want to modify the message to reflect that these parameters are not expected in the operations' parameters section

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
```diff
@@ -33,7 +37,27 @@ public class ServiceDefinitionParameters : TypedRule<Dictionary<string, SwaggerP
...
+        /// <param name="context">Rule context</param>
+        /// <param name="formatParameters">Formatted parameters</param>
+        /// <returns></returns>
+        public override bool IsValid(Dictionary<string, SwaggerParameter> ParametersMap, RuleContext context, out object[] formatParameters)
+        {
+            formatParameters = new object[0];
+            ServiceDefinition serviceDefinition = context.Root;
+            
+            // Rule does not apply if x-ms-parameterized-host extension is not applied
+            if (!serviceDefinition.Extensions.ContainsKey(XMsParameterizedHostExtensionName))
```
dsgouda: If we are not checking the management host, this becomes redundant, just simplify this to checking operation parameters instead

dsgouda:
Please check you have the tests updated to reflect this rule

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
```diff
@@ -33,7 +37,27 @@ public class ServiceDefinitionParameters : TypedRule<Dictionary<string, SwaggerP
...
+            {
+                return true;
+            }
 
+            return true == (ParametersMap?.Values.Any(parameter => parameter.Name == SubscriptionId || parameter.Name == ApiVersion));
```
dsgouda: hmm... totally missed this, should be case insensitive right?

veronicagg:
@vishrutshah looks like CI is failing and branch has some conflicts.

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
vishrutshah: I think this should be case-sensitive. That'd also facilitate us towards common definitions. Ref: https://github.com/Azure/azure-rest-api-specs-pr/pull/40/files#diff-d3a0dcdc5d2939703fa9e32dff2295c1R249

src/core/AutoRest.Core/Properties/Resources.resx
vishrutshah: Could you please point me to the rule that validates that they must be in global parameter section? 

Based on rule id [M2014](https://github.com/Azure/azure-rest-api-specs/blob/master/documentation/swagger-checklist.md) seems like this rule supposed to do that but current implementation doesn't differentiate.

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
dsgouda: Agreed, but existing specs may not be following any convention

src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -325,7 +325,7 @@ Modelers:
     <value>Parameters "subscriptionId" and "api-version" are not allowed in the operations section, define these in the global parameters section instead</value>
   </data>
   <data name="ServiceDefinitionParametersMissingMessage" xml:space="preserve">
-    <value>Parameter "{0}" is referenced but not defined in the global parameters section of Service Definition</value>
+    <value>Parameters "subscriptionId" and "api-version" must be defined in the parameters section in the swagger for the Azure Management Plane service.</value>
```
veronicagg: minor: I think we could also say for "Azure Resource Management service", not sure what's best, so feel free to pick

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
vishrutshah: I'd like to catch those getting caught with these rule then.

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
```diff
@@ -50,7 +36,21 @@ public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<
...
+        /// <param name="ParametersMap">Dictionary of swagger parameters</param>
+        /// <param name="context">Rule context</param>
+        /// <param name="formatParameters">Formatted parameters</param>
+        /// <returns></returns>
+        public override bool IsValid(Dictionary<string, SwaggerParameter> ParametersMap, RuleContext context, out object[] formatParameters)
+        {
+            formatParameters = new object[0];
+            ServiceDefinition serviceDefinition = context.Root;
+
+            return true == (ParametersMap?.Values.Any(parameter => parameter.Name == SubscriptionId || parameter.Name == ApiVersion));
```
dsgouda: Definitely think this check should be case insensitive

PR #2033, opened 2017-03-17
[Go] Fixing up some duplicate consts.

marstr:
Taking advantage of consts already declared in the Core.

### DIFF
--- a/src/generator/AutoRest.Go/Extensions.cs
+++ b/src/generator/AutoRest.Go/Extensions.cs
@@ -20,6 +20,10 @@ namespace AutoRest.Go
 
     public const string ReadOnlyConstraint = "ReadOnly";
 
+    private static readonly Regex IsApiVersionPattern = new Regex(@"^api[^a-zA-Z0-9_]?version", RegexOptions.IgnoreCase);
+
+    private static readonly Regex UnwrapAnchorTagsPattern = new Regex("([^<>]*)<a\\s*.*\\shref\\s*=\\s*[\'\"]([^\'\"]*)[\'\"][^>]*>(.*)</a>");
+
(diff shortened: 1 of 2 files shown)


marstr:
Hmm, the diff suggests that some other changes to the file got inadvertently included. I'm going to backout this change so that only the const updates get included.

marstr:
Actually, looking it over yet again, it seems the only other changes are new-line changes. I'll keep this PR open.

marstr:
Cleaning up some Regex declarations while I was at it.

src/generator/AutoRest.Go/Extensions.cs
```diff
@@ -20,6 +20,10 @@ public static class Extensions
 
     public const string ReadOnlyConstraint = "ReadOnly";
 
+    private static readonly Regex IsApiVersionPattern = new Regex(@"^api[^a-zA-Z0-9_]?version", RegexOptions.IgnoreCase);
+
+    private static readonly Regex UnwrapAnchorTagsPattern = new Regex("([^<>]*)<a\\s*.*\\shref\\s*=\\s*[\'\"]([^\'\"]*)[\'\"][^>]*>(.*)</a>");
+
```
mcardosos: nit, does it make sense to get this variables outside the func where they are used?

mcardosos: I guess it is so they are only calculated once?

marstr: Correct. I agree with the sentiment of keeping things scoped to the method they are used in. This time, though, there is enough value in only spinning up the Regex only once that I think it is worth it. Ideally, we'd even look at adding in the RegexOptions.Compiled flag. I wanted to add a way to evaluate the perf impact before I did that though, because the additional start-up time may not be worth it.

PR #2036, opened 2017-03-19
Fix exception when running AutoRest with empty arguments

fhoering:
Exception: 
System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at AutoRest.Program.Main(String[] args)

Regression of commit:
https://github.com/Azure/autorest/commit/5d824d19b6f861daccf110042c6bc026fc402e57


olydis:

PR #2038, opened 2017-03-21
Fix removing flattened types in arrays/dictionaries

jianghaolu:
@amarzavery @anuchandy 

In a Swagger, if we have type `Foo` flattened in `FooWrapper`, when we check if we should remove `Foo`, we only check if `Foo` is present as is in parameters, responses, properties, error types, etc. But we never check if `Foo` is present in a list or dictionary. This causes `List<Foo>` to fail to compile.


jianghaolu:
Reviewed by Amar yesterday at my desk. Merging to unblock Navy.

PR #2046, opened 2017-03-22
tags based grouping

erendrake:
This is just a first stab at using tags to group operations, the same way that swagger codegen and the swagger ui group them. I didnt want to get too far before i gauge interest in it

still TODO

- [ ] Support for one operation in multiple groups. I dont need this but its available in the spec
- [ ] Documentation of the setting

Again if there is no interest in having this feature please let me know and ill drop it :)

### DIFF
--- a/src/core/AutoRest.Core/Settings.cs
+++ b/src/core/AutoRest.Core/Settings.cs
@@ -73,7 +73,13 @@ Licensed under the MIT License. See License.txt in the project root for license 
             ValidationLevel = Category.Error;
             ModelsName = "Models";
             CodeGenerationMode = "rest-client";
+            GroupingMethod = "underscore";
         }
+
+        /// <summary>
+        /// Gets or sets the method for grouping operations
+        /// </summary>
+        public string GroupingMethod { get; set; }
 
         /// <summary>
         /// Gets or sets the IFileSystem used by code generation.
--- a/src/modeler/AutoRest.Swagger/SwaggerModeler.cs
+++ b/src/modeler/AutoRest.Swagger/SwaggerModeler.cs
@@ -336,16 +336,25 @@ namespace AutoRest.Swagger
         {
             if (operation == null)
             {
-                throw new ArgumentNullException("operation");
+                throw new ArgumentNullException(nameof(operation));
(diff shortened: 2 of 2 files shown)


olydis:
Interesting PR! We're currently in the process of redesigning AutoRest a bit since the amount of settings (which currently completely relies on `Settings.cs`) has a tendency to grow faster and faster, complicating our CLI, etc.... we'll gladly look at this when we have figured things out and have a good story of where to put `GroupingMethod` :-)

fearthecowboy:
So... we have a way to do that without changing AutoRest:

If you switch to using a configuration file for running AutoRest (use `autorest init` to get an example) , you can use a `directive` to modify the model.

Make a readme.md in your folder something like this:

~~~ markdown
# MyPetStore
> see https://aka.ms/autorest

This is the AutoRest configuration file for the MyPetstore.

---
## Getting Started 
To build the SDK for MyPetstore  simply [Install AutoRest](https://aka.ms/autorest/install) and in this folder, run:

> `autorest`

To see additional help and options, run:
 
> `autorest --help`
---

#### Basic Information 
These are the global settings for MyAPI.

``` yaml
# list all the input OpenAPI files (may be YAML, JSON, or Literate- OpenAPI markdown)
input-file:
  - petstore.json

# this allows you to programatically tweak the swagger file before it is modeled.
directive:
  from: swagger-document # do it globally 
  where: $.paths.*.* 
  
  # if every operation has a tag, you can be simple:
  transform: $.operationId = `${$.tags[0]}_${$.operationId}`
  
  # otherwise if you have some things that don't have a `tags` use this:
  # transform: $.operationId = `${( $.tags || [] )[0] || "some_prefix_here" }_${$.operationId}`  
   
```

---
#### Language-specific settings: CSharp

These settings apply only when `--csharp` is specified on the command line.

``` yaml $(csharp) 
csharp:
  namespace: My.PetStore
```

~~~

and then run 

`autorest --csharp readme.md`

and it will use the first tag for each operation as the group name.

erendrake:
@fearthecowboy groovy. ill close this and use the settings file.

PR #2050, opened 2017-03-22
Update readme.md

sergey-shandar:
Fix markdown headers.


olydis:
uhm, I think we had the same thoughts: https://github.com/Azure/autorest/pull/2049 👍

Issue #1936, opened 2017-03-07
ResourceModelValidation: review how to identify a "Resource"


veronicagg:
How should resource be found? A model is an allOf of a definition that has x-ms-azure-resource=true

veronicagg:
@dsgouda @ravbhatnagar @amarzavery @vishrutshah @sarangan12 
We've updated this definition to be the following:
Resources list should be built as the union of:
1. Any model that's returned by a PUT operation
2. Any model that has an "allOf" of "Resource" at any level (name of the root model is "Resource") AND it's in 200 response of a GET
3. Any model that has x-ms-azure-resource at some level (excluding the model that actually has x-ms-azure-resource tag - since this one should be the base resource)

Resources can be divided between "Tracked Resources" and "Proxy Resources". A Resource as defined above which contains "location" property as required, is a "Tracked Resource"; any other Resource, is a "Proxy Resource".

veronicagg:
I talked to @olydis about where a good place to put this logic in autorest would be, since we should run this logic once per model and have it available to all our rules. IMO, the Resource list is like a "view" of the swagger model. He suggested to create a class in parallel with RuleContext, though RuleContext doesn't seem to be in the right place in AutoRest. He'll be creating a pull request to move RuleContext most likely to AutoRest.Swagger. After that, we probably need to create a class that has a "Resource" list property and exposes methods to obtain "Resources", "Tracked Resources" and "Proxy Resources".

dsgouda:
So.... just had a thought:
We are saying that model definitions that "allOf" on a Resource type are ResourceModels. What happens if a model definition has a property that references a Resource (of any type)? Are we sure they are not of "Resource" type? I believe there shouldn't be instances that do this currently in our rest spec repo but it would be worth consideration

@ravbhatnagar @amarzavery thoughts?

dsgouda:
On similar lines, when determining a tracked resource, do we have to traverse the property references in order to search for a property named "location"

ravbhatnagar:
@dsgouda As discussed over IM, the scenario in your first comment is highly unlikely to happen. And even if it happens, we should not treat it as a "Resource" since all the properties that make the model a resource are nested inside another property. Same applies to the location as well. We should not traverse property references to find "location"

veronicagg:
Copying from Sarangan's summary:
Resource Determination logic

1. Get the list of model definitions that does not contain x-ms-azure-resource: true. 

[...]

PR #2056, opened 2017-03-23
Improved logic to determine resource models

dsgouda:
Addresses issue https://github.com/Azure/autorest/issues/1936
@olydis Addressed your comments

@veronicagg @sarangan12 @vishrutshah FYI


src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -34,10 +41,238 @@ public static bool IsTrackedResource(Schema schema, Dictionary<string, Schema> d
...
+        {
+            if (serviceDefinition.Definitions?.Any() != true)
+            {
+                return new List<string>();
+            }
+
+            var xmsAzureResourceModels = GetXmsAzureResourceModels(serviceDefinition.Definitions);
+
+            // Get all models that are returned by PUT operations (200/201 response)
+            var putOperationsResponseModels = GetOperationResponseModels("put", serviceDefinition).Except(xmsAzureResourceModels);
```
olydis: now we do `Except(xmsAzureResourceModels)` only on 200 responses of PUTs, but not on 201? Also, note that https://github.com/Azure/autorest/issues/1936#issuecomment-288244278 still doesn't mention that.

Issue #730, opened 2016-02-24
Duplicate local variables when parameter flattening


emgerner-msft:
In some cases when parameter objects are flattened (in the case where they have 2 or less properties), non-compiling code is generated. It looks like an issue where when we flatten parameters we name the parameter object variable the same as a method parameter value. 

For example:

public async Task<AzureOperationResponse> UpdateWithHttpMessagesAsync(string resourceGroupName, string availabilitySetName, IDictionary<string, string> tags, ...)

And later we do:

            UpdateParameters tags = null; // Here's where we set it a second time
            if (tags != null)
            {
                tags = new UpdateParameters();
                tags.Tags = tags;
            }

PR #2059, opened 2017-03-23
Fix absolute path

sarangan12:
@vishrutshah @veronicagg @salameer @devigned @NiklasGustafsson Please review and approve

Tested locally against the SDK.

Related to: #730. Please refer the issue for more context.

Comparison of Ruby SDK befor & after changes are available at: \\saran-dev1\shared\PR2059

To see how node sdk handles the task, refer: https://github.com/Azure/azure-sdk-for-node/blob/master/lib/services/cdnManagement/lib/cdnManagementClient.js#L160

### DIFF
--- a/src/generator/AutoRest.Ruby/Templates/MethodTemplate.cshtml
+++ b/src/generator/AutoRest.Ruby/Templates/MethodTemplate.cshtml
@@ -120,7 +120,7 @@ def @(Model.Name)_async(@(Model.MethodParameterDeclaration))
   @EmptyLine
 }
 
-  path_template = '@Model.Url'
+  path_template = '@((Model.Url.StartsWith("/") ? Model.Url.TrimStart('/') : (string)Model.Url))'
 
 @EmptyLine
   request_url = @@base_url || @(Model.ClientReference).base_url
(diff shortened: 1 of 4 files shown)


src/generator/AutoRest.Ruby/Templates/MethodTemplate.cshtml
```diff
@@ -120,7 +120,7 @@ def @(Model.Name)_async(@(Model.MethodParameterDeclaration))
   @EmptyLine
 }
 
-  path_template = '@Model.Url'
+  path_template = '@((Model.Url.StartsWith("/") ? Model.Url.TrimStart('/') : (string)Model.Url))'
```
vishrutshah: Does this change rely on information that `base_url` where the path is going to be appended must end with `/` ?

sarangan12: No. It does not rely on the information

sarangan12:
@vishrutshah Doing it in runtime will not be correct. If we do it in ms_rest, we are restricting the users (who use only ms_rest). i.e. if a user genuinely has both absolute path & relative paths, the user cannot use ms_rest. So, ms_rest should not be playing with the paths. 

Next, think about ms_rest_azure. You can do the change there. But, think of the users who use autorest with AutoRest.Ruby (not AutoRest.RubyAzure). We will come with the same issue. 

So, Yes. We need to do it in every operation file.

sarangan12:
Rel Issue: https://github.com/Azure/azure-sdk-for-net/issues/3007

vishrutshah:
@sarangan12 

Based on our discussion, https://github.com/Azure/autorest/blob/master/src/generator/AutoRest.CSharp/Templates/Rest/Client/MethodTemplate.cshtml#L133 the logic that .net generator is using. Our changes look fine but let's keep an eye on the issue and what's their view.

Issue #1980, opened 2017-03-10
ProvidersPathValidation: update rule


veronicagg:
Pattern needs to be better explained, message needs update, "operations" needs to be taken out, and pattern explained, plus additional documentation externally.

Additionally, looks like patter of /typename1/{typename1type} does not require the "{" "}", this value could be just one and therefore not a parameter, so the verification should be more in the line of "even" positions starting wiht the first "typename1" should be typenames, and "odd" positions typename1names, we probably want to flag the ones that don't have parameters as warnings, for the purpose of asking the author, is that's the intent (meaning that there's only one acceptable value for that typenametype. @amarzavery @ravbhatnagar please correct this understanding as necessary.

Please run on top 5 RPs to evaluate violations.
Needs to be discussed with Gaurav.

ID: ?
Category: RPC Violation
Severity: Warning

dsgouda:
I believe "/operations" needs to be taken out, I was wondering if we could just say 
subscriptions/{subscriptionId}/resourceGroups/{resourceGroup}/providers/namespace/type1name/{type1value}/type2name/{type2value}/<operationName>
where <operationName> can be empty
@veronicagg I don't think I understand the intent as you described. As per my understanding, we simply want to ensure that urls conform to the pattern ``` .../subscriptions/{subscriptionId}/resourceGroups/{resourceGroup}/.../providers/namespace/typename/{typevalue}/<operationName>``` and I have already run this against the rest spec repo with more or less accurate results, the only improvement needed here was the messaging
@amarzavery @ravbhatnagar please confirm

veronicagg:
@dsgouda let's get confirmation on the statement I made above from @ravbhatnagar, as this specific rule was discussed further a little later.

PR #2061, opened 2017-03-24
Fixing issue 1980

dsgouda:
Addressing issue https://github.com/Azure/autorest/issues/1980


src/modeler/AutoRest.Swagger/Validation/ProvidersPathValidation.cs
```diff
@@ -11,24 +12,39 @@ namespace AutoRest.Swagger.Validation
 {
     public class ProvidersPathValidation : TypedRule<Dictionary<string, Dictionary<string, Operation>>>
     {
-        private readonly Regex FullRegex = new Regex(@"\/Subscriptions\/\{.+\}\/ResourceGroups\/\{.+\}\/providers\/[^\/]+(\/[^\/]+\/\{[^\}]+\})*(\/[^\/]+)?([\?]?[^\/]+)*$", RegexOptions.IgnoreCase);
-        private readonly Regex ProviderRegex = new Regex(@"Subscriptions\/\{.+\}\/ResourceGroups\/\{.+\}\/providers\/.+$", RegexOptions.IgnoreCase);
+        private readonly Regex ProviderRegex = new Regex(@"Subscriptions\/\{.+\}\/ResourceGroups\/\{.+\}\/providers\/([^\/]+)(?<typenamesandvalues>[^?]+)([\?]?.+)?$", RegexOptions.IgnoreCase);
```
vishrutshah: As discussed offline, you may need to consider the case of last childtype that is `default` where we may not want to flag.

Also, please update my reged in JsonValidationListern so whoever need to update need to learn only one regex.

vishrutshah:
@dsgouda As discussed offline, could you please update JsonValidationMessages regex to 

`/providers/(?<providerNamespace>[^{/]+)((/(?<resource>[^{/]+)/)((?<resourceName>[^/]+)))+(/(?<resource>[^{/]+))`

src/modeler/AutoRest.Swagger/Validation/ProvidersPathValidation.cs
dsgouda: So, even if we consider the last child type, it would still end up in even positions ```0/typevalue0/2/typevalue2/4/typevalue4/6/typevalue6/default```
I think the logic here would suffice

I shall definitely update the regex in ````JsonValidationMessages``` and move it to ```ValidationUtilities```

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -20,9 +20,11 @@ public static class ValidationUtilities
 
         private static readonly Regex TrackedResRegEx = new Regex(@".+/Resource$", RegexOptions.IgnoreCase);
 
-        // This needs to be deprecated in favor of context.TrackedResources
+        public static readonly Regex ResourcePathPattern = new Regex(@"/providers/(?<providerNamespace>[^{/]+)((/(?<resource>[^{/]+)/)((?<resourceName>[^/]+)))+(/(?<resource>[^{/]+))");
+
         private static readonly Regex resourceProviderPathPattern = new Regex(@"/providers/(?<resPath>[^{/]+)/", RegexOptions.IgnoreCase);
 
+        // This needs to be deprecated in favor of context.TrackedResources
         public static bool IsTrackedResource(Schema schema, Dictionary<string, Schema> definitions)
```
veronicagg: when is this getting deprecated? which PR covers this?

dsgouda: This method is being referenced by existing tracked resources validation rules that @sarangan12 is working on improving, I have instructed him to remove this code once he has his changes ready

Issue #1921, opened 2017-03-06
AnonymousParameterTypes: update rule


veronicagg:
ID: M2026
Category: SDK Violation
Severity: Error

Rename class name to AnonymousBodyParameter. 2 locations for anonymous types: request body parameter, response schema. - design should evaluate the 2 rules (AnonymousParameterTypes, AvoidAnonymousTypes).

veronicagg:
Additional requirement for this rule:
If there are no properties defined for the parameter defined inline, then it's allowed. The rule should fail only if the object appears to be a complex object (has properties), for which autorest, would generate a model.

Issue #1917, opened 2017-03-06
AvoidAnonymousTypes: category, severity and message needs update


veronicagg:

veronicagg:
Message should explain what anonymous types mean.  Say something about "inline".
Category = SDK Violation
Severity: Error

If model is not explicitly defined, autorest will create a model called OperationName_parameter. This is fine in itself, but if this model should be shared with PUT, GET, PATCH operations, this would not occur, because it'll generate a model for each of the operations.

PR #2006, opened 2017-03-13
[Linter] AnonymousTypes: update rule Ids and messages

veronicagg:
Fixing https://github.com/Azure/autorest/issues/1917 and https://github.com/Azure/autorest/issues/1921.

PR #2062, opened 2017-03-24
Updating rule id and logic

veronicagg:
Fixing #1917 and #1921.


veronicagg:
I've opened a new pr for #2006, to fix history.

src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/anonymous-response-type.json
```diff
@@ -25,9 +25,15 @@
         "responses": {
           "default": {
             "description": "Unexpected error",
-            "schema": {
+			"schema": {
```
dsgouda: nit: formatting

src/modeler/AutoRest.Swagger/Validation/AnonymousBodyParameter.cs
```diff
@@ -29,8 +36,13 @@ public class AnonymousParameterTypes : TypedRule<SwaggerParameter>
...
 
         /// <summary>
-        /// The severity of this message (ie, debug/info/warning/error/fatal, etc)
+        /// An entity fails this rule if it has a schema, and that schema is an anonymous type
         /// </summary>
-        public override Category Severity => Category.Info;
+        /// <param name="entity"></param>
+        /// <returns></returns>
+        public override bool IsValid(SwaggerParameter entity) =>
+            entity?.Schema == null || AnonymousTypesRule.IsValid(entity.Schema);
```
dsgouda: No need to check ```entity``` for null, just pass ```entity.Schema``` to ```AnonymousTypesRule``` and check for null there

src/modeler/AutoRest.Swagger/Validation/AvoidAnonymousTypes.cs
```diff
@@ -26,9 +34,12 @@ public class AvoidAnonymousTypes : TypedRule<SwaggerObject>
         public override string MessageTemplate => Resources.AnonymousTypesDiscouraged;
 
         /// <summary>
-        /// The severity of this message (ie, debug/info/warning/error/fatal, etc)
+        /// An <paramref name="entity"/> fails this rule if it doesn't have a reference (meaning it's defined inline)
         /// </summary>
-        public override Category Severity => Category.Info;
+        /// <param name="entity">The entity to validate</param>
+        /// <returns></returns>
+        public override bool IsValid(SwaggerObject entity) => (entity?.GetType() == typeof(Schema) &&
```
dsgouda: Or, just do ```((Schema)entity.Properties?.Count??0) ==0```

src/modeler/AutoRest.Swagger/Validation/AnonymousBodyParameter.cs
```diff
@@ -3,22 +3,29 @@
...
 using AutoRest.Core.Logging;
 using AutoRest.Core.Properties;
-using AutoRest.Swagger.Validation.Core;
 using AutoRest.Swagger.Model;
+using AutoRest.Swagger.Validation.Core;
 
 namespace AutoRest.Swagger.Validation
 {
-    public class AnonymousParameterTypes : TypedRule<SwaggerParameter>
+    public class AnonymousBodyParameter : TypedRule<SwaggerParameter>
```
vishrutshah: You may move `Id`, `ValidationCategory`, `Severity` & `MessageTemplate ` into top level `AnonymousTypeRule<T>` abstract class that would help in removing 13 for `new AvoidAnonymousTypes()`. upto you not required

src/modeler/AutoRest.Swagger/Validation/AnonymousBodyParameter.cs
veronicagg: @dsgouda if Schema is null then isValid method should return true, that's why I'm checking for it here.

src/modeler/AutoRest.Swagger/Validation/AvoidAnonymousTypes.cs
veronicagg: @dsgouda Similarly in this case, I'd like IsValid to return true if Properties is null or if there are 0 properties, so I believe I need to keep the check here.

src/modeler/AutoRest.Swagger/Validation/AnonymousBodyParameter.cs
veronicagg: I'll probably leave as is for now, thanks.

src/modeler/AutoRest.Swagger/Validation/AvoidAnonymousTypes.cs
dsgouda: If properties is null, the expression I suggested will evaluate to 0 and pass the if condition

PR #2070, opened 2017-03-27
Updating to only https per ARM

veronicagg:


src/core/AutoRest.Core/Properties/Resources.Designer.cs
```diff
@@ -674,7 +674,7 @@ public class Resources {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Non-HTTPS/HTTP schemes have limited support.
```
dsgouda: nit: I'm assuming you have verified this is what the error message should read like

veronicagg: @dsgouda I'm basically representing what the rule does in the message, so if it's wrong, we'd need to re-evaluate the rule.

Issue #1967, opened 2017-03-08
ARMResourcePropertiesBag: update rule, reporting many false positives


veronicagg:
@ravbhatnagar Could you clarify the spec for this rule? We believe the current implementation is too strict. 

Check DataFactory swagger for this rule, this rule maybe too strict. We may only need to validate that the top level properties in the property bag of a "Resource" like object, should not have id, name, type, location, tags. It is ok if the model has a property named id one level down than that. For example "StorageAcount.properties.foo.id" is ok but "StorageAcount.properties.id" is not .

Category: RPC Violation
Severity: Error

veronicagg:
it's also flagging "properties" as not allowed - see https://github.com/Azure/azure-rest-api-specs/issues/995

ravbhatnagar:
@veronicagg - 2 rules are in question here- one is the allowed set of properties at the top level, second is that the top level properties should not be repeated inside properties.
For second - we should definitely exclude "properties" from the check. Also lets just check only one level deep and so StorageAccount.properties.id is a violation but StorageAccount.properties.foo.id is not. 

For first - since its checking the allowed set of top level properties, "properties" should be a part of the check. 

One thing to note here is that these properties can change. For example- ARM may add another top level property soon. So how can we minimize the impact on the code for the rule to change. Can the rule get the allowed set of properties from common definitions or some other config file?

veronicagg:
@ravbhatnagar thanks! Could you confirm this applies to "Resource" (and not only to "Tracked Resource")?

ravbhatnagar:
@veronicagg - Yes, this applies to the Resource

PR #2073, opened 2017-03-28
Fix ArmPropertiesBag validation rule 

dsgouda:
Addresses issue https://github.com/Azure/autorest/issues/1967


dsgouda:
@azuresdkci Test this please

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
```diff
@@ -220,6 +220,13 @@ public void ListByOperationsValidation()
         }
 
         [Fact]
+        public void ArmResourcePropertiesBagValidation()
+        {
+            var messages = ValidateSwagger(Path.Combine(Core.Utilities.Extensions.CodeBaseDirectory, "Resource", "Swagger", "Validation", "arm-resource-properties-bag.json"));
+            messages.AssertOnlyValidationMessage(typeof(ArmResourcePropertiesBag), 1);
```
veronicagg: which property violates the rule in the json file?

src/modeler/AutoRest.Swagger/Validation/ArmResourceePropertiesBag.cs
```diff
@@ -0,0 +1,62 @@
...
+        private static readonly IEnumerable<string> ArmPropertiesBag = new List<string>()
+                                                                        { "name", "id", "type", "location", "tag" };
+
+        private bool ContainsReservedProperties(IEnumerable<string> propNames) => propNames?.Intersect(ArmPropertiesBag).Any() == true;
+
+        // Verifies if a tracked resource has a corresponding get operation
+        public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string, Schema> definitions, RuleContext context)
+        {
+            var resModels = context.ResourceModels;
+            var violatingModels = resModels.Where(res => ContainsReservedProperties(definitions[res].Properties?.Keys));
```
veronicagg: is this condition checking on top level properties? If so, I think it's not necessary in this rule, and it would overlap with: https://github.com/Azure/autorest/blob/master/src/modeler/AutoRest.Swagger/Validation/BodyTopLevelProperties.cs 
Thoughts?

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
dsgouda: ```TempResource``` has a property named ```name```, I can add another property inside a ```properties``` object as an additional check here

src/modeler/AutoRest.Swagger/Validation/ArmResourceePropertiesBag.cs
dsgouda: Agreed, seems like an overlap. I wonder however if that rule should be renamed ```BodyTopLevelProperties``` doesn't tell much

veronicagg: Feel free to rename if you come up with a better name :)

PR #2075, opened 2017-03-28
Updating Rule ID and Violation category

### LINKED ISSUE #1938: ServiceDefinitionParameters: update rule

vishrutshah:
Closing PR https://github.com/Azure/autorest/pull/2027

fixes #1938


vishrutshah:
Thanks for the review!!

Issue #1746, opened 2017-01-24
[Go] Generator for composite swaggers is not including API version


mcardosos:
Try to generate the network or the compute composite swaggers.

salameer:
This wil be released in this milestone v9.0.0-beta-20170228
https://github.com/Azure/azure-sdk-for-go/milestone/6

marstr:
Fixed in PR #2080

marstr:
Done.

TamalSaha:
@marstr , does this mean composite swaggers are now supported?

marstr:
Yep!

For V9, we're releasing web as generated from it's composite swagger. In v10 we will have validated the rest of the composite swaggers and moved to that model.

PR #2080, opened 2017-03-28
Removing APIVersion from client.

marstr:
Instead, a const is generated in each method that will serve to provide the APIVersion if present. This was done in order to ensure that composite swaggers which have multiple APIVersions associated with their operations are supported.

One upside, is that APIVersion will no longer be present in swaggers that do not contain it. Pet Store should not have a reference to APIVersion.

Regenerating samples.

### DIFF
--- a/src/generator/AutoRest.Go/Model/CodeModelGo.cs
+++ b/src/generator/AutoRest.Go/Model/CodeModelGo.cs
@@ -145,7 +145,7 @@ namespace AutoRest.Go.Model
     public string GlobalDefaultParameters
     {
       get
-      {   
+      {
         var declarations = new List<string>();
         foreach (var p in Properties)
         {
--- a/src/generator/AutoRest.Go/Model/CodeModelGo.cs
+++ b/src/generator/AutoRest.Go/Model/CodeModelGo.cs
@@ -145,7 +145,7 @@ namespace AutoRest.Go.Model
     public string GlobalDefaultParameters
     {
       get
-      {
+      {   
         var declarations = new List<string>();
         foreach (var p in Properties)
         {
--- a/src/generator/AutoRest.Go/Model/MethodGo.cs
+++ b/src/generator/AutoRest.Go/Model/MethodGo.cs
@@ -22,6 +22,8 @@ namespace AutoRest.Go.Model
 
     public string PackageName { get; private set; }
 
+    public string APIVersion {get; private set; }
+
     private readonly string lroDescription = " This method may poll for completion. Polling can be canceled by passing the cancel channel argument. " +
                                              "The channel will be used to cancel polling and any outstanding HTTP requests.";
 
@@ -40,6 +42,18 @@ namespace AutoRest.Go.Model
       Owner = (MethodGroup as MethodGroupGo).ClientName;
       PackageName = cmg.Namespace;
       NextAlreadyDefined = NextMethodExists(cmg.Methods.Cast<MethodGo>());
+
+      var apiVersionParam =
+        from p in Parameters
+        let name = p.SerializedName.Value
+        where name != null && name.IsApiVersion()
+        select p.DefaultValue.Value?.Trim(new[]{'"'});
+
+      APIVersion = apiVersionParam.SingleOrDefault();
+      if(APIVersion == default(string))
+      {
+        APIVersion = cmg.ApiVersion;
+      }
 
       var parameter = Parameters.ToList().Find(p => p.ModelType.PrimaryType(KnownPrimaryType.Stream)
                                           && !(p.Location == ParameterLocation.Body || p.Location == ParameterLocation.FormData));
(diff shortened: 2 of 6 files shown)


src/generator/AutoRest.Go/Model/CodeModelGo.cs
```diff
@@ -145,7 +145,7 @@ public string HelperGlobalParameters
     public string GlobalDefaultParameters
     {
       get
-      {
+      {
```
mcardosos: Nit, spaces

marstr: Derp, I'll fix that.

marstr:
This PR fixes issue #1746

src/generator/AutoRest.Go/Model/ParameterGo.cs
```diff
@@ -76,8 +86,9 @@ public string ValueForMap()
     {
       if (SerializedName.Value.IsApiVersion())
```
mcardosos: I would replace all `SerializedName.Value.IsApiVersion()` with `IsAPIVersion`

mcardosos:
Please include a sample of how code would look like. It is easier to understand the changes too.

src/generator/AutoRest.Go/Model/ParameterGo.cs
marstr: Fixed locally.

marstr:
@mcardosos:
> Please include a sample of how code would look like. It is easier to understand the changes too.

How about instead of adding a sample here, I update Azure/azure-sdk-for-go#569 with the newly generated code and link that commit?

src/generator/AutoRest.Go/Model/MethodGo.cs
```diff
@@ -41,6 +43,18 @@ internal void Transform(CodeModelGo cmg)
...
       NextAlreadyDefined = NextMethodExists(cmg.Methods.Cast<MethodGo>());
 
+      var apiVersionParam =
+        from p in Parameters
+        let name = p.SerializedName.Value
+        where name != null && name.IsApiVersion()
+        select p.DefaultValue.Value?.Trim(new[]{'"'});
+
+      APIVersion = apiVersionParam.SingleOrDefault();
+      if(APIVersion == default(string))
```
jhendrixMSFT: Curious why this check is necessary.

src/generator/AutoRest.Go/Model/ParameterGo.cs
```diff
@@ -42,14 +42,24 @@ public string AddToMap(string mapVariable)
 
     public string GetParameterName()
     {
-      return IsClientProperty
-                      ? "client." + Name.Value.Capitalize()
-                      : Name.Value;
+      string retval;
+      if (IsAPIVersion) {
```
jhendrixMSFT: Nit, formatting of curly braces is inconsistent.

src/generator/AutoRest.Go/Model/ParameterGo.cs
```diff
@@ -225,8 +236,8 @@ public static string Validate(this IEnumerable<ParameterGo> parameters, HttpMeth
 
       foreach (var p in parameters)
       {
-        var name = p.SerializedName.Value.IsApiVersion()
-            ? "client." + ParameterGo.ApiVersionName
+        var name = p.IsAPIVersion
+            ? "APIVersion"
```
jhendrixMSFT: Shouldn't we use the constant here?

src/generator/AutoRest.Go/Model/ParameterGo.cs
marstr: Fixed locally

src/generator/AutoRest.Go/Model/MethodGo.cs
marstr: Ahhh, so I hit at least one case where the APIVersion in the parameter list was the empty string. In that case, we should default back to the APIVersion defined at package level. Perhaps the above isn't super readable, how about this:

``` C#
if apiVersionParam.Any()
{
    APIVersion = apiVerisonParam.Single();
}
else
{
    APIVersion = cmg.ApiVersion;
}
```

jhendrixMSFT: Interesting.  I'm fine with either approach I would just add a comment with this explanation.

marstr: Fixed locally

src/generator/AutoRest.Go/Model/ParameterGo.cs
marstr: In my latest local version, I actually skip the parameters that return true for IsAPIVersion, so the string const is gone :)

marstr:
Alright folks (@mcardosos, @jhendrixMSFT ),

You can see the code that was generated by the version of the generator under review here:
https://github.com/Azure/azure-sdk-for-go/pull/569/commits/ac61e9baa67d949d8a8a2df95ed4abe5cd49f001

Note, there's another commit after this one for handling more recent swaggers as well.

PR #2084, opened 2017-03-29
changes for Node Code Generator 2.0

amarzavery:
- supports promises
- supports ES6 syntax
- updated d.ts files according to TS version 2.2.1


balajikris:
/cc: @fiveisprime - This is change to autorest that's required so that when we regenerate the sdk packages, we'll get all ES6 syntax, promises, type definition support etc.

Issue #2072, opened 2017-03-28
Linter complains about missing description in a referenced swagger file


sergey-shandar:
For example

https://travis-ci.org/Azure/azure-rest-api-specs/jobs/214861168 complains about NetworkInterfaceIPConfigurationPropertiesFormat / publicIPAddress property which is defined in 

https://github.com/jobatzil/azure-rest-api-specs/blob/02be03fee65b3378ad3f6da88e5c40aae8d6e4cf/arm-network/2017-03-01/swagger/networkInterface.json#L387

vishrutshah:
Seems like rule should not be fired for the model property with "$refs". 

I'll look into the rule and send PR to fix it.

veronicagg:
@vishrutshah should we not check on the referenced model? I was thinking we needed to provide an appropriate path to the definition (with the file that actually the violation is at).

vishrutshah:
I think "check on the referenced model" would be covered by the same rule when that rule gets fired for that model.

vishrutshah:
As per offline sync, the scenario @veronicagg brought was "If definitions are in separate file then we may not catch that through the validation". That is totally valid scenario for us. We need to think through the scenarios as it is applicable to all the rules.

veronicagg:
@olydis can we get the json path to contain the path to the file where the referenced definition actually is?

olydis:
you as in the consumer side of AutoRest (VSCode extension, messages printed to command line) will absolutely get the correct paths :-) That's what my source map efforts were for - "you" as the validation rule writers may report precisely the jsonpath you see and it'll all work fine!

veronicagg:
@vishrutshah is this still a problem? do we have a way to provide a proper json path in the rule message with the current implementation of autorest or does it get resolved when we move to the typescript entry point of autorest?

vishrutshah:
As per @veronicagg's suggestion, I verified that with the new source mapping we have this fixed. Now the linter points to the right file. Example:

```
gulp autorest --azure-arm --input-file=/Users/vishrut/git-repos/azure-rest-api-specs/arm-network/2017-03-01/swagger/applicationGateway.json --message-format=json

{
  "type": "Warning",
  "code": "DescriptionMissing",
  "message": "'publicIPAddress' model/property lacks 'description' property. Consider adding a 'description' element. Accurate description is essential for maintaining reference documentation.",
  "id": "M4000",
  "validationCategory": "SDKViolation",
  "providerNamespace": null,
[...]

PR #2086, opened 2017-03-29
[Linter] When parameters have refs, linter shuld not complain

vishrutshah:
fixes https://github.com/Azure/autorest/issues/2072

### DIFF
--- a/src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/definition-missing-description.json
+++ b/src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/definition-missing-description.json
@@ -53,6 +53,14 @@
           "description": "description for property2"
         }
       }
+    },
+    "ParamWithRefProperties": {
+      "description": "Description for the param",
+      "properties": {
+        "propertywithref": {
+          "$ref": "#/definitions/ParamWithProperties"
+        }
+      }
     }
   },
   "parameters": {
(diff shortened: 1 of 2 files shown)


olydis:
Can we move forward with this? :-)

olydis:
@veronicagg Ah nice, which issue? I assume it's not https://github.com/Azure/autorest/issues/2072 since that is just talking about source paths lately - not the fact *that* those messages are emitted when they shouldn't. I'm just that curious since I'm currently assigned to a PR where >50% of all linter errors are false positives of this rule (that would be fixed by this PR, but I'm cool with any solution 😄).

veronicagg:
@olydis I was referring to @2072. Could you share a link to the false positives you're referring to above?

olydis:
@veronicagg Sure! :-) https://travis-ci.org/Azure/azure-rest-api-specs/jobs/223017830

olydis:
@vishrutshah @veronicagg @sergey-shandar Can we close this one or do you wanna leave it open due to outstanding discussion or similar?

vishrutshah:
@olydis If the fix cannot address the issue / improve linter rule then i'd keep the issue open and close the PR. Thoughts?

olydis:
@vishrutshah sure thing, I'm fine with any solution. I just went through "old" PRs and wanted to make sure they are still open for a reason :-)

vishrutshah:
@olydis This has been fixed with your new source mapping 💯 Thank You!!!

PR #2087, opened 2017-03-29
Tracked resource validation linter rules

sarangan12:
@dsgouda @vishrutshah @veronicagg @salameer Please review and approve


src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByResourceGroup.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Collections.Generic;
+using AutoRest.Swagger.Model;
+using System.Text.RegularExpressions;
+using AutoRest.Swagger.Validation.Core;
+
+namespace AutoRest.Swagger.Validation
+{
+    public class TrackedResourceListByResourceGroup : TypedRule<Dictionary<string, Schema>>
+    {
+        private readonly Regex listByRgRegEx = new Regex(@".+_ListByResourceGroup$", RegexOptions.IgnoreCase);
```
dsgouda: mark this as a static variable so we dont create it per instance

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListBySubscription.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Collections.Generic;
+using AutoRest.Swagger.Model;
+using System.Text.RegularExpressions;
+using AutoRest.Swagger.Validation.Core;
+
+namespace AutoRest.Swagger.Validation
+{
+    public class TrackedResourceListBySubscription : TypedRule<Dictionary<string, Schema>>
+    {
+        private readonly Regex listBySidRegEx = new Regex(@".+_(List|ListBySubscriptionId|ListBySubscription|ListBySubscriptions)$", RegexOptions.IgnoreCase);
```
dsgouda: mark this as static

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListBySubscription.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            // Retrieve the list of TrackedResources
+            IEnumerable<string> trackedResources = context.TrackedResourceModels;
+
+            // Retrieve the list of getOperations
+            IEnumerable<Operation> getOperations = ValidationUtilities.GetOperationsByRequestMethod("get", context.Root);
+
+            foreach (string trackedResource in trackedResources)
+            {
+                bool listBySubscriptionsCheck = ValidationUtilities.ListByXCheck(getOperations, listBySidRegEx, trackedResource, definitions);
```
dsgouda: Now that I think about it, I think ```ListByXCheck``` should be a part of an abstract class instead of ValidationUtilities, just like we did for OperationNamingValidation.

src/modeler/AutoRest.Swagger/Validation/TrackedResourcePatchOperationValidation.cs
```diff
@@ -38,24 +38,28 @@ public class TrackedResourcePatchOperationValidation : TypedRule<Dictionary<stri
         /// </summary>
         public override Category Severity => Category.Error;
 
-        // Verifies if a tracked resource has a corresponding PATCH operation
```
dsgouda: Don't think you need to do anything (apart from probably adding specifics to the FileObjectPath)

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -43,20 +43,71 @@ class TrackedResourceListByImmediateParent : TypedRule<Dictionary<string, Schema
...
+            KeyValuePair<string, Operation>  operation = operations.First();
+
+            IEnumerable<KeyValuePair<string, OperationResponse>> responses = operation.Value.Responses.Where(resp => resp.Key.Equals("200"));
+            if (responses.Count() == 0) return null;
+            KeyValuePair<string, OperationResponse> response = responses.First();
+
+            return GetReferencedModel(response.Value.Schema.Reference, definitions);
+        }
+
+        private string GetReferencedModel(String schema, Dictionary<string, Schema> definitions)
```
veronicagg: do we have anything in utilities that can do this? I believe we walk the reference tree in other cases, so I'm wondering how/if we can avoid repeating that code. If we don't have something already, should this (and may be some of the other methods here) be moved to ValidationUtilities?

veronicagg: Also, is there an example of this, do we need to walk the reference tree for what you're intending to use it?

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -43,20 +43,71 @@ class TrackedResourceListByImmediateParent : TypedRule<Dictionary<string, Schema
...
         /// </summary>
         /// <param name="childResourcePaths">paths that fits the child resource criteria</param>
         /// <returns>name of the immediate parent resource name</returns>
-        private string GetImmediateParentResourceName(IEnumerable<KeyValuePair<string, Dictionary<string, Operation>>> childResourcePaths)
+        private string GetImmediateParentResourceName(string pathToEvaluate)
         {
-            string pathToEvaluate = childResourcePaths.First().Key;
-            pathToEvaluate = pathToEvaluate.Substring(0, pathToEvaluate.LastIndexOf("s/{"));
+            pathToEvaluate = pathToEvaluate.Substring(0, pathToEvaluate.LastIndexOf("/{"));
+            pathToEvaluate = pathToEvaluate.Substring(0, pathToEvaluate.LastIndexOf("/{"));
```
veronicagg: is this statement repeated twice? (line 109 and 110?)

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -35,6 +35,8 @@ class TrackedResourceListByImmediateParent : TypedRule<Dictionary<string, Dictio
         /// </summary>
         public override Category Severity => Category.Warning;
 
+        private Regex pathRegEx = new Regex("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/[^/]+/[^/]+/{[^/]+}.*/(\\w+)$", RegexOptions.IgnoreCase);
```
veronicagg: can you add a comment on what you're trying to find with regex? It'd be good for someone that reads the code to get an idea of what you're looking to do without having to understand the whole regex, particularly since we're not using a shared one for this case.

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -0,0 +1,135 @@
...
+            return pathToEvaluate.Substring(pathToEvaluate.LastIndexOf("/") + 1);
+        }
+
+        /// <summary>
+        /// Validates if the child tracked resources have List by immediate parent operation.
+        /// </summary>
+        /// <param name="definitions"></param>
+        /// <param name="context"></param>
+        /// <returns></returns>
+        public override bool IsValid(Dictionary<string, Dictionary<string, Operation>> paths, RuleContext context, out object[] formatParameters)
```
veronicagg: what's the json path that gets reported in the output of this rule?

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -265,8 +266,7 @@ public static IEnumerable<string> GetTrackedResources(IEnumerable<string> resour
         /// </summary>
         /// <param name="op">Operation for which to check the x-ms-pageable extension</param>
         /// <returns>true if operation is x-ms-pageable</returns>
-        public static bool IsXmsPageableResponseOperation(Operation op) => (op.Extensions.GetValue<object>(XmsPageable) != null);
-
+        public static bool IsXmsPageableResponseOperation(Operation op) => (op.Extensions != null && op.Extensions.GetValue<object>(XmsPageable) != null);
```

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -406,6 +406,23 @@ public static bool IsReferenceOfType(string reference, Dictionary<string, Schema
...
+                       operation.Responses.Any(
+                           response => response.Key.Equals("200") &&
+                           response.Value.Schema != null && IsArrayOf(response.Value.Schema.Reference, definitionKey, definitions))
+                    );
+        }
+
+        private static bool IsArrayOf(string reference, string referenceToMatch, Dictionary<string, Schema> definitions)
+        {
+            Schema schema = Schema.FindReferencedSchema(reference, definitions);
+            return schema.Properties.Any(property => property.Value.Type == DataType.Array && property.Value.Items != null && property.Value.Items.Reference.EndsWith("/" + referenceToMatch));
```

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByResourceGroup.cs
sarangan12: Done

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListBySubscription.cs
sarangan12: Done

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListBySubscription.cs
sarangan12: I do not see a reason to move it to the abstract class. ValidationUtilities seems to be reasonable place.

src/modeler/AutoRest.Swagger/Validation/TrackedResourcePatchOperationValidation.cs
sarangan12: Yes. No additional op required

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -406,6 +406,23 @@ public static bool IsReferenceOfType(string reference, Dictionary<string, Schema
...
         }
 
+        public static bool ListByXCheck(IEnumerable<Operation> getOperations, Regex regEx, string definitionKey, Dictionary<string, Schema> definitions)
+        {
+            return getOperations.Any(operation =>
+                       regEx.IsMatch(operation.OperationId) &&
+                       IsXmsPageableResponseOperation(operation) &&
+                       operation.Responses.Any(
+                           response => response.Key.Equals("200") &&
+                           response.Value.Schema != null && IsArrayOf(response.Value.Schema.Reference, definitionKey, definitions))
```

src/modeler/AutoRest.Swagger/Validation/DescriptionMissing.cs
```diff
@@ -22,11 +22,9 @@ public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<
         {
             foreach (KeyValuePair<string, Schema> definition in definitions)
             {
-                string key = definition.Key;
-                Schema schema = definition.Value;
-                if (string.IsNullOrWhiteSpace(schema.Description))
+                if (string.IsNullOrWhiteSpace(definition.Value.Description))
                 {
-                    yield return new ValidationMessage(new FileObjectPath(context.File, context.Path), this, string.Format(ModelTypeFormatter, key));
+                    yield return new ValidationMessage(new FileObjectPath(context.File, context.Path), this, string.Format(ModelTypeFormatter, definition.Key));
```
sarangan12: Done

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
sarangan12: I do not see anything in utilities.

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
sarangan12: "file:///C:/workspace/autorest/src/modeler/AutoRest.Swagger.Tests/bin/Debug/netcoreapp1.0/Resource/Swagger/Validation/swagger-list-by-immediate-parent.json#/paths/#/definitions/crs"

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// <summary>
+        /// The severity of this message (ie, debug/info/warning/error/fatal, etc)
+        /// </summary>
+        public override Category Severity => Category.Warning;
+
+        private static Regex pathRegEx = new Regex("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/[^/]+/[^/]+/{[^/]+}.*/(\\w+)$", RegexOptions.IgnoreCase);
+
+        private KeyValuePair<string, string> GetChildResourceName(string path, Dictionary<string, Dictionary<string, Operation>> paths, Dictionary<string, Schema> definitions)
+        {
+            Regex pathRegEx = new Regex("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/[^/]+/[^/]+/{[^/]+}.*/(\\w+)/{[^/]+}$", RegexOptions.IgnoreCase);
```
dsgouda: mark this as a private static readonly member instead

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -0,0 +1,133 @@
...
+            Match match = pathRegEx.Match(path);
+            KeyValuePair<string, string> result = new KeyValuePair<string, string>();
+            if (match.Success)
+            {
+                string childresourcename = match.Groups[1].Value;
```
dsgouda: Use named regex instead so you don't rely on array index

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -0,0 +1,133 @@
...
+            KeyValuePair<string, OperationResponse> response = responses.First();
+
+            return GetReferencedModel(response.Value.Schema.Reference, definitions);
+        }
+
+        private string GetReferencedModel(String schema, Dictionary<string, Schema> definitions)
+        {
+            Schema referencedSchema = Schema.FindReferencedSchema(schema, definitions);
+
+            if (referencedSchema == null) return null;
```
dsgouda: You could do ```if (referencedSchema?.Reference == null) return null;```

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -0,0 +1,133 @@
...
+
+            if (referencedSchema.Reference == null)
+            {
+                IEnumerable<KeyValuePair<string, Schema>> definition = definitions.Where(def => def.Value == referencedSchema);
```
dsgouda: I think you can just do ```return definition.FirstOrDefault()```
(please verify though)

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -0,0 +1,133 @@
...
+        private KeyValuePair<string, string> GetChildResourceName(string path, Dictionary<string, Dictionary<string, Operation>> paths, Dictionary<string, Schema> definitions)
+        {
+            Regex pathRegEx = new Regex("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/[^/]+/[^/]+/{[^/]+}.*/(\\w+)/{[^/]+}$", RegexOptions.IgnoreCase);
+            Match match = pathRegEx.Match(path);
+            KeyValuePair<string, string> result = new KeyValuePair<string, string>();
+            if (match.Success)
+            {
+                string childresourcename = match.Groups[1].Value;
+                string immediateparentresourcename_path = GetImmediateParentResourceName(path);
+                string immediateparentresourcename_actual = GetActualParentResourceName(immediateparentresourcename_path, paths, definitions);
```
dsgouda: We might be digging ourselves a hole here. Why not just use the regex that @vishrutshah had implemented to pick apart resourceTypes and resourceTypeValues? We could help you with it

sarangan12: That was missing some scenarios in this case. Why is this a hole???

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
```diff
@@ -378,21 +378,21 @@ public void SkuModelValidation()
         }
 
         [Fact]
-        public void TrackedResource1Validation()
```
sarangan12: :-) That one is a deleted test case. Nothing to add

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -20,6 +20,7 @@ public static class ValidationUtilities
 
         private static readonly Regex TrackedResRegEx = new Regex(@".+/Resource$", RegexOptions.IgnoreCase);
 
+        // This needs to be deprecated in favor of context.TrackedResources
```

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// This may contain placeholders '{0}' for parameterized messages.
+        /// </remarks>
+        public override string MessageTemplate => Resources.TrackedResourceListByImmediateParentOperationMissing;
+
+        /// <summary>
+        /// The severity of this message (ie, debug/info/warning/error/fatal, etc)
+        /// </summary>
+        public override Category Severity => Category.Warning;
+
+        private static Regex pathRegEx = new Regex("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/[^/]+/[^/]+/{[^/]+}.*/(\\w+)$", RegexOptions.IgnoreCase);
```
sarangan12: Done

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
sarangan12: done

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
sarangan12: The logic is if referenced schema is null, return null. If the reference is null then do something. Else do something else. So, I am not sure the code change will apply. Not changing here

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
sarangan12: It is not handling the count condition. So, no change here

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
sarangan12: Initially, I tried with named regex. unfortunately, that is handing non word characters also as opposed to \w. That was the reason I moved to index. Since there is only one extraction, I do not see the problem here

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -406,6 +386,26 @@ public static bool IsReferenceOfType(string reference, Dictionary<string, Schema
...
+            return getOperations.Any(operation =>
+                       regEx.IsMatch(operation.OperationId) &&
+                       IsXmsPageableResponseOperation(operation) &&
+                       operation.Responses.Any(
+                           response => response.Key.Equals("200") &&
+                           IsArrayOf(response.Value.Schema?.Reference, definitionKey, definitions))
+                    );
+        }
+
+        private static bool IsArrayOf(string reference, string referenceToMatch, Dictionary<string, Schema> definitions)
```
veronicagg: please add comments on the intent/description of this operation

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -406,6 +386,26 @@ public static bool IsReferenceOfType(string reference, Dictionary<string, Schema
             return false;
         }
 
+        public static bool ListByXCheck(IEnumerable<Operation> getOperations, Regex regEx, string definitionKey, Dictionary<string, Schema> definitions)
```
veronicagg: please add comments describing the operation.

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
veronicagg: what scenarios were missing? were there specific for this case?

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
veronicagg: ok, would it make sense to move it there?

src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -262,6 +262,15 @@ Modelers:
...
     <value>Operations API must be implemented for '{0}'.</value>
   </data>
+  <data name="TrackedResourceListByImmediateParentOperationMissing" xml:space="preserve">
+    <value>The child tracked resource, '{0}' with immediate parent '{1}', must have a list by immediate parent operation.</value>
+  </data>
+  <data name="TrackedResourceListByResourceGroupOperationMissing" xml:space="preserve">
+    <value>The tracked resource, '{0}', must have a list by resource group operation.</value>
+  </data>
+  <data name="TrackedResourceListBySubscriptionsOperationMissing" xml:space="preserve">
+    <value>The tracked resource, '{0}', must have a list by subscriptions operation.</value>
```
dsgouda: nit: Do we want to suggest how the operation id should look like?

src/modeler/AutoRest.Swagger/Model/ServiceDefinition.cs
```diff
@@ -102,6 +103,7 @@ public ServiceDefinition()
         [Rule(typeof(ProvidersPathValidation))]
         [CollectionRule(typeof(XmsPathsMustOverloadPaths))]
         [CollectionRule(typeof(BodyPropertiesNamesCamelCase))]
+        [Rule(typeof(TrackedResourceListByImmediateParent))]
```
dsgouda: I have reservations about applying this to custom paths, they overload an existing path, and may return non-resource items, we need to confirm this.

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -20,30 +20,11 @@ public static class ValidationUtilities
 
         private static readonly Regex TrackedResRegEx = new Regex(@".+/Resource$", RegexOptions.IgnoreCase);
 
-        public static readonly Regex ResourcePathPattern = new Regex(@"/providers/(?<providerNamespace>[^{/]+)((/(?<resource>[^{/]+)/)((?<resourceName>[^/]+)))+(/(?<resource>[^{/]+))");
+        // This needs to be deprecated in favor of context.TrackedResources
```
dsgouda: We can take off this comment now :)

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -407,6 +387,41 @@ public static bool IsReferenceOfType(string reference, Dictionary<string, Schema
...
+        /// <param name="referenceToMatch"></param>
+        /// <param name="definitions"></param>
+        /// <returns></returns>
+        private static bool IsArrayOf(string reference, string referenceToMatch, Dictionary<string, Schema> definitions)
+        {
+            if (reference == null)
+                return false;
+
+            Schema schema = Schema.FindReferencedSchema(reference, definitions);
+            return schema.Properties.Any(property => property.Value.Type == DataType.Array && (bool)property.Value.Items?.Reference?.EndsWith("/" + referenceToMatch));
```
dsgouda: Don't typecast to bool, just check for true (```val==true```)

src/modeler/AutoRest.Swagger/Validation/Core/JsonValidationLogListener.cs
```diff
@@ -27,7 +27,7 @@ public void Log(LogMessage message)
                     .FirstOrDefault();
                 var pathComponents = AutoRest.Swagger.Model.Utilities.ValidationUtilities.ResourcePathPattern.Match(path ?? "");
                 var pathComponentProviderNamespace = pathComponents.Groups["providerNamespace"];
-                var pathComponentResourceType = pathComponents.Groups["resourceType"];
+                var pathComponentResourceType = pathComponents.Groups["resource"];
```
dsgouda: @vishrutshah You will definitely need to verify whether these changes affect the code here

src/modeler/AutoRest.Swagger/Validation/Core/JsonValidationLogListener.cs
```diff
@@ -27,7 +27,7 @@ public void Log(LogMessage message)
                     .FirstOrDefault();
                 var pathComponents = AutoRest.Swagger.Model.Utilities.ValidationUtilities.ResourcePathPattern.Match(path ?? "");
                 var pathComponentProviderNamespace = pathComponents.Groups["providerNamespace"];
-                var pathComponentResourceType = pathComponents.Groups["resourceType"];
```
dsgouda: While you're at it you can also consider removing the ```using System.Text.RegularExpressions;``` in this file

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -0,0 +1,118 @@
...
+        /// </summary>
+        public override Category Severity => Category.Warning;
+
+        private KeyValuePair<string, string> GetChildResourceName(string path, Dictionary<string, Dictionary<string, Operation>> paths, Dictionary<string, Schema> definitions)
+        {
+            Match match = ValidationUtilities.ResourcePathPattern.Match(path);
+            KeyValuePair<string, string> result = new KeyValuePair<string, string>();
+            if (match.Success && match.Groups["unparameterizedresource"].Success)
+            {
+                string childResourceName = match.Groups["unparameterizedresource"].Value;
```
dsgouda: Is a child resource always ```unparameterizedresource```?

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -0,0 +1,118 @@
...
+                string immediateParentResourceNameInPath = match.Groups["resource"].Value;
+                string immediateParentResourceNameActual = GetActualParentResourceName(immediateParentResourceNameInPath, paths, definitions);
+                result = new KeyValuePair<string, string>(childResourceName, immediateParentResourceNameActual);
+            }
+
+            return result;
+        }
+
+        private string GetActualParentResourceName(string nameinpath, Dictionary<string, Dictionary<string, Operation>> paths, Dictionary<string, Schema> definitions)
```
dsgouda: nit: variable naming convention

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -0,0 +1,118 @@
...
+        {
+            Regex pathRegEx = new Regex("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/.*/" + nameinpath + "/{[^/]+}$", RegexOptions.IgnoreCase);
+
+            IEnumerable<KeyValuePair<string, Dictionary<string, Operation>>> pths = paths.Where((KeyValuePair<string, Dictionary<string, Operation>> pth) => pathRegEx.IsMatch(pth.Key));
+            if (pths.Count() == 0) return null;
```
dsgouda: Consider using ```.Any()``` instead

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -0,0 +1,118 @@
...
+            }
+
+            return result;
+        }
+
+        private string GetActualParentResourceName(string nameinpath, Dictionary<string, Dictionary<string, Operation>> paths, Dictionary<string, Schema> definitions)
+        {
+            Regex pathRegEx = new Regex("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/.*/" + nameinpath + "/{[^/]+}$", RegexOptions.IgnoreCase);
+
+            IEnumerable<KeyValuePair<string, Dictionary<string, Operation>>> pths = paths.Where((KeyValuePair<string, Dictionary<string, Operation>> pth) => pathRegEx.IsMatch(pth.Key));
```
dsgouda: nit: Rename it to something useful like ```parentPaths/mathingPaths```

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -0,0 +1,118 @@
...
+        public override Category Severity => Category.Warning;
+
+        private KeyValuePair<string, string> GetChildResourceName(string path, Dictionary<string, Dictionary<string, Operation>> paths, Dictionary<string, Schema> definitions)
+        {
+            Match match = ValidationUtilities.ResourcePathPattern.Match(path);
+            KeyValuePair<string, string> result = new KeyValuePair<string, string>();
+            if (match.Success && match.Groups["unparameterizedresource"].Success)
+            {
+                string childResourceName = match.Groups["unparameterizedresource"].Value;
+                string immediateParentResourceNameInPath = match.Groups["resource"].Value;
```
dsgouda: You might want to check one level up if ```unparameterizedresource``` is not found

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -0,0 +1,118 @@
...
+        /// Validates if the child tracked resources have List by immediate parent operation.
+        /// </summary>
+        /// <param name="definitions"></param>
+        /// <param name="context"></param>
+        /// <returns></returns>
+        public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string, Dictionary<string, Operation>> paths, RuleContext context)
+        {
+            foreach(KeyValuePair<string, Dictionary<string, Operation>> pathDefinition in paths)
+            {
+                KeyValuePair<string, string> childResourceMapping = GetChildResourceName(pathDefinition.Key, paths, context.Root.Definitions);
```
dsgouda: nit: ```GetChildResourceName``` could be ```GetChildAndImmediateParentResource```

sarangan12: Done

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
sarangan12: Done

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
sarangan12: Done

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
sarangan12: Done

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
sarangan12: Changed as discussed

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
sarangan12: Changed as discussed

src/modeler/AutoRest.Swagger/Validation/Core/JsonValidationLogListener.cs
sarangan12: Done

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs

src/modeler/AutoRest.Swagger/Model/ServiceDefinition.cs
sarangan12: On second thoughts, I dont think it is needed. Removed it

src/core/AutoRest.Core/Properties/Resources.resx
sarangan12: No. We already have rules insisting on names. I prefer not to take that route

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        private KeyValuePair<string, string> GetChildAndImmediateParentResource(string path, Dictionary<string, Dictionary<string, Operation>> paths, Dictionary<string, Schema> definitions)
+        {
+            Match match = ValidationUtilities.ResourcePathPattern.Match(path);
+            KeyValuePair<string, string> result = new KeyValuePair<string, string>();
+            // Look for unparameterized resource. If it is not available, pick the last resource provided it has a parent resource
+            if (match.Success && (match.Groups["unparameterizedresource"].Success || (match.Groups["resource"].Success && match.Groups["resource"].Captures.Count > 1)))
+            {
+                string childResourceName = match.Groups["unparameterizedresource"].Success? match.Groups["unparameterizedresource"].Value: match.Groups["resource"].Captures[match.Groups["resource"].Captures.Count - 1].Value;
+                string immediateParentResourceNameInPath = match.Groups["resource"].Captures[match.Groups["resource"].Captures.Count - 2].Value;
```
dsgouda: if ```match.Groups["unparameterizedresource"].Success``` you will want to consider the first match for resource group, also check for length for captures or you might run into null exceptions

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -0,0 +1,119 @@
...
+        public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string, Dictionary<string, Operation>> paths, RuleContext context)
+        {
+            foreach(KeyValuePair<string, Dictionary<string, Operation>> pathDefinition in paths)
+            {
+                KeyValuePair<string, string> childResourceMapping = GetChildAndImmediateParentResource(pathDefinition.Key, paths, context.Root.Definitions);
+                if(childResourceMapping.Key != null && childResourceMapping.Value != null)
+                {
+                    string operationIdToFind = childResourceMapping.Key + "_ListBy" + childResourceMapping.Value;
+                    bool listByImmediateParent = paths.Any(filteredPath =>
+                        filteredPath.Value.Where(operationKeyValuePair => operationKeyValuePair.Value.OperationId.Equals(operationIdToFind, StringComparison.CurrentCultureIgnoreCase)).Count() == 1
```
dsgouda: Just do ```filteredPath.Value.Any(```

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -0,0 +1,120 @@
...
+        public override Category Severity => Category.Warning;
+
+        private KeyValuePair<string, string> GetChildAndImmediateParentResource(string path, Dictionary<string, Dictionary<string, Operation>> paths, Dictionary<string, Schema> definitions)
+        {
+            Match match = ValidationUtilities.ResourcePathPattern.Match(path);
+            KeyValuePair<string, string> result = new KeyValuePair<string, string>();
+            // Look for unparameterized resource. If it is not available, pick the last resource provided it has a parent resource
+            if (match.Success && (match.Groups["unparameterizedresource"].Success || (match.Groups["resource"].Success && match.Groups["resource"].Captures.Count > 1)))
+            {
+                int previousResourceLocation = match.Groups["resource"].Captures.Count - 1;
```
dsgouda: nit: you still Have to check if ```previousResourceLocation<=1```, return early

sarangan12: not necessary... i am already checking match.Groups["resource"].Captures.Count > 1

src/modeler/AutoRest.Swagger/Validation/Core/JsonValidationLogListener.cs
veronicagg: are the tests not enough to verify the code? if not, please add any relevant tests.

PR #2101, opened 2017-04-03
fixed 3 validation rules (1x NRE, 2x KeyNotFoundExceptions)

olydis:
(the bug in `TrackedResourcePatchOperationValidation` actually hit me,  `CollectionObjectPropertiesNamingValidation` looks like it has the same bug)

### DIFF
--- a/src/modeler/AutoRest.Swagger/Validation/TrackedResourcePatchOperationValidation.cs
+++ b/src/modeler/AutoRest.Swagger/Validation/TrackedResourcePatchOperationValidation.cs
@@ -1,15 +1,15 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-using AutoRest.Core.Properties;
 using AutoRest.Core.Logging;
-using AutoRest.Swagger.Model.Utilities;
-using System.Collections.Generic;
+using AutoRest.Core.Properties;
+using AutoRest.Core.Utilities;
 using AutoRest.Swagger;
 using AutoRest.Swagger.Model;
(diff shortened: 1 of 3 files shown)


src/modeler/AutoRest.Swagger/Validation/TrackedResourcePatchOperationValidation.cs
```diff
@@ -46,8 +46,8 @@ public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<
             IEnumerable<Operation> patchOperations = ValidationUtilities.GetOperationsByRequestMethod("patch", serviceDefinition);
 
             // enumerate all the models returned by all PATCH operations (200/201 responses)
-            var respModels = patchOperations.Select(op => op.Responses["200"]?.Schema?.Reference?.StripDefinitionPath());
-            respModels.Union(patchOperations.Select(op => op.Responses["201"]?.Schema?.Reference?.StripDefinitionPath())).Where(modelName=>!string.IsNullOrEmpty(modelName));
+            var respModels = patchOperations.Select(op => op.Responses?["200"]?.Schema?.Reference?.StripDefinitionPath());
+            respModels.Union(patchOperations.Select(op => op.Responses?["201"]?.Schema?.Reference?.StripDefinitionPath())).Where(modelName=>!string.IsNullOrEmpty(modelName));
```
dsgouda: How did you find this issue, just curious since we can start running the rest of our validation rules to check for errors likewise

olydis: Just ran it against an actual swagger spec... should have failed for most specs in the repo. As always: Would be best to run rules against the entire repo.

Issue #1951, opened 2017-03-08
New rule: Properties in PATCH operations must be optional and not provide default values


veronicagg:
M2016: For PATCH operation, every property in the model for the request body MUST be optional and the property MUST NOT provide any default value. The absence of default values applies recursively to complex properties in the model definition.
Category: RPC Violation
Severity: Error

dsgouda:
Rule has been added

PR #2113, opened 2017-04-04
Validating PATCH operation request body params

dsgouda:
Addresses issue https://github.com/Azure/autorest/issues/1951


dsgouda:
@azuresdkci Test this please

src/modeler/AutoRest.Swagger/Validation/PatchBodyParametersSchemaValidation.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// </summary>
+        public override ValidationCategory ValidationCategory => ValidationCategory.RPCViolation;
+
+        /// <summary>
+        /// The template message for this Rule. 
+        /// </summary>
+        /// <remarks>
+        /// This may contain placeholders '{0}' for parameterized messages.
+        /// </remarks>
+        public override string MessageTemplate => "PATCH operation '{0}' has reference to model definition '{1}' as request body parameter which has {2} properties";
```
veronicagg: I would write the violation text and then what's violating the rule, something like: "Properties in the request body of a PATCH operation must not be required and must not provide any default values. The absence of default values applies recursively to complex properties in the model definition.  Patch operation: "X". Model definition: "Y". Properties: x, y, z"

dsgouda: It's a bit tricky since I also need to make the message short so that it appears properly on the linter. Let me see what I can do.

Issue #1986, opened 2017-03-10
New rule: For a PUT operation, the response body of 200 should be an Azure Resource


veronicagg:
it's ubermost "allOf" must have x-ms-azure-resource:true.

dsgouda:
Need more information like ValidationCategory, Id

brjohnstmsft:
@veronicagg I assume this rule will not apply to data plane APIs, correct?

veronicagg:
@brjohnstmsft correct, our linter rules are currently all about management plane (ARM), we are ignoring errors on swaggers that are not ARM for now, we'll introduce categories and applicability of rules a little later.

veronicagg:
@dsgouda I'd do RPC Violation for this one, please confirm with Gaurav. Id, please choose next number based on currently implemented ones, and add it to the checklist. thanks!

dsgouda:
Rule added

PR #2123, opened 2017-04-05
Put response resource model validation

dsgouda:
Addressing issue https://github.com/Azure/autorest/issues/1986


src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -393,4 +393,7 @@ Modelers:
   <data name="LongRunningResponseNotValid" xml:space="preserve">
     <value>A '{0}' operation '{1}' with x-ms-long-running-operation extension must have a valid terminal success status code {2}.</value>
   </data>
+  <data name="PutOperationResourceResponseValidationMessage" xml:space="preserve">
+    <value>Put operation '{0}' has response model '{1}' which does not have x-ms-azure-resource extension set to true in its hierarchy</value>
```
veronicagg: Suggestion: "The 200 response model of a PUT operation, should be an Azure resource and have 'x-ms-azure-resource' extension set to true in its hierarchy. Operation: 'X'. Model: 'Y'.".

src/modeler/AutoRest.Swagger/Validation/PutResponseResourceValidation.cs
```diff
@@ -0,0 +1,59 @@
...
+            var ops = ValidationUtilities.GetOperationsByRequestMethod("put", serviceDefinition);
+            var putRespModels = ops.Select(op => (op.Responses?.ContainsKey("200") != true) ? null : op.Responses["200"].Schema?.Reference?.StripDefinitionPath());
+            putRespModels = putRespModels.Where(modelName => !string.IsNullOrEmpty(modelName));
+            var xmsResModels = ValidationUtilities.GetXmsAzureResourceModels(serviceDefinition.Definitions);
+            var violatingModels = putRespModels.Where(respModel => !ValidationUtilities.EnumerateModelHierarchy(respModel, serviceDefinition.Definitions).Intersect(xmsResModels).Any());
+            foreach (var model in violatingModels)
+            {
+                var violatingOp = ops.Where(op => (op.Responses?.ContainsKey("200") == true) && op.Responses["200"].Schema.Reference.StripDefinitionPath() == model).First();
+                var violatingPath = paths.Where(pathObj => pathObj.Value.Values.Where(op => op.OperationId == violatingOp.OperationId).Any()).First();
+                var violatingOpVerb = violatingPath.Value.Keys.Where(key => key.EqualsIgnoreCase("put")).First();
```
veronicagg: we know this is "put" right? if so, why are we looking it up?

dsgouda: Yes, but since we are doing a case-insensitive search, we need to get the exact name of the node which we can then add to the output for linter.

Issue #1958, opened 2017-03-08
New rule or schema update: "name" property should not be an empty string


veronicagg:
M2047: A parameter object MUST have a "name" property defined with a non-empty string as its value.
Category: SDK Violation/Schema violation
Severity: Error

dsgouda:
Fix merged to master

PR #2124, opened 2017-04-05
Adding rule to validate empty parameter names

dsgouda:
Addressing issue https://github.com/Azure/autorest/issues/1958


src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/empty-parameter-name.json
```diff
@@ -0,0 +1,85 @@
+﻿{
+  "swagger": "2.0",
+  "info": {
+    "title": "Checks if the body parameter is named 'parameters'",
```
veronicagg: update title and description?

Issue #2127, opened 2017-04-06
ResourceModelValidation flags false positives


dsgouda:
Validation rule ResourceModelValidation throws an error message even for models which have the "name", "id" and "type" properties marked as readOnly

Eg.: run autorest against spec [here](https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/arm-recoveryservices/2016-06-01/swagger/vaults.json)

FYI @veronicagg @vishrutshah @sarangan12

dsgouda:
The issue is we are enforcing that every model named as ```resource``` Must have the properties 'name', 'id', 'type', 'location' and 'tags'. We have many instances where TrackedResource allOfs on a resource model and then adds the 'location' as a required property. We should relax this rule to check for only 'name', 'id' and 'type' properties.
Any thoughts/suggestions here?

veronicagg:
To add some extra info, we had the following issues opened on this rule: 
https://github.com/Azure/autorest/issues/1936
https://github.com/Azure/autorest/issues/1861 
if appears that something got lost in the discussions, regarding this rule. 
I agree with you, we should be checking whether id, name and type are 'required' properties. I thought we had another rule for this, but can't find it, so if you can't find it either, feel free to turn this one into that. 
Thanks!!

PR #2135, opened 2017-04-07
Addressing issue 2127

dsgouda:
Relaxing validation rule to check for readonly properties 'name', 'id' and 'type' only. Addresses https://github.com/Azure/autorest/issues/2127


src/modeler/AutoRest.Swagger/Validation/ResourceModelValidation.cs
```diff
@@ -44,33 +47,19 @@ public class ResourceModelValidation: TypedRule<Dictionary<string, Schema>>
         /// </summary>
         /// <param name="definitions">Operation Definition to validate</param>
         /// <returns>true if the resource model is valid.false otherwise.</returns>
-        public override bool IsValid(Dictionary<string, Schema> definitions)
+        public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string, Schema> entity, RuleContext context)
         {
-            foreach(string key in definitions.Keys)
+            var resModelList = entity.Where(pair => pair.Key.ToLower().Equals("resource"));
```
veronicagg: should we be applying this to the list of resources already computed for this spec, based on the definition we came up with? why would we apply it just to something named "resource"?

dsgouda: I merely modified what was already implemented. Yes, we can consider our new definition of "resource" here but we will still need to traverse the model hierarchy to get the root model definition which should contain the "name", "id" and "type" properties.

src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -157,21 +157,6 @@ Modelers:
     TypeName: CompositeSwaggerModeler, AutoRest.CompositeSwagger
     </value>
   </data>
-  <data name="CodeGenerationFailed" xml:space="preserve">
```
olydis: NICE cleanup!!! 👍 👍 👍 👍 👍 👍 👍 👍 👍

src/modeler/AutoRest.Swagger/Validation/ResourceModelValidation.cs
```diff
@@ -39,38 +42,59 @@ public class ResourceModelValidation: TypedRule<Dictionary<string, Schema>>
...
-                    if (resourceSchema == null || resourceSchema.Properties.Count == 0)
-                        return false;
-
-                    if (!this.validateSchemaProperty(resourceSchema, "id", true)   ||
-                        !this.validateSchemaProperty(resourceSchema, "name", true) ||
-                        !this.validateSchemaProperty(resourceSchema, "type", true) ||
-                        !this.validateSchemaProperty(resourceSchema, "location", false) ||
-                        !this.validateSchemaProperty(resourceSchema, "tags", false))
-                        return false;
+                    yield return new ValidationMessage(new FileObjectPath(context.File, context.Path.AppendProperty(rootResourceModel)), this, rootResourceModel);
```
olydis: 👍

src/modeler/AutoRest.Swagger/Validation/ResourceModelValidation.cs
veronicagg: @dsgouda do we need to enforce that this is the case? does it matter for this rule? or just the fact that the properties exist in the full set of properties of the model and that they are read-only is enough?

dsgouda:
@azuresdkci Test this please

Issue #1954, opened 2017-03-08
New rule: Every operation should define x-ms-examples


veronicagg:
M2022: Every operation in the swagger specification MUST have an "x-ms-examples" extension applied with scenarios that provide the MINIMUM AND MAXIMUM properties of the request and response payload. Please feel free to provide more example scenarios for a particular operation.
Category: SDK Violation
Severity: Error

dsgouda:
Per offline discussion with @veronicagg will assign severity to Warning and only check for ```x-ms-examples``` extension set on operations

dsgouda:
@kirthik Please confirm severity of this rule

dsgouda:
Also, should we request ```x-ms-examples``` for Operations_list operation?

kirthik:
@dsgouda : This should be an error not warning. I don't think we should request `x-ms-examples` for Operations_List operation since they only take subscription Id and API Version as parameters.

dsgouda:
Validation rule implemented

PR #2139, opened 2017-04-07
Adding validation rule to check for x-ms-examples

dsgouda:
Addresses issue https://github.com/Azure/autorest/issues/1954


src/modeler/AutoRest.Swagger/Validation/XmsExamplesProvidedValidation.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// </summary>
+        /// <remarks>
+        /// This may contain placeholders '{0}' for parameterized messages.
+        /// </remarks>
+        public override string MessageTemplate => "Please provide x-ms-examples describing minimum/maximum property set for response/request payloads for operation '{0}' ";
+
+        /// <summary>
+        /// The severity of this message (ie, debug/info/warning/error/fatal, etc)
+        /// </summary>
+        public override Category Severity => Category.Error;
```
veronicagg: @salameer @kirthik could you confirm if you'd like this rule as an error or a warning?

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
```diff
@@ -555,6 +555,18 @@ public void BodyParametersReferencedValidation()
             var messages = ValidateSwagger(Path.Combine(Core.Utilities.Extensions.CodeBaseDirectory, "Resource", "Swagger", "Validation", "body-parameters-validation-2.json"));
             messages.AssertOnlyValidationMessage(typeof(BodyParametersValidation), 1);
         }
+
+
+        /// <summary>
+        /// Verifies that sku object
```
veronicagg: update comment

src/modeler/AutoRest.Swagger/Validation/XmsExamplesProvidedValidation.cs
kirthik: @salameer  and I discussed this and this should be an error.

olydis:
@dsgouda ...but fix the failing tests ;)

veronicagg:
After talking to Samer, as this will cause lots of messages, we'd like to do the following: if there are more than 10 issues, show only 1 message that indicates that "x-ms-examples" are missing for more than 10 operations, and point to "paths" as the path of the issue. If there are less than 10 errors, then show the appropriate messages, pointing to each of the corresponding operations/places where the rule is violated. 
If this gets too complicated to implement, let's chat. thanks!

src/modeler/AutoRest.Swagger/Validation/XmsExamplesProvidedValidation.cs
salameer: Yep it should be an error.

One possiblity that veronica and I just discussed is:

- If we have less that ten (or another threshold) we should show all the errors to help the user track.
- if we have more than the threshold it should be (one Generic error + 10 errors to help the user navigate)

if this is too complicated we can show just one  generic (but the user wouldn't have swiggles)

let me know

Issue #1955, opened 2017-03-08
New rule: list operations must apply x-ms-pageable with nextLinkName:null


veronicagg:
M2024: Operations that return a single page and have the response schema {"value": [Array of items] }, MUST apply the "x-ms-pageable" with it's "nextLinkName" property explicitly set to "null". This will ensure that value property gets unwrapped into an IEnumerable.
Category: SDK Violation
Severity: Error

jaredmoo:
@veronicagg when I add `x-ms-pageable` to list operations API (https://github.com/Azure/azure-rest-api-specs/blob/master/arm-sql/2014-04-01/swagger/sql.core.json#L1434), I get this error.

```
ERROR: Since operation 'ListOperations' response has model definition 'x-ms-pageable', it should be named as "list_*"
        Path: file:///E:/dsmain1/retail/amd64/Xdb/Manifest/Svc/ManagementService.WebApiHosting.ARM.Tools/swagger/swagger.json#/paths
```

Azure services aren't consistent on what the operationid for GET /operations should be. Some have `Operations_List`, some have `ListOperations` so the method is on the client object. SQL is in the latter camp. What's your guidance here? If latter option is ok, then we need to suppress that linter error.

jaredmoo:
Also if one or the other operationIds is recommended, that should be a new linter rule. :)

PR #2145, opened 2017-04-10
Updating message to suggest nextLinkName options

dsgouda:
Addressing issue https://github.com/Azure/autorest/issues/1955
I have attempted to address the issue by improving the message on validation rule that flags absence of a ```nextLinkName``` field in an ```x-ms-pageable``` operation


src/modeler/AutoRest.Swagger/Validation/Extensions/NextLinkPropertyMustExist.cs
```diff
@@ -54,7 +54,7 @@ public override bool IsValid(object pageable, RuleContext context, out object[]
         /// <remarks>
         /// This may contain placeholders '{0}' for parameterized messages.
         /// </remarks>
-        public override string MessageTemplate => "The property '{0}' specified by nextLinkName does not exist in the 200 response schema.";
+        public override string MessageTemplate => "The property '{0}' specified by nextLinkName does not exist in the 200 response schema. \nFor the method to return an IEnumerable, set nextLinkName to null. \nFor a method to return a chained object, set nextLinkName to appropriate field in the list object.";
```
olydis: > For the method to return an IEnumerable

Careful with `IEnumerable`:
- this is Swagger validation, not C#
- people may think *"oh I want the signature to return IEnumerable! :)"* although they do have multiple pages. They may *reasonably* think that all pages are concat'ed into a single enumerable - something we should do anyways, so this message may even become deprecated.

I'd go with something like `If the API returns a single page`/`If the API does not return a chain of ...`

veronicagg: should we take a text extract from https://github.com/Azure/autorest/blob/master/docs/extensions/readme.md#x-ms-pageable ? Maybe something like: "The property '{0}' specified by nextLinkName does not exist in the 200 response schema. \nPlease, specify the name of the property that provides the nextLink. If the model does not have the nextLink property then specify null"
Even though this does not include the full explanation, we could have the rest in the documentation.

Also, does the rule need any update to allow a "null" value for this property?

dsgouda:
@veronicagg If the nextLinkName is explicitly set to null in the extension, it wouldn't violate this rule.

Issue #1952, opened 2017-03-08
New rule: Same model for body parameter and response in PUT


veronicagg:
M2017: The model definition for the body parameter and the response MUST be the same for a PUT operation. This ensures that the same entity will be reusable between GET and PUT.
Category: SDK Violation
Severity: Error

dsgouda:
I'm concerned whether this will upset the usual structure most RPs use for PUT operations i.e., using a "ModelCreateParameters" model as request body schema and "Model" as their response model schema. Do we still want this as an error?

PR #2147, opened 2017-04-11
Validating PUT operation request/response schemas

dsgouda:
Addressing issue https://github.com/Azure/autorest/issues/1952


Samples/2a-arm-validation/shell/stderr.txt
```diff
@@ -6,6 +6,8 @@ ERROR (ListByOperationsValidation/M1003/SDKViolation): Operation Usage_List must
         Path: https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/arm-storage/2015-06-15/swagger/storage.json#$.paths["/subscriptions/{subscriptionId}/providers/Microsoft.Storage/usages"].get.operationId
 ERROR (OperationsAPIImplementationValidation/M3023/RPCViolation): Operations API must be implemented for '/providers/Microsoft.Storage/operations'.
         Path: https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/arm-storage/2015-06-15/swagger/storage.json#$.paths
+ERROR (PutRequestResponseValidation/M2017/SDKViolation): "A PUT operation request body schema must be the same as its 200 response schema. Operation: StorageAccounts_Create"
```
olydis: Just a thought, scrolling through these messages: Maybe it makes sense to write what the schemas involved are (something like `A PUT operation request body schema (found 'ABC') must be the same as its 200 response schema (found 'XYZ'). Operation: StorageAccounts_Create`).
That way we know what we're dealing with even without looking at the Swagger :-)

dsgouda: Feedback much appreciated!

veronicagg: what is this sample file for? I haven't seen it before.

olydis: @veronicagg I added new (more real life) examples and modified CI a bit: stdout, stderr and exitcode of the examples is recorded as well - both for people to see what, say, "activating validation" or "supressing" *really* means and for us to detect unwanted changes in our API (AutoRest has *way* too little end-to-end tests) :-)

olydis: so in this very case, it already makes us think "wait, does this new validation rule really apply to this example?"

dsgouda:
@olydis @veronicagg Is this better?
As a side note I'm curious whether this will upset the usual structure most RPs use for PUT operations i.e., using a "ModelCreateParameters" model as request body schema and "Model" as their response model schema. Do we still want this as an error?

olydis:
@dsgouda I share that concern :-D

dsgouda:
@ravbhatnagar @amarzavery Definitely need you rinputs here!

src/autorest-core/test/directive.ts
```diff
@@ -33,7 +33,7 @@ import { Message, Channel } from "../lib/message";
     // muted run
     await autoRest.ResetConfiguration();
     await autoRest.AddConfiguration({ "azure-arm": true });
-    await autoRest.AddConfiguration({ directive: { suppress: ["AvoidNestedProperties", "ModelTypeIncomplete", "DescriptionMissing"] } });
+    await autoRest.AddConfiguration({ directive: { suppress: ["AvoidNestedProperties", "ModelTypeIncomplete", "DescriptionMissing", "PutRequestResponseValidation"] } });
```
veronicagg: what is this "add configuration" directive about?

dsgouda: Adds a suppression so we can enforce the assertions for no warnings are found in the document is true.

Issue #1950, opened 2017-03-08
New rule: Noun in operationId must be different from Noun in Model definition


veronicagg:
M1002: The NOUN in the OperationId "NOUN_VERB" MUST not match the name of the Model definition, to avoid a collision in the namespace, use the plural form of NOUN in the operationId.
Category: SDK Violation
Severity: Error

dsgouda:
This is going to be rather difficult to implement. Even if we validate that the "NOUN" in "NOUN_VERB" does not match with any of the defined models, we won't be able to clearly suggest a plural form for the model (the classic ```Factory``` and ```Factories``` example).
The best we can do is throw a more generic error that simply states that the NOUN provided conflicts with one of the model definitions, and the author should therefore modify it.
Tagging @sarangan12 since he ran into a similar issue with one of the rules.

veronicagg:
@dsgouda I agree, let's not try to get smart suggesting a specific name, let's just point out the conflict and suggest in the error text, that they could use the plural form of the NOUN to avoid the conflict.

matthchr:
@vishrutshah suggested I comment here.

The issue with the check right now is that you assume that there is actually a conflict when there may not be. For example this operation:
`Certificate_Get` and an entity `Certificate` do not actually collide in most languages.

In C# you get `CertificateOperations.Get` which doesn't collide with the model `Certificate`.
In Python you get `CertificateOperations.Get` which doesn't collide with the model `Certificate`
In Java you get `CertificateOperations.Get` which doesn't collide with the model `Certificate`.

I think you get the point =)

IMO this rule is too generator specific and you're making an assumption about the structure of the generated code which isn't even true. The only language that I know of where this actually conflicts is Node, and (IMO) the fix isn't to flag errors in the generator the fix is to make nodes type `CertificateOperations` instead of `Certificate` like it is now.

This avoids the awkward "add an s" (since then it's `CertificatesOperations` everywhere which reads pretty awkwardly), _and_ it also is just more correct to boot (`CertificateOperations` is more clear about what the class is actually doing than `Certificates`).

vishrutshah:
@dsgouda & @veronicagg Feel free to throw in your thoughts here...

dsgouda:
@matthchr ```CertificateOperations``` is actually a name resolved by AutoRest when it sees conflicting names, for eg. if you were to name model and operation ids as lets say ```Cert``` and ```Certificate_get```, we would see ```Certificate.cs``` with all the operations.
[...]

PR #2150, opened 2017-04-12
Validation rule to check if operationId nouns conflict with model names

dsgouda:
Addressing issue https://github.com/Azure/autorest/issues/1950


src/modeler/AutoRest.Swagger/Validation/OperationIdNounPluralizedFormValidation.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Linq;
+using System.Text.RegularExpressions;
+using AutoRest.Core.Logging;
+using AutoRest.Core.Properties;
+using AutoRest.Swagger.Model;
+using AutoRest.Swagger.Validation.Core;
+
+namespace AutoRest.Swagger.Validation
+{
+    public class OperationIdNounPluralizedFormValidation : TypedRule<string>
```
veronicagg: I'm not sure about the naming of this class, should be more like "OperationIdNounConflictWithModel" ? What the rule is trying to enforce is not about pluralizing the noun, it's about not conflicting with another name being used for another class name (class name of a model vs. class name of the one containing the operations).

src/modeler/AutoRest.Swagger/Validation/OperationIdNounPluralizedFormValidation.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// </summary>
+        /// <remarks>
+        /// This may contain placeholders '{0}' for parameterized messages.
+        /// </remarks>
+        public override string MessageTemplate => Resources.OperationIdNounPluralizedMessage;
+
+        /// <summary>
+        /// The severity of this message (ie, debug/info/warning/error/fatal, etc)
+        /// </summary>
+        public override Category Severity => Category.Error;
```
veronicagg: Let's check on this being an Error or not, as I was looking into this rule, I see that AutoRest will disambiguate on its own, calling the class that contains the operations "<Noun>Operations", and keeping the "<Noun>" as the name of the class of the model. Since this doesn't cause a code generation problem, but more of a "better name" for customers, it should be a warning in my opinion. Thoughts? @salameer @kirthik @amarzavery

src/modeler/AutoRest.Swagger/Validation/OperationIdNounPluralizedFormValidation.cs
```diff
@@ -0,0 +1,75 @@
...
+            return true;
+        }
+
+        /// <summary>
+        /// The template message for this Rule. 
+        /// </summary>
+        /// <remarks>
+        /// This may contain placeholders '{0}' for parameterized messages.
+        /// </remarks>
+        public override string MessageTemplate => Resources.OperationIdNounPluralizedMessage;
```
veronicagg: if you update the name of the class, please update this message name also.

src/modeler/AutoRest.Swagger/Validation/OperationIdNounPluralizedFormValidation.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// </summary>
+        public override string Id => "M2063";
+
+        /// <summary>
+        /// Violation category of the Rule.
+        /// </summary>
+        public override ValidationCategory ValidationCategory => ValidationCategory.SDKViolation;
+
+        /// <summary>
+        /// This rule passes if the operation id doesn't contain a repeated value before and after the underscore
```
veronicagg: this doesn't look like the right comment.

src/modeler/AutoRest.Swagger/Validation/OperationIdNounPluralizedFormValidation.cs
dsgouda: @olydis thoughts?

src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -349,4 +349,7 @@ Modelers:
   <data name="ParametersPropertiesValidation" xml:space="preserve">
     <value>Parameter Must have the "name" property defined with non-empty string as its value</value>
   </data>
+  <data name="OperationIdNounConflictingModelNamesMessage" xml:space="preserve">
+    <value>OperationId has a noun that conflicts with one of the model names in definitions section. Consider using the pluralized form for '{0}'</value>
```
veronicagg: should it be "Consider using the plural form of..." instead of "pluralized ? :)

src/modeler/AutoRest.Swagger/Validation/OperationIdNounPluralizedFormValidation.cs
olydis: Agree with @veronicagg. Of course the decision makers might have thought of this and just love our customers very much: Indeed those suffixes are annoying - not only because they are long, but they also sometimes appear/disappear during tiny Swagger revisions (conflicting model definitions was introduced, existing model definition was renamed, etc...), which unnecessarily breaks existing code.

src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/operationid-pluralized-noun-validation.json
```diff
@@ -0,0 +1,57 @@
+﻿{
+  "swagger": "2.0",
```
veronicagg: you may want to update the name of the .json file.

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
```diff
@@ -592,6 +592,13 @@ public void EmptyParameterNameValidation()
             var messages = ValidateSwagger(Path.Combine(Core.Utilities.Extensions.CodeBaseDirectory, "Resource", "Swagger", "Validation", "empty-parameter-name.json"));
             messages.AssertOnlyValidationMessage(typeof(ParameterNameValidation), 2);
         }
+
+        [Fact]
+        public void OperationIdNounPluralizedValidationTest()
```
veronicagg: could you update the name of the method here?

src/modeler/AutoRest.Swagger/Validation/OperationIdNounPluralizedFormValidation.cs
```diff
@@ -0,0 +1,72 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
```
veronicagg: could you update the name of the .cs file?

src/modeler/AutoRest.Swagger/Validation/OperationIdNounPluralizedFormValidation.cs
veronicagg: @dsgouda maybe my comment wasn't clear enough, I would turn this into a warning at the moment, since it doesn't break code generation. You could also add in the message, if the name is not update, generated code will add a string like "Operations". to disambiguate.

PR #2152, opened 2017-04-12
Updating validation rule BodyTopLevelProperties to reuse resource definition

dsgouda:
This has already been implemented correctly under ResourceModelValidation


veronicagg:
@dsgouda per our conversation, the ResourceModelValidation rule is not doing the same thing as BodyTopLevelProperties, so we should keep both. I agree with you on that BodyTopLevelProperties, should be updated to use the new "shared" Resource definition - please open an issue on this if there isn't one.

src/modeler/AutoRest.Swagger/Validation/BodyTopLevelProperties.cs
```diff
@@ -28,62 +28,25 @@ public class BodyTopLevelProperties : TypedRule<Dictionary<string, Operation>>
...
-                                    }
-                                }
-                            }
-                          }  
-                        }
-                    }
-                }
-            formatParameters = new[] { string.Join(", ", notAllowedProperties.ToArray()) };
-            return (notAllowedProperties.Count() == 0);
+                yield return new ValidationMessage(new FileObjectPath(context.File, context.Path.AppendProperty(violatingModel).AppendProperty("properties")), this,
```
veronicagg: is the path to the object missing the actual property violating the rule? could you paste an example output?

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
```diff
@@ -230,7 +230,7 @@ public void CollectionObjectsPropertiesNamingValidation()
         public void BodyTopLevelPropertiesValidation()
         {
             var messages = ValidateSwagger(Path.Combine(Core.Utilities.Extensions.CodeBaseDirectory, "Resource", "Swagger", "Validation", "body-top-level-properties.json"));
-            messages.AssertOnlyValidationMessage(typeof(BodyTopLevelProperties), 2);
+            messages.AssertOnlyValidationMessage(typeof(BodyTopLevelProperties), 1);
```
veronicagg: did the count go down because we're no flagging the inline body parameter issue? If so, I'm ok, because we'd be violating another rule that requests not to have inline body params defined, but we may want to remove that from the spec used for test (as it may confuse us later). It'd be nice to add a comment in the rule itself, that inline body parameters are not being flagged, because we're flagging them as not permitted by another rule.

src/modeler/AutoRest.Swagger/Validation/BodyTopLevelProperties.cs
dsgouda: Since we are returning array of violating properties in the error message, the violating path is the beginning of the properties section of the model definition. If we end up returning a violation message per property we might be able to return the exact path to violating property but I feel that would be a lot of noise.

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
dsgouda: The count went down because we are using a different definition of a resource, the earlier definition considered models named ```Resource``` as resource models too but in the new definition we disregard models that are named as ```Resource```, ```TrackedResource``` or ```ProxyResource```

veronicagg: was Resource flagged before? I didn't think it was violating the rule.

dsgouda: Yes, it returned the 'Resource' as a resource model because it simply checked for the model named 'Resource' and all models that allOfed on it.

DheerendraRathor:
@dsgouda Prior to this change, there was a case insensitive validation for top level properties
```
if (!allowedTopLevelProperties.Contains(prop.Key.ToLower()))
 -                                        {
 -                                            notAllowedProperties.Add(defName + "/" + prop.Key);
 -                                        }
```

Is there any ARM guidelines behind making it case sensitive?

dsgouda:
I might have missed that when Linq-ifying that code. Will open a PR for this fix soon.

dsgouda:
In hindsight, if you put properties that are not in lower case, it would not be the recommended camelCase pattern, regardless, I have opened a PR with a fix [here](https://github.com/Azure/azure-openapi-validator/pull/85)

Issue #2128, opened 2017-04-06
Potential Issue with ServiceDefinitionParameters validation rule


dsgouda:
For ARM specs we enforce parameters ```subscriptionId``` and ```api-version``` are defined globally by a case insensitive search for all globally defined parameters. However, specs often have these defined as ```subscriptionIdParameter``` or ```ApiVersionParameter```. We need to either relax the matching method to use ```StartsWith``` instead of ```Equals``` or explicitly search for the two variations as well.

@veronicagg @vishrutshah @sarangan12 thoughts?

veronicagg:
the rule should be checking for the "name" of the property "api-version"  and not for the name of the model "ApiVersionParameter". As far as I know, autorest depends on this property "name" being "api-version"
  "parameters": {
    "ApiVersionParameter": {
      "name": "api-version",
      "in": "query",
      "required": true,
      "type": "string",
      "description": "Client Api Version."
    }

vishrutshah:
Example seems like https://travis-ci.org/Azure/azure-rest-api-specs/jobs/226238975#L1457

dsgouda:
Ran the latest version of validator for the example mentioned here (and another test spec) and not able to repro the issue. Please reopen if this persists.

PR #2157, opened 2017-04-13
Remove api-version validation

sarangan12:
@vishrutshah @veronicagg @salameer @dsgouda  Could you please review and approve?

### DIFF
--- a/src/modeler/AutoRest.Swagger/Validation/OperationParametersValidation.cs
+++ b/src/modeler/AutoRest.Swagger/Validation/OperationParametersValidation.cs
@@ -11,7 +11,6 @@ namespace AutoRest.Swagger.Validation
     public class OperationParametersValidation : TypedRule<SwaggerParameter>
     {
         private const string SubscriptionId = "subscriptionid";
-        private const string ApiVersion = "api-version";
 
         /// <summary>
         /// Id of the Rule.
@@ -30,7 +29,7 @@ namespace AutoRest.Swagger.Validation
         /// <param name="paths"></param>
         /// <returns></returns>
         public override bool IsValid(SwaggerParameter Parameter) => 
-           (!string.IsNullOrEmpty(Parameter.Reference) ||Parameter?.Schema != null || !(Parameter?.Name?.ToLower().Equals(SubscriptionId) == true || Parameter?.Name?.ToLower().Equals(ApiVersion) == true));
+           (!string.IsNullOrEmpty(Parameter.Reference) || Parameter.Schema != null || Parameter.Name?.ToLower().Equals(SubscriptionId) == false);
         
         /// <summary>
         /// The template message for this Rule. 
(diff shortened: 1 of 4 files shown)


src/modeler/AutoRest.Swagger/Validation/OperationParametersValidation.cs
```diff
@@ -30,7 +29,7 @@ public class OperationParametersValidation : TypedRule<SwaggerParameter>
         /// <param name="paths"></param>
         /// <returns></returns>
         public override bool IsValid(SwaggerParameter Parameter) => 
-           (!string.IsNullOrEmpty(Parameter.Reference) ||Parameter?.Schema != null || !(Parameter?.Name?.ToLower().Equals(SubscriptionId) == true || Parameter?.Name?.ToLower().Equals(ApiVersion) == true));
```
dsgouda: You'll have to update the validation message and regenerate the code

src/modeler/AutoRest.Swagger/Validation/OperationParametersValidation.cs
```diff
@@ -11,7 +11,6 @@ namespace AutoRest.Swagger.Validation
     public class OperationParametersValidation : TypedRule<SwaggerParameter>
     {
         private const string SubscriptionId = "subscriptionid";
-        private const string ApiVersion = "api-version";
```
dsgouda: Just add a to-do saying we need to add api-version in the check some time in the future when we enable the single-swagger spec mode for validation.

veronicagg:
@sarangan12 @dsgouda Why are we removing this? could you please provide some context in the PR on this?

dsgouda:
When composite

src/modeler/AutoRest.Swagger/Validation/OperationParametersValidation.cs
sarangan12: Updated the message. What do you mean by regenerate the code?

dsgouda:
Just run ```gulp build``` and ```gulp regenerate``` and update diffs if any.

sarangan12:
@veronicagg 

The basic idea of the rule is that the subscription id and api version should not be defined inline and must be defined globally. For subscription id, it is not an issue. But, for the api-version it is. Why?

Autorest core treats api-version as special case. It automatically picks the definition of the api-version and substitutes it in the servicedefinition. So, at the time of execution of the validation, we will not know whether the api-version is defined inline or globally. This creates a lot of noise. So, removing this

sarangan12:
@dsgouda regenerate did not add anything new.

dsgouda:
While we are at it, do you think you could resolve [this](https://github.com/Azure/autorest/issues/2128) too, should be a very minimal change

sarangan12:
@dsgouda That issue is not for this rule. It is for the ServiceDefinitionParameters. I am taking it up as next task. Not merging with this

veronicagg:
@sarangan12 thanks for the explanation! based on our conversation, let's open an issue to see how we could validate api-version being the same across a swagger file, since that's one issue we would be missing by removing this check. It's basically that rule should apply to each individual file and not a composite, which is related to what @dsgouda and I have discussed as a feature to add to our validation .(https://github.com/Azure/autorest/issues/1940)

sarangan12:
@veronicagg Sent a note to @kirthik

PR #2160, opened 2017-04-14
Fixing bug with ResourceModelValidation, added additional tests

dsgouda:
Fixes issue with ResourceModelValidation rule


src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/positive/valid-resource-model-readonly-props.json
```diff
@@ -0,0 +1,104 @@
...
+        },
+        "name": {
+          "readOnly": true,
+          "type": "string",
+          "description": "Resource type"
+        }
+      },
+      "x-ms-azure-resource": true
+    },
+    "MyResource": {
```
veronicagg: I think we should make this swagger more compliant/valid, by referencing this model in some operation above.

PR #2163, opened 2017-04-15
[Linter] Message update for BodyParametersValidation

veronicagg:


sarangan12:
@veronicagg Can you close this since I am goin g to remove this rule altogether?

PR #2165, opened 2017-04-17
Validation Rule Categorization

olydis:
/cc @dsgouda

Axis:

individual vs merged (can only be one of 'em)
category (can be in multiple)
Most validation rules should probably be classified to work on the merged OpenAPI definition.

### DIFF
--- a/src/modeler/AutoRest.Swagger/Model/ServiceDefinitionMetadata.cs
+++ b/src/modeler/AutoRest.Swagger/Model/ServiceDefinitionMetadata.cs
@@ -1,0 +1,39 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using AutoRest.Core.Utilities;
+using System;
+using System.Collections.Generic;
+
+namespace AutoRest.Swagger.Model
+{
+    /// <summary>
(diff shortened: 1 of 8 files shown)


src/modeler/AutoRest.Swagger/Model/ServiceDefinitionMetadata.cs
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    /// <summary>
+    /// Represents categories for OpenAPI definitions that are associated with special treatment with regards to validation.
+    /// </summary>
+    [Flags]
+    public enum ServiceDefinitionCategory
+    {
+        Regular = 0,
+        ARM = 1 << 0
```
dsgouda: Out of curiosity, why the bit operation?

olydis: I like to do this for `Flags` style enums: For further categories, this will look like
```
... = 1 << 0,
... = 1 << 1,
... = 1 << 2,
... = 1 << 3,
... = 1 << 4,
... = 1 << 5
```
which I find just slightly more readable (and maybe less error prone if, say, someone decides to remove a category from the middle or something).

Kinda makes it clear that the categories actually *are* represented by bits ;-)

olydis:
Closing as @dsgouda has created a new PR based on this branch.

PR #2177, opened 2017-04-18
Adding documentation for BooleanPropertyNotRecommended rule

veronicagg:
Adding documentation for rule. Please provide your feedback, so we could use this as example to add documentation for the rest of the rules.
There are some existing file in the same folder where I'm placing this. Those files maybe out of date, as they were previously written, we can review and update those if we agree on a format and type of content to include.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,6 @@ compare-results/*
 *.user
 *.sln.docstates
 .vs
-developer/
 launchSettings.json
 
 # Build results
--- a/docs/developer/validation-rules/boolean-property.md
+++ b/docs/developer/validation-rules/boolean-property.md
@@ -1,4 +1,4 @@
-# BooleanPropertyNotRecommended
+# M3018 - BooleanPropertyNotRecommended
 ## Description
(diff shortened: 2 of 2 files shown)


docs/developer/validation-rules/boolean-property.md
```diff
@@ -0,0 +1,13 @@
+# BooleanPropertyNotRecommended
```
salameer: Lets include the ID of the rules in the MD file name

.gitignore
```diff
@@ -10,7 +10,6 @@ compare-results/*
 *.user
 *.sln.docstates
 .vs
-developer/
```
dsgouda: nit: expected diff?

veronicagg: yup, so, it looks like this makes the "developer/" folder not be tracked, and since these doc files are in there, I remove it. I don't think we want to ignore these files, do we? Maybe this was leftover from previous autorest structure? feel free to check with some other autorest folks on this as needed, as I don't know why we could want to ignore things under "developer".

veronicagg:
@dsgouda the specs repo will end up pointing to this docs as needed.

PR #2180, opened 2017-04-19
[Go] Paging using channels

mcardosos:
PTAL
@jhendrixMSFT @marstr

I don't like what I wrote for the Lazy operation documentation :/ There must be a better way to write it.


src/generator/AutoRest.Go/Templates/MethodTemplate.cshtml
```diff
@@ -193,3 +196,70 @@ func (client @(Model.Owner)) @(Model.ResponderMethodName)(resp *http.Response) (
...
 @:}
 }
+
+@:@EmptyLine
+@:// @(Model.LazyMethodName) gets all results from the list and puts them inside a channel.
+@:func (client @(Model.Owner)) @(Model.LazyMethodName) @(Model.MethodParametersSignatureLazy) @(Model.MethodReturnSignatureLazy) {
+    @:respChan := make(chan @((Model.ListElement.ModelType as SequenceTypeGo).GetElement), 1)
+    @:errChan := make(chan error, 1)
+
+    @:go func() {
```
marstr: I think in general, we can probably reduce the amount of complexity in this method. I'll chat about this with you offline.

src/generator/AutoRest.Go/Model/MethodGo.cs
```diff
@@ -102,6 +103,19 @@ public string MethodParametersSignature
...
 
+    public string MethodParametersSignatureLazy
+    {
+      get
+      {     
+        var signature = new StringBuilder("(");
+        signature.Append(MethodParametersSignature);
+        signature.Append(MethodParametersSignature.Length > 0
+          ? ", stop <-chan struct{})"
+          : "stop <-chan struct{})");
```
mcardosos: I don't want it to collide with the cancel channel used in long running operations.

src/generator/AutoRest.Go/Model/MethodGo.cs
```diff
@@ -112,8 +126,37 @@ public string MethodReturnSignature
...
+            return p;
+          }
+        }
+        return null;
+      }
+    }
+
     public string NextMethodName => $"{Name}NextResults";
 
+    public string LazyMethodName => $"{Name}Lazy";
```
mcardosos: There are already some `ListAll` funcs  in the swaggers (they are for list by subscription instead than by resource group). To not erpeat funcs names, I chose `ListComplete`, but we can discuss which is a better name.

src/generator/AutoRest.Go/Model/MethodGo.cs
```diff
@@ -365,15 +408,30 @@ public Response ReturnValue()
...
     public bool NextMethodExists(IEnumerable<MethodGo> methods)
     {
-      if (Extensions.ContainsKey(AzureExtensions.PageableExtension))
+      string next = NextOperationName;
+      if (!string.IsNullOrEmpty(next))
+      {
+          return methods.Any(m => m.Name.Value.EqualsIgnoreCase(next));
+      }
+      return false;
+      // return string.IsNullOrEmpty(NextMethodName);
```
mcardosos: fixed

mcardosos:
https://github.com/Azure/autorest/pull/2186
I have to remove stuff! :D

jhendrixMSFT:
Do you have some sample codegen that we can review?

mcardosos:
https://github.com/mcardosos/azure-sdk-for-go/blob/paging/arm/graphrbac/applications.go#L303

@jhendrixMSFT There it is.
Right now I can gen like this. I want to gen `Next` funcs that take the list result as receiver too (WIP).

mcardosos:
We happily talked about not including the `model.Next()` thing here, I have updated the PR to only include the `GetTheUniverse()` func. I also added it to all paging acceptance tests.
PTAL @marstr @jhendrixMSFT

jhendrixMSFT:
Do you have an updated sample of the generated code?

mcardosos:
Generated code for the acceptance tests looks like this...
``` go

// GetMultiplePagesComplete gets all elements from the list without paging.
func (client PagingClient) GetMultiplePagesComplete(clientRequestID string, maxresults *int32, timeout *int32, cancel <-chan struct{}) (<-chan Product, <-chan error) {
	resultChan := make(chan Product)
	errChan := make(chan error, 1)
	go func() {
		var err error
		defer func() {
			errChan <- err
			close(resultChan)
			close(errChan)
		}()
		list, err := client.GetMultiplePages(clientRequestID, maxresults, timeout)
		if err != nil {
			return
		}
		if list.Values != nil {
			for _, item := range *list.Values {
				select {
				case <-cancel:
					return
				default:
					resultChan <- item
				}
			}
		}
		for list.NextLink != nil {
			list, err = client.GetMultiplePagesNextResults(list)
			if err != nil {
				return
			}
			if list.Values != nil {
				for _, item := range *list.Values {
					select {
					case <-cancel:
						return
					default:
						resultChan <- item
					}
				}
			}
		}
	}()
	return resultChan, errChan
}
```

src/generator/AutoRest.Go/Templates/MethodTemplate.cshtml
```diff
@@ -229,3 +211,59 @@ func (client @(Model.Owner)) @(Model.ResponderMethodName)(resp *http.Response) (
...
+{
+@:@EmptyLine
+@:// @(Model.ListCompleteMethodName) gets all elements from the list without paging.
+@:func (client @(Model.Owner)) @(Model.ListCompleteMethodName) @(Model.MethodParametersSignatureComplete) @(Model.MethodReturnSignatureComplete) {
+    @:resultChan := make(chan @((Model.ListElement.ModelType as SequenceTypeGo).GetElement))
+    @:errChan := make(chan error, 1)
+    @:go func() {
+        @:var err error
+        @:defer func() {
+            @:errChan <- err
```
jhendrixMSFT: I think you want to change this to write to errChan only if err is not nil; this will allow the select on channels pattern to work properly (I made the same change in the generator for LROs).

mcardosos: Fixed

src/generator/AutoRest.Go/Templates/MethodTemplate.cshtml
```diff
@@ -229,3 +211,59 @@ func (client @(Model.Owner)) @(Model.ResponderMethodName)(resp *http.Response) (
...
+            @:return
+        @:}
+        @:if list.@(Model.ListElement.Name) != nil {
+            @:for _, item := range *list.@(Model.ListElement.Name) {
+                @:select {
+                @:case <- cancel:
+                    @:return
+                @:default:
+                    @:resultChan <- item
+                @:}
```
marstr: As written, the above will never really consider cancellation. Instead of using "default" the code we generate should look like this:

``` Go
for _, item := range *list.Items {
    select {
    case <- cancel:
        return
    case resultChan <- item:
        // Intentionally Left Blank
    }
}
```

mcardosos: Fixed

mcardosos:
Generated code now looks like this...
```go
// GetMultiplePagesComplete gets all elements from the list without paging.
func (client PagingClient) GetMultiplePagesComplete(clientRequestID string, maxresults *int32, timeout *int32, cancel <-chan struct{}) (<-chan Product, <-chan error) {
	resultChan := make(chan Product)
	errChan := make(chan error, 1)
	go func() {
		defer func() {
			close(resultChan)
			close(errChan)
		}()
		list, err := client.GetMultiplePages(clientRequestID, maxresults, timeout)
		if err != nil {
			errChan <- err
			return
		}
		if list.Values != nil {
			for _, item := range *list.Values {
				select {
				case <-cancel:
					return
				default:
					resultChan <- item
				}
			}
		}
		for list.NextLink != nil {
			list, err = client.GetMultiplePagesNextResults(list)
			if err != nil {
				errChan <- err
				return
			}
			if list.Values != nil {
				for _, item := range *list.Values {
					select {
					case <-cancel:
						return
					case resultChan <- item:
						// Intentionally left blank
					}
				}
			}
		}
	}()
	return resultChan, errChan
}
```

src/generator/AutoRest.Go/Model/MethodGo.cs
```diff
@@ -79,6 +79,78 @@ internal void Transform(CodeModelGo cmg)
...
+                return signature.ToString();
+            }
+        }
+
+        public string ParametersDocumentation
+        {
+            get
+            {
+                StringBuilder sb = new StringBuilder();
+                foreach (var parameter in LocalParameters)
```
marstr: I don't see a new line being added into the documentation anywhere. I think it makes sense to have each parameter have its own line.

src/generator/AutoRest.Go/Model/MethodGo.cs
```diff
@@ -79,6 +79,78 @@ internal void Transform(CodeModelGo cmg)
...
+                var properties = (ReturnType.Body as CompositeTypeGo).Properties.Cast<PropertyGo>().ToList();
+                foreach (PropertyGo p in properties)
+                {
+                    if (p.ModelType is SequenceTypeGo)
+                    {
+                        return p;
+                    }
+                }
+                return null;
+            }
```
marstr: I think we can simplify this method somewhat. What about this?

``` C#
get
{
    var body = ReturnType.Body as CompositeTypeGo
    return PropertyGo(body.Properties.Where(p => p is SequenceTypeGo).FirstOrDefault())
}
````

File: src/generator/AutoRest.Go/Model/MethodGo.cs [hunk 9]
```diff
@@ -402,15 +480,56 @@ public Response ReturnValue()
...
+
+        public string NextOperationName
+        {
+            get
+            {
+                if (NextMethod != null)
+                {
+                    return NextMethod.Name.Value;
+                }
+                return null;
```
marstr: The above can be simplified to:
``` C#
get
{
    return NextMethod?.Name.Value
}
```

mcardosos:

PR #2182, opened 2017-04-19
Adding new categorizations for validation rules

dsgouda:
Addressing issue https://github.com/Azure/autorest/issues/1940
@veronicagg @vishrutshah @sarangan12 @olydis PTAL
Tested with composite swaggers, the behavior is not different than the existing behavior
By default the rules are categorized as "After Merge" and "ARM" swagger documents.


src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
```diff
@@ -41,20 +42,25 @@ internal static void AssertOnlyValidationMessage(this IEnumerable<ValidationMess
...
-                    CodeGenerator = "None",
-                    Namespace = "Test",
-                    Input = input
+                    OpenApiDocumentType = openapiDocType,
+                    MergeState = ServiceDefinitionMergeState.After
                 };
-                var modeler = new SwaggerModeler();
-                var messages = new List<LogMessage>();
+
                 Logger.Instance.AddListener(new SignalingLogListener(Category.Info, messages.Add));
```
olydis: Simply return `validator.GetValidationExceptions(new Uri(input), serviceDefinition, metaData)` instead of setting up a listener only to collect back the messages.

olydis:
> By default the rules are categorized as "Before Merge" and "Default/Regular" swagger documents.

think you mean "After Merge" :-)

src/modeler/AutoRest.Swagger/Validation/Core/RecursiveObjectValidator.cs
```diff
@@ -35,20 +35,40 @@ public RecursiveObjectValidator(Func<PropertyInfo, string> resolver)
...
         {
-            return RecursiveValidate(entity, ObjectPath.Empty, new RuleContext(entity, filePath), Enumerable.Empty<Rule>());
+            // TODO: By default, set validation rule merge state to After
+            return RecursiveValidate(entity, ObjectPath.Empty, new RuleContext(entity, filePath), Enumerable.Empty<Rule>(), metadata);
         }
 
         /// <summary>
         /// Recursively validates <paramref name="entity"/> by traversing all of its properties
         /// </summary>
         /// <param name="entity">The object to validate</param>
```
vishrutshah: missing param `metaData` in docs as well as return

src/modeler/AutoRest.Swagger/Validation/Core/RecursiveObjectValidator.cs
```diff
@@ -35,20 +35,40 @@ public RecursiveObjectValidator(Func<PropertyInfo, string> resolver)
...
+        }
+
+
+        /// <summary>
+        /// Recursively validates <paramref name="entity"/> by traversing all of its properties
+        /// </summary>
+        /// <param name="entity">The object to validate</param>
         /// <param name="parentContext">The rule context of the object that <paramref name="entity"/> belongs to</param>
         /// <param name="rules">The set of rules from the parent object to apply to <paramref name="entity"/></param>
         /// <param name="traverseProperties">Whether or not to traverse this <paramref name="entity"/>'s properties</param>
```
vishrutshah: please update documentation

src/modeler/AutoRest.Swagger/Validation/Core/RecursiveObjectValidator.cs
```diff
@@ -104,10 +124,11 @@ private IEnumerable<LogMessage> RecursiveValidate(object entity, ObjectPath enti
         /// <param name="parentContext"></param>
```
vishrutshah: please update documentation

src/modeler/AutoRest.Swagger/Validation/Core/RecursiveObjectValidator.cs
```diff
@@ -116,7 +137,7 @@ private IEnumerable<LogMessage> RecursiveValidate(object entity, ObjectPath enti
             return classRules.SelectMany(rule => rule.GetValidationMessages(entity, parentContext));
         }
 
-        private IEnumerable<LogMessage> ValidateProperty(PropertyInfo prop, object value, ObjectPath entityPath, RuleContext parentContext)
+        private IEnumerable<LogMessage> ValidateProperty(PropertyInfo prop, object value, ObjectPath entityPath, RuleContext parentContext, ServiceDefinitionMetadata metaData)
```
vishrutshah: minor: add documentation (Personally, I'd be more inclined to have detailed docs when ever possible as that saves me for bugging people :) )

src/modeler/AutoRest.Swagger/Validation/Core/Rule.cs
```diff
@@ -21,6 +22,18 @@ protected Rule()
...
         public virtual string Id => "!!! implement me and make me abstract !!!";
 
+
+        /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public virtual ServiceDefinitionDocumentType OpenApiDocumentValidationType => ServiceDefinitionDocumentType.Default;
+
+        /// <summary>
+        /// When to apply the validation rule, before or after it has been merged as a part of its composite document
```
vishrutshah: minor: I'd refrain myself from calling composite document as it confuses me with our legacy method. How aobut if we call "Merged" document from markdown file

src/modeler/AutoRest.Swagger/Validation/Core/Rule.cs
```diff
@@ -21,6 +22,18 @@ protected Rule()
         /// </summary>
         public virtual string Id => "!!! implement me and make me abstract !!!";
```
vishrutshah: Orthogonal to your change: Do we still have rules without IDs?

src/modeler/AutoRest.Swagger/Model/ServiceDefinitionMetadata.cs
```diff
@@ -0,0 +1,40 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace AutoRest.Swagger.Model
+{
+    /// <summary>
+    /// Describes whether a service definition has already been processed and merged.
+    /// This can influence ARM validation (e.g. one would not expect an operations API on individual OpenAPI definitions).
+    /// </summary>
+    public enum ServiceDefinitionMergeState
```
vishrutshah: Please add as much detailed information as you can here explaining what "Before" & "After" means and in what context like be explicit that this only works with markdown structured mechanism

src/modeler/AutoRest.Swagger/Validation/Core/Rule.cs
dsgouda: I think there are a couple. We can open a different issue for it.

dsgouda:
@vishrutshah do you think we should add this categorization information in the ValidationMessages?

src/modeler/AutoRest.Swagger/Model/ServiceDefinitionMetadata.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// AutoRest needs this categorization to choose document type-specific 
+    /// rules for execution
+    /// </summary>
+    [Flags]
+    public enum ServiceDefinitionDocumentType
+    {
+        Default = 0 >> 1,   // A plain regular swagger like pet store
+        ARM = 1 >> 2,       // An Azure Resource Management type document
+        DataPlane = 2 >> 3  // Data Plane specifications
+    }
```
olydis: wat? all of these values are 0!

should be
```
   public enum ServiceDefinitionDocumentType
   {
       Default = 0,   // <- all bits off
       ARM = 1 << 0,       // bit 0 set
       DataPlane = 1 << 1  // bit 1 set
   }
```

dsgouda: do'oh! will fix this.

dsgouda:
@olydis merge conflicts here are a little hard to understand, do you think you can take a look

dsgouda:
@olydis I tried resolving the merge conflicts but seems like there is a large diff in pipeline.ts. I tried merging it without much success, can you give it a try.

olydis:
@dsgouda sure thing

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
```diff
@@ -180,14 +182,18 @@ public void NonAppJsonTypeServiceDefinitionForConsumes()
         [Fact]
         public void NonHttpsServiceDefinitionForScheme()
         {
-            var messages = ValidateSwagger(Path.Combine(Core.Utilities.Extensions.CodeBaseDirectory, "Resource", "Swagger", "Validation", "non-https-service-def-scheme.json"));
+            var messages = ValidateSwagger(Path.Combine(Core.Utilities.Extensions.CodeBaseDirectory, "Resource", "Swagger", "Validation", "non-https-service-def-scheme.json"),
```
vishrutshah: I agree with you exhaustive testing should be a separate task. I was just thinking it'd help us if we have simple one test as that reflects your changes are working and if someone breaks it theoretically we know before we complete more testing. thoughts?

src/modeler/AutoRest.Swagger/Model/ServiceDefinitionMetadata.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    /// <summary>
+    /// Metadata for service definition that describes further what it represents (e.g. single vs. merged, ARM vs non-ARM, ...).
+    /// </summary>
+    public class ServiceDefinitionMetadata
+    {
+        public ServiceDefinitionMergeState MergeState { get; set; }
+
+        public ServiceDefinitionDocumentType OpenApiDocumentType { get; set; }
```
vishrutshah: Haha I am not good with names but up to you if you come up with some creative name :)

`OpenApiDocumentType ` is just a variable of type `ServiceDefinitionDocumentType ` why not keep that name? Just a thought

src/modeler/AutoRest.Swagger/SwaggerModeler.cs
```diff
@@ -80,7 +80,11 @@ public CodeModel Build(ServiceDefinition serviceDefinition)
             {
                 // Look for semantic errors and warnings in the document.
                 var validator = new RecursiveObjectValidator(PropertyNameResolver.JsonName);
-                foreach (var validationEx in validator.GetValidationExceptions(ServiceDefinition.FilePath, ServiceDefinition))
+                foreach (var validationEx in validator.GetValidationExceptions(ServiceDefinition.FilePath, ServiceDefinition, new ServiceDefinitionMetadata
+                {   // LEGACY MODE! set defaults for the metadata, the validator doesn't care from this codepath
```
vishrutshah: when we say `the validator doesn't care from this codepath` is it a dead code?

src/modeler/AutoRest.Swagger/Validation/Core/RecursiveObjectValidator.cs
```diff
@@ -34,21 +34,45 @@ public RecursiveObjectValidator(Func<PropertyInfo, string> resolver)
         /// <summary>
         /// Recursively validates <paramref name="entity"/> by traversing all of its properties
         /// </summary>
+        ///
```
vishrutshah: minor: missing `filePath` param

src/modeler/AutoRest.Swagger/Validation/Core/RecursiveObjectValidator.cs
```diff
@@ -34,21 +34,45 @@ public RecursiveObjectValidator(Func<PropertyInfo, string> resolver)
...
+            if (metaData.OpenApiDocumentType != ServiceDefinitionDocumentType.Default)
+            {
+                openapiTypeRules = openapiTypeRules.Concat(rules.Where(rule => (rule.OpenApiDocumentValidationType & metaData.OpenApiDocumentType)!=0));
+            }
+
+            // Filter by the current merge state, and return
+            return openapiTypeRules.Where(rule => rule.ValidationRuleMergeState == metaData.MergeState);
         }
 
+
```
vishrutshah: minor: save the line :)

src/modeler/AutoRest.Swagger/Validation/Core/RecursiveObjectValidator.cs
```diff
@@ -116,7 +142,17 @@ private IEnumerable<LogMessage> RecursiveValidate(object entity, ObjectPath enti
             return classRules.SelectMany(rule => rule.GetValidationMessages(entity, parentContext));
         }
 
-        private IEnumerable<LogMessage> ValidateProperty(PropertyInfo prop, object value, ObjectPath entityPath, RuleContext parentContext)
+
```
vishrutshah: minor: save the line

src/modeler/AutoRest.Swagger/Validation/Core/Rule.cs
```diff
@@ -21,6 +22,19 @@ protected Rule()
...
 
+
```
vishrutshah: minor: save the line

src/modeler/AutoRest.Swagger/SwaggerModeler.cs
dsgouda: It is not dead yet, but marked to be deprecated.

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
dsgouda: @olydis did you suggest testing using an md file here, I don't think that would be possible since these tests run directly with the ServiceDefinition (which is the post autorest-core processing step), the tests won't be able to work with a md file directly.
Am I missing something here.

olydis: yessss that test has to be in the core, not here ;) kinda an end-to-end thing

src/autorest-core/lib/pipeline/plugins/autorest-dotnet.ts
```diff
@@ -3,10 +3,125 @@
...
+      throw new Error(`The AutoRest dotnet extension does not offer a plugin called '${pluginName}'.`);
+    }
+    // process
+    const success = await ep.Process(pluginName, configuration, inputScope, outputScope, onMessage, CancellationToken.None);
+    if (!success) {
+      throw new Error(`Plugin ${pluginName} failed.`);
+    }
+  }
+
+  async Validate(swagger: DataHandleRead, workingScope: DataStoreView, onMessage: (message: Message) => void, isIndividual: boolean, openapiType: string): Promise<void> {
```
veronicagg: Let's add comments to the method to indicate what "isIndividual" and "openapoi-type is intended for

veronicagg: if understand correctly "openapi-type" is for "applicability" purposes, to know if the azure validator should or should not run certain rules. If these "flags" are validator specific and may change as we add more validators, would we be needing to update the signature and method implementation in this class (which seems quite generic)?

src/core/AutoRest/Plugins/AzureValidator.cs
```diff
@@ -82,7 +83,12 @@ private void LogValidationMessage(ValidationMessage validationMessage)
 
         var serviceDefinition = SwaggerParser.Load(files[0], fs);
         var validator = new RecursiveObjectValidator(PropertyNameResolver.JsonName);
-        foreach (ValidationMessage validationEx in validator.GetValidationExceptions(new Uri(files[0], UriKind.RelativeOrAbsolute), serviceDefinition))
+        var metadata = new ServiceDefinitionMetadata
+            {
+                ServiceDefinitionDocumentType = (ServiceDefinitionDocumentType)Enum.Parse(typeof(ServiceDefinitionDocumentType), (await GetValue("openapi-type"))?.ToString().ToUpper() ?? ServiceDefinitionDocumentType.ARM.ToString()),
+                MergeState = await GetValue<bool?>("is-individual-swagger") == true ? ServiceDefinitionMergeState.Before : ServiceDefinitionMergeState.After
```
veronicagg: can you add some more comments on what "merge-state" means, also this doesn't look like something created for this PR, so I'm wondering how you got to know about the existence of this flag..

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
```diff
@@ -41,21 +42,21 @@ internal static void AssertOnlyValidationMessage(this IEnumerable<ValidationMess
     [Collection("Validation Tests")]
     public partial class SwaggerModelerValidationTests
     {
-        private IEnumerable<ValidationMessage> ValidateSwagger(string input)
+        private IEnumerable<ValidationMessage> ValidateSwagger(string input, ServiceDefinitionDocumentType serviceDefDocType = ServiceDefinitionDocumentType.ARM,
```
veronicagg: could you add comments to this? I think "mergeState" is not providing much information on what it represents, or how before or after will change the output of the test, if there's something that can help make the code more readable / easier understanding of the intent, I think it'd be good to do it.

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
```diff
@@ -180,14 +181,18 @@ public void NonAppJsonTypeServiceDefinitionForConsumes()
         [Fact]
         public void NonHttpsServiceDefinitionForScheme()
         {
-            var messages = ValidateSwagger(Path.Combine(Core.Utilities.Extensions.CodeBaseDirectory, "Resource", "Swagger", "Validation", "non-https-service-def-scheme.json"));
+            var messages = ValidateSwagger(Path.Combine(Core.Utilities.Extensions.CodeBaseDirectory, "Resource", "Swagger", "Validation", "non-https-service-def-scheme.json"),
+                                           ServiceDefinitionDocumentType.ARM,
+                                           ServiceDefinitionMergeState.Before);
```
veronicagg: referring to my previous comment here, when I read ARM, I could guess this is ARM related rules, but reading MergeState.Before doesn't tell me much

src/autorest-core/lib/configuration.ts
```diff
@@ -44,6 +44,7 @@ export interface AutoRestConfigurationImpl {
   "package-version"?: string;
   "sync-methods"?: "all" | "essential" | "none";
   "payload-flattening-threshold"?: number;
+  "openapi-type"?: string
```
dsgouda: This is to specify the swagger document type (ARM/Data-Plane/Default), this is the best I could think of.

src/autorest-core/lib/pipeline/plugins/autorest-dotnet.ts
dsgouda: Do you mean we should allow for multiple openapi-types to be passed as input arguments?

src/core/AutoRest/Plugins/AzureValidator.cs
dsgouda: I have added explanation under ```ServiceDefinitionMetadata``` is that not sufficient? Not sure I understand the latter part of the comment.

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
dsgouda: I will try to add some comments here.

src/autorest-core/lib/configuration.ts
```diff
@@ -37,6 +37,8 @@ export interface AutoRestConfigurationImpl {
   "fluent"?: boolean;
   "azure-arm"?: boolean;
```
dsgouda: @olydis why does AutoRest still need this flag? to determine whether to run the validator? If yes, I think it should be renamed to something more meaningful, given we are making these changes.

olydis: `to determine whether to run the validator?` - only for legacy behavior, we already have more meaningful flags (https://github.com/Azure/autorest/blob/master/src/autorest-core/lib/pipeline/pipeline.ts#L110) and can innovate in that space.

What the flag really is for now is switching to Azure-style code gen - that's why we still have it.

src/autorest-core/lib/pipeline/plugins/autorest-dotnet.ts
olydis: not sure I understand correctly, but 99% "at the AzureValidator endpoint in validator itself" is the right answer. The core (this right here) shouldn't bother too much about details like that and just "forward" as many settings as possible. That allows updates in the validation space without changing the core.

src/autorest-core/lib/configuration.ts
dsgouda: ``` switching to Azure-style code gen``` don't think I get you

src/autorest-core/lib/pipeline/plugins/autorest-dotnet.ts
dsgouda: Fair enough. Let me work on this.

src/autorest-core/lib/configuration.ts
olydis: switching between what was formerly known as "CSharp" and "CSharp.Azure" or "Java" and "Java.Azure" is done with that flag

src/autorest-core/lib/pipeline/plugins/autorest-dotnet.ts
```diff
@@ -3,10 +3,129 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
```
olydis: this file was merged in a way that undos previous work - it's really supposed to look like https://github.com/Azure/autorest/blob/master/src/autorest-core/lib/pipeline/plugins/autorest-dotnet.ts

specifically, the class `AutoRestDotNetPlugin` is not longer used or required.

dsgouda: What happens to the ```ValidateIndividual``` and ```ValidateComposite``` methods?

olydis: handled by https://github.com/Azure/autorest/pull/2182/files#diff-10eaf6f76ff8877812109b5a5a608955R59 and calling the plugin on `config.GetPluginViews("azure-validator-individual")` right [here](https://github.com/Azure/autorest/pull/2182/files#diff-a6dae286c5e4ee8eac28f0dbb0822f1eR161)

src/modeler/AutoRest.Swagger/Model/ServiceDefinitionMetadata.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// Explanation:
+    /// 'Before Merge' state: This is the state where the json provided is considered 
+    /// in isolation and the corresponding rules are applied. This is equivalent to 
+    /// running validator in the  non-CompositeSwagger mode per legacy API
+    /// 'After Merge' state: For a given json, the AutoRest core will find 
+    /// its corresponding '.md' file and construct a ServiceDefinition object. 
+    /// This will be equivalent to running AutoRest in CompositeSwagger mode per 
+    /// legacy cli. 
+    /// </summary>
+    public enum ServiceDefinitionMergeState
```
olydis: lol when (and why) did you change those things to all upercase!? that's like not really according to C# style guidelines o.O

src/modeler/AutoRest.Swagger/Model/ServiceDefinitionMetadata.cs
```diff
@@ -0,0 +1,52 @@
...
+
+    /// <summary>
+    /// Represents Service Definition Document Type 
+    /// AutoRest needs this categorization to choose document type-specific 
+    /// rules for execution
+    /// </summary>
+    [Flags]
+    public enum ServiceDefinitionDocumentType
+    {
+        DEFAULT = 0 << 1,   // A plain regular swagger like pet store
```
olydis: WTF this is broken AGAIN!

src/modeler/AutoRest.Swagger/Model/ServiceDefinitionMetadata.cs
```diff
@@ -0,0 +1,52 @@
...
+
+    /// <summary>
+    /// Metadata for service definition that describes further what it represents 
+    /// (e.g. single vs. merged, ARM vs non-ARM, ...).
+    /// </summary>
+    public class ServiceDefinitionMetadata
+    {
+        public ServiceDefinitionMergeState MergeState { get; set; }
+
+        public ServiceDefinitionDocumentType[] ServiceDefinitionDocumentTypes { get; set; }
```
olydis: not an array! The `[Flags]` already enable multiple document types...

src/core/AutoRest/Plugins/AzureValidator.cs
```diff
@@ -11,14 +11,36 @@
...
 public class AzureValidator : NewPlugin
 {
     public AzureValidator(Connection connection, string sessionId) : base(connection, sessionId)
     { }
 
+    private readonly ServiceDefinitionDocumentType[] DefaultDocumentTypes = new ServiceDefinitionDocumentType[] { ServiceDefinitionDocumentType.ARM };
+
     private readonly Regex resPathPattern = new Regex(@"/providers/(?<providerNamespace>[^{/]+)/((?<resourceType>[^{/]+)/)?");
 
+    private ServiceDefinitionDocumentType[] ParseServiceDefinitionDocumentType(string input)
```
olydis: `Enum.Parse` already parses `,` separated `Flags` enums. Think this method is redundant.

src/autorest-core/test/resources/literate-example/csharp/Models/Error.cs
```diff
@@ -0,0 +1,48 @@
+// Code generated by Microsoft (R) AutoRest Code Generator 1.0.1.0
+// Changes may cause incorrect behavior and will be lost if the code is
```
olydis: why are these files (`src/autorest-core/test/resources/literate-example/csharp/`) in this PR?

src/modeler/AutoRest.Swagger/Model/ServiceDefinitionMetadata.cs
olydis: please refer to https://github.com/Azure/autorest/pull/2182#discussion_r113528757 for explanation

src/modeler/AutoRest.Swagger/Model/ServiceDefinitionMetadata.cs
dsgouda: Did not know that!

olydis: it's what we've discussed last week -- that's why we should also not use `==` for checking but `&` to see if the "sets" overlap

src/modeler/AutoRest.Swagger/Model/ServiceDefinitionMetadata.cs
dsgouda: Yeah, this got me thinking whether ```Parse``` consider the string to be case insensitive, so simply set everything to uppercase. 
Alternative suggestions?

src/modeler/AutoRest.Swagger/Model/ServiceDefinitionMetadata.cs
dsgouda: Messed this when I went converting everything to uppercase.

src/autorest-core/test/resources/literate-example/csharp/Models/Error.cs
dsgouda: Autogened stuff, no idea how these got generated, let me check

src/modeler/AutoRest.Swagger/Model/ServiceDefinitionMetadata.cs
olydis: it is case insensitive if you pass `true` as 3rd param: https://msdn.microsoft.com/en-us/library/kxydatf9(v=vs.110).aspx

src/core/AutoRest/Plugins/AzureValidator.cs
```diff
@@ -85,7 +85,7 @@ private void LogValidationMessage(ValidationMessage validationMessage)
         var validator = new RecursiveObjectValidator(PropertyNameResolver.JsonName);
         var metadata = new ServiceDefinitionMetadata
             {
-                ServiceDefinitionDocumentType = (ServiceDefinitionDocumentType)Enum.Parse(typeof(ServiceDefinitionDocumentType), (await GetValue("openapi-type"))?.ToString().ToUpper() ?? ServiceDefinitionDocumentType.ARM.ToString()),
+                ServiceDefinitionDocumentType = (ServiceDefinitionDocumentType)Enum.Parse(typeof(ServiceDefinitionDocumentType), (await GetValue("openapi-type"))?.ToString().ToUpper() ?? ServiceDefinitionDocumentType.ARM.ToString(), true),
```
olydis: please remove `.ToUpper()` to remove noise

src/modeler/AutoRest.Swagger/Model/ServiceDefinitionMetadata.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// Represents Service Definition Document Type 
+    /// AutoRest needs this categorization to choose document type-specific 
+    /// rules for execution
+    /// </summary>
+    [Flags]
+    public enum ServiceDefinitionDocumentType
+    {
+        Default = 0,   // A plain regular swagger like pet store
+        ARM = 1 << 2,       // An Azure Resource Management type document
+        DataPlane = 2 << 3  // Data Plane specifications
```
olydis: just making sure we don't miss it: this is still not fixed.

olydis:
@dsgouda the enum is still weird! Please please please refer to https://github.com/Azure/autorest/pull/2182#discussion_r113528757 or argue why your version makes more sense ;) Like it is currently, I would find it slightly confusing to extend or edit this enum.

Samples/1c-multiple-input-swaggers/Client/IStorageManagementClient.cs
```diff
@@ -0,0 +1,63 @@
+// Code generated by Microsoft (R) AutoRest Code Generator 1.0.1.0
```
olydis: why is this part of this PR? I'd expect zero change in code generation due to your changes

olydis: hmmm seems like some PR in the past overruled CI in the past when it really shouldn't, because this kind of change is unacceptable :-( will investigate, but it's fine to submit those with this PR

dsgouda: Thanks, I wasn't sure if it was my change that did this but multiple regens gave this same diff

Samples/2a-arm-validation/shell/stderr.txt
```diff
@@ -56,3 +56,61 @@ WARNING (DescriptionMissing/M4000/SDKViolation): 'properties' model/property lac
     - https://raw.githubusercontent.com/Azure/azure-rest-api-specs/d374d03801e97737ddb32e01f20513e7b2bbd9c3/arm-storage/2015-06-15/swagger/storage.json:716:8 ($.definitions.StorageAccountUpdateParameters.properties.properties)
 WARNING (DescriptionMissing/M4000/SDKViolation): 'keyName' model/property lacks 'description' property. Consider adding a 'description' element. Accurate description is essential for maintaining reference documentation.
     - https://raw.githubusercontent.com/Azure/azure-rest-api-specs/d374d03801e97737ddb32e01f20513e7b2bbd9c3/arm-storage/2015-06-15/swagger/storage.json:726:8 ($.definitions.StorageAccountRegenerateKeyParameters.properties.keyName)
+WARNING (NonAppJsonTypeWarning/S2004/SDKViolation): Please make sure that media types other than 'application/json' are supported by your service.
```
olydis: aaaand this is why we have that CI in place: these messages look duplicate to me, so maybe the validator is run with the *same* mode twice instead of switching between individual and composite? Note that this should also have been caught by an actual test case - this regen diff should always be the last resort of finding out about stuff ;-)

src/core/AutoRest/Plugins/AzureValidator.cs
```diff
@@ -71,22 +72,26 @@ private void LogValidationMessage(ValidationMessage validationMessage)
...
-        var content = await ReadFile(files[0]);
-        var fs = new MemoryFileSystem();
-        fs.WriteAllText(files[0], content);
+            var serviceDefinition = SwaggerParser.Load(file, fs);
+            var validator = new RecursiveObjectValidator(PropertyNameResolver.JsonName);
+            
+            var metadata = new ServiceDefinitionMetadata
+            {
+                ServiceDefinitionDocumentType = (ServiceDefinitionDocumentType)Enum.Parse(typeof(ServiceDefinitionDocumentType), (await GetValue("openapi-type"))?.ToString() ?? ServiceDefinitionDocumentType.ARM.ToString(), true),
+                MergeState = await GetValue<string>("merge-state") == "composed" ? ServiceDefinitionDocumentState.Individual : ServiceDefinitionDocumentState.Composite
```
olydis: wrong way around

olydis:
@veronicagg what do you think about the PR?

docs/developer/guide/writing-a-new-plugin.md
```diff
@@ -5,7 +5,20 @@ Add an entry to the (configuration file)[https://github.com/Azure/autorest/blob/
 add the entry in the corresponding subsection (eg. validators should go in the ```Validators``` section)
 Please note that the ```transform``` step is pre-cursor to most plugins since it transforms the raw json into a usable model.
```
olydis: that's not what `transform` does. 99% of the time it does nothing, it handles overrides requested by
- `directive` clauses in the configuration (say someone wants a different method name when generating for Ruby than when generating for C# - no way to express that in Swagger)
- description overrides delivered by our localization engine

src/autorest-core/resources/default-configuration.md
```diff
@@ -93,9 +93,33 @@ pipeline:
...
+Activate `azure-validator` when setting `azure-arm`!?
+
+``` yaml $(azure-arm)
+azure-validator: true
+```
+
+``` yaml $(azure-validator)
+azure-validator-composed:
+  merge-state: composed
+  openapi-type: arm
```
olydis: I would not state that in here (and below) due to duplication! even worse, since nested settings have more priority than "top level" ones, people cannot simply override those (which is kind of an important thing to enable)!

They would also have to write
```
azure-validator-composed:
  openapi-type: foo
azure-validator-individual:
  openapi-type: foo
```
which is horrible. Just set this default on the top level and everything is fine ;)

docs/developer/guide/writing-a-new-plugin.md
```diff
@@ -5,7 +5,20 @@ Add an entry to the (configuration file)[https://github.com/Azure/autorest/blob/
...
+Eg.:
+
+```
+swagger-document/individual/azure-validator:
+    input: individual/transform
+    scope: azure-validator-individual
+
+azure-validator-individual:
+  merge-state: individual
+  openapi-type: arm
```
olydis: remove this, as explained in my comment on `default-configuration.md`

docs/developer/guide/writing-a-new-plugin.md
```diff
@@ -5,7 +5,20 @@ Add an entry to the (configuration file)[https://github.com/Azure/autorest/blob/
...
+```
+
+In the example above, ```swagger-document/individual/azure-validator``` is the plugin to run, ```individual/transform``` is the step before this plugin is executed or the input which this plugin would process and ```azure-validator-individual``` is its scope. The inputs to this scope are ```merge-state``` and ```openapi-type``` which are defined in a different section.
```
olydis: 👍  (also remove `openapi-type`)

docs/developer/guide/writing-a-new-plugin.md
```diff
@@ -5,7 +5,20 @@ Add an entry to the (configuration file)[https://github.com/Azure/autorest/blob/
...
 
 ## Add a plugin entry to the Configuration object
 Add an entry to the (Configuration object)[https://github.com/Azure/autorest/blob/master/src/autorest-core/lib/configuration.ts]. This entry determines the input flag that will enable/disable the plugin
```
olydis: no, this object currently does the mapping between the name of the plugin and the actual thing to execute - this will also move to config file as soon as Garrett has solved acquisition

src/autorest-core/resources/default-configuration.md
dsgouda: not sure I follow this, defining multiple settings under the same scope is nested? What does setting default on the top mean?

docs/developer/guide/writing-a-new-plugin.md
dsgouda: Why should we remove ```openapi-type```, until we set categorizations for each validation rule, it would make sense to set this as a default, no?

src/autorest-core/resources/default-configuration.md
olydis: move the default out of the scopes. just a single line on the root level instead of two copies in a deeper level.

docs/developer/guide/writing-a-new-plugin.md
olydis: default yes, but not in that scope, as described above

docs/developer/guide/writing-a-new-plugin.md
```diff
@@ -0,0 +1,25 @@
...
+
+azure-validator-individual:
+  merge-state: individual
+```
+
+In the example above, ```swagger-document/individual/azure-validator``` is the plugin to run, ```individual/transform``` is the step before this plugin is executed or the input which this plugin would process and ```azure-validator-individual``` is its scope. The setting for this scope is ```merge-state``` which is defined in a different section.
+
+## Add a plugin entry to the Configuration object
+Add an entry to the [configuration file](https://github.com/Azure/autorest/blob/master/src/autorest-core/lib/configuration.ts). This entry denotes name of the plugin. This name is then mapped to the actual process in ```configuration.ts```
+and the prcoess to be loaded when the plugin is requested. This process need to be a node or C# process and is completely independent of the AutoRest process, the inputs to be provided to this process
```
veronicagg: minor typo "process" instead of "prcoess"

src/core/AutoRest/Plugins/AzureValidator.cs
```diff
@@ -71,22 +72,26 @@ private void LogValidationMessage(ValidationMessage validationMessage)
...
 
-        var content = await ReadFile(files[0]);
-        var fs = new MemoryFileSystem();
-        fs.WriteAllText(files[0], content);
+            var serviceDefinition = SwaggerParser.Load(file, fs);
+            var validator = new RecursiveObjectValidator(PropertyNameResolver.JsonName);
+            
+            var metadata = new ServiceDefinitionMetadata
+            {
+                ServiceDefinitionDocumentType = (ServiceDefinitionDocumentType)Enum.Parse(typeof(ServiceDefinitionDocumentType), (await GetValue("openapi-type"))?.ToString() ?? ServiceDefinitionDocumentType.ARM.ToString(), true),
```
veronicagg: which file is this info getting pulled from? is it the default config file? If so, I'd expect we will always find the values defined right?

veronicagg: why are we calling this "ServiceDefinitionDocumentType" and the value in the config file "openapi-type", should these be the same?

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
```diff
@@ -180,14 +185,18 @@ public void NonAppJsonTypeServiceDefinitionForConsumes()
...
+                                           ServiceDefinitionDocumentType.ARM,
+                                           ServiceDefinitionDocumentState.Individual);
             messages.AssertOnlyValidationWarning(typeof(SupportedSchemesWarning));
         }
 
         [Fact]
         public void NonHttpsOperationsForScheme()
         {
-            var messages = ValidateSwagger(Path.Combine(Core.Utilities.Extensions.CodeBaseDirectory, "Resource", "Swagger", "Validation", "non-https-operations-scheme.json"));
+            var messages = ValidateSwagger(Path.Combine(Core.Utilities.Extensions.CodeBaseDirectory, "Resource", "Swagger", "Validation", "non-https-operations-scheme.json"),
```
veronicagg: why do we need to add the servicedefintiondocumenttype to the test? doesn't each rule know whether it's applicable at individual or composed level?

src/modeler/AutoRest.Swagger/Model/ServiceDefinitionMetadata.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Collections.Generic;
+
+namespace AutoRest.Swagger.Model
+{
+    /// <summary>
+    /// Describes whether a service definition has already been processed and merged.
+    /// This can influence ARM validation (e.g. one would not expect an operations API 
+    /// on individual OpenAPI definitions).
+    /// Explanation:
+    /// 'Before Merge' state: This is the state where the json provided is considered
```
veronicagg: should comments be updated?

src/modeler/AutoRest.Swagger/SwaggerModeler.cs
```diff
@@ -73,7 +73,11 @@ public CodeModel Build(ServiceDefinition serviceDefinition)
             {
                 // Look for semantic errors and warnings in the document.
                 var validator = new RecursiveObjectValidator(PropertyNameResolver.JsonName);
-                foreach (var validationEx in validator.GetValidationExceptions(ServiceDefinition.FilePath, ServiceDefinition))
+                foreach (var validationEx in validator.GetValidationExceptions(ServiceDefinition.FilePath, ServiceDefinition, new ServiceDefinitionMetadata
+                {   // LEGACY MODE! set defaults for the metadata, marked to be depracted
```
veronicagg: minor: typo in comment deprecated

src/modeler/AutoRest.Swagger/SwaggerModeler.cs
```diff
@@ -179,7 +183,11 @@ public override IEnumerable<LogMessage> Compare()
 
             // Look for semantic errors and warnings in the new document.
             var validator = new RecursiveObjectValidator(PropertyNameResolver.JsonName);
-            var LogMessages = validator.GetValidationExceptions(newDefintion.FilePath, newDefintion).ToList();
+            var LogMessages = validator.GetValidationExceptions(newDefintion.FilePath, newDefintion, new ServiceDefinitionMetadata
```
veronicagg: is this also for legacy? if not, I'm not sure why we're setting defaults.

src/modeler/AutoRest.Swagger/Validation/Core/RecursiveObjectValidator.cs
```diff
@@ -34,21 +34,44 @@ public RecursiveObjectValidator(Func<PropertyInfo, string> resolver)
...
         /// </summary>
-        /// <param name="entity">The object to validate</param>
-        public IEnumerable<LogMessage> GetValidationExceptions(Uri filePath, ServiceDefinition entity)
+        /// <param name="filePath">uri path to the servicedefinition document <paramref name="entity"/></param>
+        /// <param name="entity">The object to validate</param><param name="entity">The object to validate</param>
+        /// <param name="metaData">The metadata associated with serviceDefinition to which <paramref name="metaData"/> belongs to</param>
+        public IEnumerable<LogMessage> GetValidationExceptions(Uri filePath, ServiceDefinition entity, ServiceDefinitionMetadata metadata)
         {
-            return RecursiveValidate(entity, ObjectPath.Empty, new RuleContext(entity, filePath), Enumerable.Empty<Rule>());
+            // TODO: By default, set validation rule merge state to After
```
veronicagg: what is this TODO about?

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
```diff
@@ -40,6 +40,11 @@ public class ServiceDefinitionParameters : TypedRule<Dictionary<string, SwaggerP
         public override string MessageTemplate => Resources.ServiceDefinitionParametersMissingMessage;
 
         /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
```
veronicagg: could a rule fall in 2 buckets? ARM and Data-plane?

src/autorest-core/lib/configuration.ts
dsgouda: @olydis do we?

src/core/AutoRest/Plugins/AzureValidator.cs
dsgouda: 1. It can be either from the default config file or the command line. Command line arguments trump whatever is specified in the config file provided we also pass in the correct scope.
2. ```openapi-type``` describes the type of specification provided while ```ServiceDefinitionDocumentType``` is more like metadata for the service document object generated from the provided spec, hence the naming.

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
dsgouda: Each rule does know where its applicable, but the test harness should know which rules to pick, think of this like the flag we pass from the default config file.

src/modeler/AutoRest.Swagger/SwaggerModeler.cs
dsgouda: Not legacy, just redundant code lying around scheduled to be cleaned up soon. Had to add defaults to appease the compiler. @olydis can you confirm this

src/modeler/AutoRest.Swagger/Validation/ServiceDefinitionParameters.cs
dsgouda: Not that I can think of, but if it does, we can always use the ```Default``` flag.

src/core/AutoRest/Plugins/AzureValidator.cs
```diff
@@ -80,11 +80,25 @@ private void LogValidationMessage(ValidationMessage validationMessage)
...
             var serviceDefinition = SwaggerParser.Load(file, fs);
             var validator = new RecursiveObjectValidator(PropertyNameResolver.JsonName);
-            
+            var docTypeInput = (await GetValue<string>("openapi-type"));
+            var docStateInput = (await GetValue<string>("merge-state"));
+
+            ServiceDefinitionDocumentType docType;
+            if (!Enum.TryParse<ServiceDefinitionDocumentType>(docTypeInput, true, out docType))
+            {
+                throw new Exception("Invalid Input for openapi-type: " + docTypeInput + ". Valid values are 'arm', 'data-plane' or 'default'.");
```
veronicagg: do we need to include openapi-type in documentation somewhere, so folks know what values should be used and for what?

veronicagg:
CI is failing though, so make sure to check on it.

src/core/AutoRest/Plugins/AzureValidator.cs
dsgouda: Will add this in the documentation

PR #2183, opened 2017-04-19
Some end to end validation fixes

olydis:
a couple of fixes to validation rules and the way their messages are processed in order to make sure the existing reporting tools work (https://github.com/olydis/azure-rest-api-specs/blob/master/scripts/getStatus.js is broken).


dsgouda:
Might need a regen, but looks good otherwise

olydis:
Thank you sir!

PR #2186, opened 2017-04-20
Long-running operations will now return the object they create.

jhendrixMSFT:
LROs have been returning an autorest.Response object instead of the object
they create which requires the end-user to make a separate API call to
retrieve the object which is clunky and inefficient.  This change remove
that behavior so the API will return the object the LRO created.


olydis:
I love PRs that remove code to deliver a feature 😄

Issue #1953, opened 2017-03-08
New rule: property with "enum" should have an x-ms-enum extension


veronicagg:
M2018: A parameter or a model property having an "enum": [] constraint, MUST have an "x-ms-enum" extension and the extension's "modelAsString" property MUST be set to false where applicable. (clarify "where applicable")
Category: SDK Violation
Severity: Error

dsgouda:
While enforcing an ```x-ms-enum``` extension for enums is a good idea, I'm not sure why we should encourage what ```modelAsString``` should be set as. At most I think the message should simply convey that not setting ```modelAsString``` to true generates client sdks that will have breaking changes whenever the enums' values are modified.
Thoughts or suggestions?

veronicagg:
@amarzavery could you comment on this?
I believe the intention is to enforce "x-ms-enum" but not necessarily "modelAsString" set to false (because it says where applicable), I think the way this is written in the checklist is a bit misleading. 
Also, is it true that we should have every enum be x-ms-enum? what happens if it's a real enum and people don't use x-ms-enum, what effect does it cause during code generation?

dsgouda:
On similar lines, I wonder if it makes sense to have a linter rule with warning severity that simply flags ```modelAsString:false``` might result into a breaking change every time the enum values change.

veronicagg:
@dsgouda I think that may make sense. @kirthik thoughts? I'm wondering where we draw the line between warnings and guidelines, it seems like a good idea to throw warnings for specific things found in the specs, so people don't have to read the full set of guidelines. On the other hand, we should be careful on flooding the output with warnings. I'm wondering if we should bring in an "Info" category for the "guidelines".

PR #2187, opened 2017-04-20
Validation rule to suggest x-ms-enum extension for enums

dsgouda:
Fixing issue https://github.com/Azure/autorest/issues/1953


docs/developer/validation-rules/x-ms-enum.md
```diff
@@ -0,0 +1,5 @@
+# ValidFormats
+## Description
+For enum types, we should encourage authors to use the `x-ms-enum` extension and apply appropriate options. Please refer to the dcoumentation details [here](https://github.com/Azure/azure-rest-api-specs/blob/master/documentation/swagger-extensions.md#x-ms-enum)
```
veronicagg: typo: documenation. 
nit: Instead of "here", I'd probably just say "x-ms-enum".

docs/developer/validation-rules/x-ms-enum.md
```diff
@@ -0,0 +1,5 @@
...
+## How to fix
+Add the `x-ms-enum` extension to the enum and set appropriate properties including `modelAsString`.
```
veronicagg: should we write here what you pointed out before... setting modelAsString: true, can help avoid SDK breaking changes in the future if you think there could be more values added to the enum. ?

docs/developer/validation-rules/x-ms-enum.md
```diff
@@ -0,0 +1,5 @@
+# ValidFormats
```
veronicagg: This title doesn't seem to match the content? could you add the ID at the top? similar to what's in https://github.com/Azure/autorest/commit/e225e61e36644ba9119aa4996e2d3554cbe2c78c

veronicagg:
please check CI as well.

docs/developer/validation-rules/x-ms-enum.md
dsgouda: I would prefer we write a separate linter rule with warning severity to flag this, I have proposed this on the issue too.

PR #2190, opened 2017-04-21
use default value in constructor (if you ask for it nicely)

olydis:
This PR puts a customization point into the C# generator that allows enabling default values on model constructors. 

This is how you activate it (for all composite types):
``` yaml
  directive:
    from: code-model-v1
    where: $..*[?(@.$type == "AutoRest.Core.Model.CompositeType, AutoRest.Core")].properties[*]
    transform: $.useDefaultInConstructor = true
```


olydis:
addresses #781 in C#

Issue #2167, opened 2017-04-17
Allow upto three consecutive upper case letter in the properties casing


salameer:
This is to Support when Acronyms collide with the beginning of another word

Issue #2064, opened 2017-03-25
[Potential Bug] Should camelCase apply to @odata.***


vishrutshah:
Linter Rule [DefinitionsPropertiesNamesCamelCase](https://github.com/Azure/autorest/blob/1fbf2aa7ca0fd65c9421e3043062e3f14e0aa5f6/src/modeler/AutoRest.Swagger/Validation/DefinitionsPropertiesNamesCamelCase.cs#L57)

**Reported by** @brjohnstmsft
**Issue:** M3016: "Property named: "@odata.type", for definition: "Analyzer" must follow camelCase style. Example: "@odata.type"." I don't understand why the example exactly matches the property name. In any case, the casing of this name is dictated by the OData protocol and we cannot change it. Same for "@odata.etag"
**PR:** https://github.com/Azure/azure-rest-api-specs-pr/pull/47

@ravbhatnagar : Can you please verify wether this requirement is valid for odata type?

**cc:**
@veronicagg

brjohnstmsft:
This requirement is not valid for OData annotations. Here is the relevant part of the OData specification: http://docs.oasis-open.org/odata/odata-json-format/v4.0/errata03/os/odata-json-format-v4.0-errata03-os-complete.html#_Toc453766627

dsgouda:
To minimize the scope of exceptional cases, I will modify the code to ignore cases where the property names start with `@odata.` Feel free to point out any edge cases we can run into.

brjohnstmsft:
@dsgouda If you plan to run these rules on data plane specs, then please exclude all property names that start with `@`. Built-in annotations are covered by `@odata`, but services are free to define their own custom annotation, which can have any name as long as it starts with `@`.

dsgouda:
@brjohnstmsft Thanks, appreciate that!

veronicagg:
@dsgouda the rule may be looking for '@' and in this case, it doesn't look to have it, we may need to ignore starting with "odata" as well.

PR #2191, opened 2017-04-21
Improvements for camel casing based validation rule

dsgouda:
Addressing issues https://github.com/Azure/autorest/issues/2064 and https://github.com/Azure/autorest/issues/2167

### DIFF
--- a/src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/property-names-casing.json
+++ b/src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/property-names-casing.json
@@ -88,6 +88,15 @@
           "type": "string",
           "description": "Required. Gets or sets the location of the resource. This will be one of the supported and registered Azure Geo Regions (e.g. West US, East US, Southeast Asia, etc.). The geo region of a resource cannot be changed once it is created, but if an identical geo region is specified on update the request will succeed."
         },
+        "@odata.etag": {
+          "x-ms-client-name": "ETag",
+          "type": "string",
+          "description": "The ETag of the synonym map."
+        },
+        "tripleCAPitalCASe": {
+          "type": "string",
+          "description": "testing 3 capitalized letters in camelCase"
+        },
         "tags": {
           "type": "object",
           "additionalProperties": {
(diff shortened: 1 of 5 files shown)


src/modeler/AutoRest.Swagger.Tests/Resource/Swagger/Validation/property-names-casing.json
```diff
@@ -88,6 +88,15 @@
           "type": "string",
           "description": "Required. Gets or sets the location of the resource. This will be one of the supported and registered Azure Geo Regions (e.g. West US, East US, Southeast Asia, etc.). The geo region of a resource cannot be changed once it is created, but if an identical geo region is specified on update the request will succeed."
         },
+        "@odata.etag": {
+          "x-ms-client-name": "ETag",
+          "type": "string",
+          "description": "The ETag of the synonym map."
+        },
+        "tripleCAPitalCASe": {
```
veronicagg: should we call the property publicIPAddress to show a real example, and keep the description you have, which explains this case?

src/modeler/AutoRest.Swagger/Validation/BodyPropertiesNamesCamelCase.cs
```diff
@@ -41,7 +41,7 @@ public class BodyPropertiesNamesCamelCase : TypedRule<Dictionary<string, Operati
         /// <summary>
         /// The severity of this message (ie, debug/info/warning/error/fatal, etc)
         /// </summary>
-        public override Category Severity => Category.Error;
+        public override Category Severity => Category.Warning;
```
veronicagg: why are we turning this one into a Warning?

dsgouda: Since we discussed turning naming related validation rules to Warnings, I thought this would also fall under the same requirements.

veronicagg: Oh, I see. I agree that this is a naming convention rule, I'm not sure we want to turn this one into a Warning yet though. I think the naming ones we want to turn into Warnings are the ones about operation names, I'd start with those and we can evaluate this one if needed afterwards.

veronicagg:
Changes look good to me, I would just leave this one as an Error for now.

Issue #2184, opened 2017-04-19
Turn naming convention rules into warnings


veronicagg:
Let's turn all rules on naming conventions to Warnings. I expect most should be SDKViolations, if there are any RPC violations, let's evaluate which ones those are before changing Severity.

PR #2192, opened 2017-04-21
Convert naming rules to warnings

dsgouda:
Addressing https://github.com/Azure/autorest/issues/2184


src/modeler/AutoRest.Swagger/Validation/OperationIdSingleUnderscore.cs
```diff
@@ -39,7 +39,7 @@ public override bool IsValid(string entity)
         /// <summary>
         /// The severity of this message (ie, debug/info/warning/error/fatal, etc)
         /// </summary>
-        public override Category Severity => Category.Error;
+        public override Category Severity => Category.Warning;
```
veronicagg: this may be a rule we need to enforce, due to AutoRest behavior, therefore leaving it as an error. What would happen at code generation time if there's more than 1 underscore?

src/modeler/AutoRest.Swagger/Validation/ParameterNameValidation.cs
```diff
@@ -32,7 +32,7 @@ public class ParameterNameValidation : TypedRule<SwaggerParameter>
         /// <summary>
         /// The severity of this message (ie, debug/info/warning/error/fatal, etc)
         /// </summary>
-        public override Category Severity => Category.Error;
```
veronicagg: this doesn't look to be a naming convention rule, I think we need to skip this one.

src/modeler/AutoRest.Swagger/Validation/OperationIdSingleUnderscore.cs
dsgouda: Good catch, I went a little overboard there.

PR #2197, opened 2017-04-24
Docs

fearthecowboy:
More docs


docs/installing-autorest.md
```diff
@@ -0,0 +1,27 @@
...
+### Updating Autorest
+  To update AutoRest if you have previous versions installed, please run:
+    
+  ``` powershell
+  autorest --latest
+  ``` 
+or 
+  ```powershell
+  # Removes all other versions and installs the latest
+  autorest --reset
```
olydis: that one actually errors for missing input files I think... will apply same fix as for `--latest`

PR #2178, opened 2017-04-18
Code changes to remove BodyParametersValidation Rule

sarangan12:
@vishrutshah @veronicagg @salameer @dsgouda Could you please review and approve?

Related to: https://github.com/Azure/autorest/issues/2173

PR #2199, opened 2017-04-25
Remove bp rule

sarangan12:


veronicagg:
This is the same as https://github.com/Azure/autorest/pull/2178, right?

veronicagg:
any thoughts on what's causing the CI is failing?

sarangan12:
@veronicagg Yes. That was what I was playing with for the past few days. I finally figured the difference in the dotnet environment that I used and fixed. So, now the build is passing. Can you approve?

PR #2200, opened 2017-04-25
Fix formatting.

jhendrixMSFT:
The formatting of MethodGo.cs is all screwy so I used VS to auto-format
it.  There are no code changes with this commit.


marstr:
Looks like I have the wrong line above, but my only request is that I don't have to fight against VSCode to get things formatted.

jhendrixMSFT:
And it's more than just MethodGo.cs, see https://github.com/Azure/autorest/pull/2040.

jhendrixMSFT:
@olydis @fearthecowboy Can you guys take a look at the changes to settings.json please?

PR #2201, opened 2017-04-25
Make long-running operations asynchronous.

jhendrixMSFT:
The current implementation requires end-users to write their own logic to
make LROs asynchronous; this change provides a default implementation.
LROs will execute in a go routine, using channels to communicate their
return values.

### DIFF
--- a/src/generator/AutoRest.Go/Templates/MethodTemplate.cshtml
+++ b/src/generator/AutoRest.Go/Templates/MethodTemplate.cshtml
@@ -40,19 +40,21 @@
 }
 
 func (client @(Model.Owner)) @(Model.MethodSignature) (@Model.MethodReturnSignature(false)) {
-@if (Model.IsLongRunningOperation())
+@if (Model.ReturnViaChannel)
 {
+    @:resultChan := make(chan @(Model.MethodReturnType), 1)
     @:errChan:= make(chan error, 1)
 }
 @if (Model.ParameterValidations.Length > 0)
 {
     @:if err := validation.Validate([]validation.Validation{
     @:@(Model.ParameterValidations)}); err != nil {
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
An example of the codegen can be see here: https://github.com/jhendrixMSFT/azure-sdk-for-go/commit/f9b17379fbcbe38b3a7e72bafac779a22cd79fc3

mcardosos:
qq, now that I am working on paging, am I correct in assuming `ListEverything` in LROs should also return a chan of whatevers? Or do you have another idea in mind?

jhendrixMSFT:
@mcardosos yes I think it should return a slice via a channel.

jhendrixMSFT:
@mcardosos @marstr don't forget about this one.  :)

mcardosos:
qq. how does the generated code for paged LROs look?

src/generator/AutoRest.Go/Templates/MethodTemplate.cshtml
```diff
@@ -39,35 +39,69 @@
...
-         @:@(Model.ParameterValidations)}); err != nil {
-             @:return result, @(Model.ValidationError)
-    @:}
-    @:@EmptyLine
+    @:@(Model.ParameterValidations)}); err != nil {
+    @if (Model.IsLongRunningOperation())
+    {
+    @:errChan <- validation.NewErrorWithValidationError(err, "redis.GroupClient", "Create")
+    @:close(errChan)
+    @:return nil, errChan
```
marstr: Returning the nil channel will cause anybody reading it to block indefinitely. :) Demonstrated here: https://play.golang.org/p/09rkpDatlH

I'd refactor this so that the creation and deferral of closing them happens all before even this error handling.

Issue #2122, opened 2017-04-05
Linter: flatten warning ignores explicit `"x-ms-flatten":false`


jaredmoo:
AutoRest gives me this warning:

```
WARNING: Consider using x-ms-client-flatten to provide a better end user experience
        Path: file:///E:/dsmain1/retail/amd64/Xdb/Manifest/Svc/ManagementService.WebApiHosting.ARM.Tools/swagger/swagger.json#/definitions/Operation/properties/properties
```

This is a valuable linter rule, but for this particular property I don't want to flatten. I tried added `"x-ms-flatten":false` to explicitly not flatten the property, but the linter still gives the warning.

dsgouda:
@jaredmoo I could not replicate this locally. I get a feeling you are using `"x-ms-flatten":false` instead of `"x-ms-client-flatten":true`
Please confirm this

dsgouda:
@veronicagg What should be the expected behavior here. I guess we should suggest authors use the `x-ms-client-flatten` extension and set it to true for non-flattened properties, explicitly setting it to false should not really suppress the warning.

jaredmoo:
I am totally certain that I don't want to flatten this property. So this warning is noise to me. It would be really helpful if I could silence it. It seems to me that setting flatten explicitly to false should do that, but perhaps you might come up with another way to silence it.

dsgouda:
I guess we should rely on suppressions in cases where we explicitly chose to ignore a warning and silence it. I'm still open to any other ideas here.

veronicagg:
@dsgouda In this case, if we know that the author explicitly set "x-ms-flatten":"false", it may be fair to not show the warning, as it appears like an intentional decision. @kirthik, @vishrutshah @sarangan12 @amarzavery thoughts?

vishrutshah:
I'd agree that when customers are making an "explicit conscious" choice then our linter rule should respect users decision. When user wants to fix the rule / will decide to never fix that in future then i'd consider that for suppression.

kirthik:
Agreed. We should not show the warning when `x-ms-flatten` is set to `false`

PR #2204, opened 2017-04-26
Disabling warning if x-ms-client-flatten explicitly set to false

dsgouda:
Addresses issue https://github.com/Azure/autorest/issues/2122

### DIFF
--- a/src/modeler/AutoRest.Swagger/Validation/AvoidNestedProperties.cs
+++ b/src/modeler/AutoRest.Swagger/Validation/AvoidNestedProperties.cs
@@ -12,7 +12,7 @@ namespace AutoRest.Swagger.Validation
 {
     public class AvoidNestedProperties : TypedRule<Schema>
     {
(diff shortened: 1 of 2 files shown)


src/modeler/AutoRest.Swagger/Validation/AvoidNestedProperties.cs
```diff
@@ -35,8 +35,7 @@ public override bool IsValid(Schema schema, RuleContext context)
 
         private static bool IsClientFlattenUsed(Dictionary<string, object> extensions)
             => extensions.ContainsKey(ClientFlattenExtensionName)
-            && extensions[ClientFlattenExtensionName] is bool
-            && (bool)extensions[ClientFlattenExtensionName] == true;
+            && extensions[ClientFlattenExtensionName] is bool;
```
veronicagg: could you add a valid test example for this rule with explicit x-ms-client-flatten set to false? could you also add a comment somewhere here, to indicate that not checking on the value is on purpose?

Issue #1849, opened 2017-02-16
C# codegen generates invalid code for int array parameters


kurt-m:
The following parameter definition:

```json
 "parameters": [
    {
      "name": "ids",
      "in": "query",
      "description": "The ids to filter by",
      "required": false,
      "type": "array",
      "items": {
        "type": "integer",
        "format": "int32"
      },
      "collectionFormat": "multi"
    }
  ]
```

Generates the following invalid C# code:

```cs
if (ids != null)
{
    ids.Count == 0)
    {
        _queryParameters.Add(string.Format("ids={0}", System.Uri.EscapeDataString(string.Empty)));
    }
    else
    {
        foreach (var _item in ids)
        {
            _queryParameters.Add(string.Format("ids={0}", System.Uri.EscapeDataString(_item ?? string.Empty)));
        }
    }
}
```

This results in a compiler error **CS0019 - Operator '??' cannot be applied to operands of type 'int?' and 'string'**

olydis:
I think AutoRest only supports string collection as query/path parameters right now.

olydis:
fixed by above PR

PR #2208, opened 2017-04-27
fix 2206, 1900, 1849

### LINKED ISSUE #1900: Operator '??' cannot be applied to operands of type 'Guid' and 'string'
When generating C# against an API that takes an array of guid on the query string, autorest outputs code that fails to compile

```C#
foreach (var _item in ids)
{
    _queryParameters.Add(string.Format("ids={0}", System.Uri.EscapeDataString(_item ?? string.Empty)));
}
```

**_item ?? string.Empty**  causes 

> CS0019 Operator '??' cannot be applied to operands of type 'Guid' and 'string'

Seems to be the same root issue as #1648 

The below swagger causes a repro on 1.0.0-Nightly20170212

``` javascript
{
    "swagger": "2.0",
    "info": {
        "version": "1.0",
        "title": "title"
    },
    "host": "localhost",
    "schemes": ["http"],
    "paths": {
        "/test": {
            "delete": {
                "tags": ["test"],
                "operationId": "deleteTest",
                "consumes": [],
                "produces": [],
                "parameters": [{
                    "name": "ids",
                    "in": "query",
                    "required": true,
                    "type": "array",
                    "items": {
                        "type": "string",
                        "format": "uuid",
                        "x-nullable": false
                    },
                    "collectionFormat": "multi",
                    "x-nullable": true
                }],
                "responses": {
                    "204": {
                        "description": "No Content"
                    }
                }
            }
        },
    },
    "definitions": {

    }
}
```

### LINKED ISSUE #2206: don't reuse modeler output for all generators
...since some settings you might wanna have different per generation pass already flow into the code model. E.g., having `namespace: ...` in the `java` settings will not make it to the generated Java code, since the namespace is already consumed by the modeler.

olydis:
- fixes #2206 as pointed out by @NiklasGustafsson
- `multi` collection format with types other than string (fixes #1900 and #1849)

### DIFF
--- a/Samples/1b-code-generation-multilang/CSharp/ISwaggerPetstore.cs
+++ b/Samples/1b-code-generation-multilang/CSharp/ISwaggerPetstore.cs
@@ -2,7 +2,7 @@
 // Changes may cause incorrect behavior and will be lost if the code is
 // regenerated.
 
-namespace Petstore
+namespace CSharpNamespace
 {
     using Microsoft.Rest;
     using Models;
--- a/Samples/1b-code-generation-multilang/CSharp/Models/Error.cs
+++ b/Samples/1b-code-generation-multilang/CSharp/Models/Error.cs
@@ -2,7 +2,7 @@
 // Changes may cause incorrect behavior and will be lost if the code is
 // regenerated.
 
-namespace Petstore.Models
+namespace CSharpNamespace.Models
 {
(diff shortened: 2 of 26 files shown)


olydis:
looks like a transient error in Jenkins... will merge #2203 as I'm scared someone else will merge first and create a conflict.

PR #2209, opened 2017-04-27
[C#] Use a different constructor for System.Net.Http.HttpRequestMessage

brywang-msft:
Small change to use a different constructor for System.Net.Http.HttpRequestMessage. When setting System.Net.Http.HttpRequestMessage.RequestUri in PowerShell Core Alpha12+ in an assembly compiled by the Add-Type cmdlet, there's a problem with type redirection that results in an error calling the generated operation method.

How verified:
gulp build
gulp test
gulp regenerate
gulp fix-line-endings (I ended up with a bunch of line ending changes still, so I only added the .cs files. The .cs files might have borked line endings, though, so let me know if I need to manually fix them)
Tested Azure and non-Azure C# SDKs on PowerShell Core 6.0.0-alpha18. Failed before change, worked after.

### DIFF
--- a/Samples/1a-code-generation-minimal/Client/SwaggerPetstore.cs
+++ b/Samples/1a-code-generation-minimal/Client/SwaggerPetstore.cs
@@ -184,10 +184,8 @@ namespace Petstore
             var _baseUrl = BaseUri.AbsoluteUri;
             var _url = new System.Uri(new System.Uri(_baseUrl + (_baseUrl.EndsWith("/") ? "" : "/")), "pet").ToString();
             // Create HTTP transport objects
-            var _httpRequest = new HttpRequestMessage();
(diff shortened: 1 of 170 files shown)


brywang-msft:
Got a test failure, figured I should check with you guys before I change the expected result. The errors seem basically the same:

Test: AutoRest.CSharp.Azure.Tests.AcceptanceTests.PagingSadPathTests
Expected: typeof(System.UriFormatException)
Actual: typeof(System.InvalidOperationException): An invalid request URI was provided. The request URI must either be an absolute URI or BaseAddress must be set.

olydis:
@brywang-msft changing the expected result is good with me

olydis:
Closing due to inactivity. Feel free to reopen if this is still important to you.

PR #2211, opened 2017-04-27
Added a more clear message for Security Definitions Validation

sarangan12:
@vishrutshah @veronicagg @salameer Could you please review and approve?

### DIFF
--- a/src/core/AutoRest.Core/Properties/Resources.Designer.cs
+++ b/src/core/AutoRest.Core/Properties/Resources.Designer.cs
@@ -557,7 +557,7 @@ namespace AutoRest.Core.Properties {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to A swagger must have security definitions and must adhere to the specific structure..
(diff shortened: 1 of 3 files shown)


src/core/AutoRest.Core/Properties/Resources.Designer.cs
```diff
@@ -557,7 +557,7 @@ public class Resources {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to A swagger must have security definitions and must adhere to the specific structure..
+        ///   Looks up a localized string similar to Every swagger/configuration must have a security definitions section and it must adhere to the structure described in: https://github.com/Azure/autorest/tree/master/docs/developer/validation-rules/security-definitions-structure-validation.md.
```
vishrutshah: curious: are we planning to add link to each of the rules in the message?

vishrutshah:
@sarangan12 Also link the bug if we have associated with this change in the PR `description`

src/core/AutoRest.Core/Properties/Resources.Designer.cs
sarangan12: If required, yes

PR #2212, opened 2017-04-27
Small bugfix in Go generator for LROs.

jhendrixMSFT:
Fixed a missed case in the recent LRO feature work.


jhendrixMSFT:
@marstr can you take a quick look I missed a case when I made changes from feedback.

PR #2220, opened 2017-04-29
Fixes golint issue

mcardosos:
Fixed golint issue in CI in PR https://github.com/Azure/azure-sdk-for-go/pull/599

[diff](https://github.com/Azure/azure-sdk-for-go/pull/599/commits/049841ada1c46aa438542e6d0b39dd0890e33726) when regenerating with this changes.

PTAL @marstr @jhendrixMSFT

### DIFF
--- a/src/generator/AutoRest.Go/CodeNamerGo.cs
+++ b/src/generator/AutoRest.Go/CodeNamerGo.cs
@@ -37,43 +37,43 @@ namespace AutoRest.Go
         // CommonInitialisms are those "words" within a name that Golint expects to be uppercase.
         // See https://github.com/golang/lint/blob/master/lint.go for detail.
         private string[] CommonInitialisms => new string[] {
-                                                            "Acl",
-                                                            "Api",
-                                                            "Ascii",
-                                                            "Cpu",
-                                                            "Css",
-                                                            "Dns",
-                                                            "Eof",
-                                                            "Guid",
-                                                            "Html",
-                                                            "Http",
-                                                            "Https",
-                                                            "Id",
-                                                            "Ip",
-                                                            "Json",
-                                                            "Lhs",
-                                                            "Qps",
-                                                            "Ram",
-                                                            "Rhs",
-                                                            "Rpc",
-                                                            "Sla",
-                                                            "Smtp",
-                                                            "Sql",
-                                                            "Ssh",
-                                                            "Tcp",
-                                                            "Tls",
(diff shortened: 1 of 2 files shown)


src/generator/AutoRest.Go/CodeNamerGo.cs
```diff
@@ -37,43 +37,43 @@ public new static CodeNamerGo Instance
         // CommonInitialisms are those "words" within a name that Golint expects to be uppercase.
         // See https://github.com/golang/lint/blob/master/lint.go for detail.
         private string[] CommonInitialisms => new string[] {
```
jhendrixMSFT: Probably should change this to a HashSet<string> that uses StringComparer.OrdinalIgnoreCase for its IEqualityComparer.

src/generator/AutoRest.Go/CodeNamerGo.cs
```diff
@@ -418,9 +418,27 @@ public void ReserveNamespace(string ns)
...
             {
-                builder.Append(CommonInitialisms.Contains(s) ? s.ToUpper() : s);
+                string word = words[i];
+                if (CommonInitialisms.Contains(word.ToLower()))
+                {
+                    word = word.ToUpper();
+                }
+                else
+                {
+                    if (i < words.Length-1)
```
jhendrixMSFT: This if block can be merged into the preceding else clause.

src/generator/AutoRest.Go/CodeNamerGo.cs
```diff
@@ -418,9 +418,27 @@ public void ReserveNamespace(string ns)
...
+                    {
+                        var concat = words[i] + words[i+1];
```
jhendrixMSFT: It would be good to add a comment explaining why this is happening.

src/generator/AutoRest.Go/CodeNamerGo.cs
mcardosos: Fixed

src/generator/AutoRest.Go/CodeNamerGo.cs
mcardosos: Fixed

src/generator/AutoRest.Go/CodeNamerGo.cs
mcardosos: Fixed

Issue #2225, opened 2017-05-01
[Linter] Adding Validation rule category to indicate service-side breaking change impact


dsgouda:
Per offline discussions, it would be helpful if we could assign a category to validation rules that would indicate if adhering to/implementing the suggestions provided by our validation rules would cause a breaking change at the service end (most changes recommended by validation rules would actually cause some sort of sdk-breaking change. Moreover, for non-breaking changes like adding features to an existing service would get incorrectly categorized as sdk-breaking changes) 

Tagging @veronicagg

PR #2230, opened 2017-05-02
Adding new categorization to indicate breaking changes

dsgouda:
Addressing issue https://github.com/Azure/autorest/issues/2225

Validation rules which may or may not need this categorization 
[M3006](https://github.com/Azure/autorest/blob/204c91d68799bc5cd074ba75107573a37012efe4/src/modeler/AutoRest.Swagger/Validation/BodyTopLevelProperties.cs), [M3017](https://github.com/Azure/autorest/blob/a2d51c88366b683e1ca92cbeceb91c16b4ffdb6c/src/modeler/AutoRest.Swagger/Validation/GuidValidation.cs), [S2004](https://github.com/Azure/autorest/blob/a2d51c88366b683e1ca92cbeceb91c16b4ffdb6c/src/modeler/AutoRest.Swagger/Validation/NonAppJsonTypeWarning.cs) and [M2014](https://github.com/Azure/autorest/blob/6cee6087d075d9969dfd233a2ade096ccbabd464/src/modeler/AutoRest.Swagger/Validation/OperationParametersValidation.cs)


src/modeler/AutoRest.Swagger/Validation/Core/ValidationChangesImpact.cs
```diff
@@ -0,0 +1,15 @@
...
+using System;
+
+namespace AutoRest.Swagger.Validation.Core
+{
+    [Flags]
+    public enum ValidationChangesImpact
+    {
+        None = 1 << 0,
+        PotentialServiceBreakingChanges = 1 << 1,
+        NonBreakingServiceChanges = 1 << 2
```
veronicagg: should we call it NoServiceChangesRequired? or SdkImpacting or PotentialSdkChanges? Thinking about this, it seems like every property should have this property set, so a rule can impact the Service, or the SDK or nothing - but I wouldn't expect any rule to have no impact (as there wouldn't be any benefit on having the rule). 
Another option could be to name them "ServiceLevelImpacting"/"SdkLevelImpacting" or just "ServiceImpact"/"SdkImpact". 
@kirthik thoughts?

src/modeler/AutoRest.Swagger/Validation/OperationsAPIImplementationValidation.cs
```diff
@@ -40,6 +40,11 @@ public class OperationsAPIImplementationValidation : TypedRule<Dictionary<string
         public override Category Severity => Category.Error;
 
         /// <summary>
+        /// What kind of change implementing this rule can cause.
+        /// </summary>
+        public override ValidationChangesImpact ValidationChangesImpact => ValidationChangesImpact.NonBreakingServiceChanges;
```
veronicagg: this is one where I'm not sure we want to categorize between breaking or non-breaking, maybe having "potential service impact" and "potential sdk impact" could work better and in that case, I think this would fall into "potential service impact" even though it may not be breaking.
@kirthik would you want this one to show up in reports as service impacting?

src/modeler/AutoRest.Swagger/Validation/Core/Rule.cs
```diff
@@ -27,6 +27,11 @@ protected Rule()
         public virtual ValidationCategory ValidationCategory => ValidationCategory.None; // !!! implement me and make me abstract !!!
 
         /// <summary>
+        /// What kind of change implementing this rule can cause.
+        /// </summary>
+        public virtual ValidationChangesImpact ValidationChangesImpact => ValidationChangesImpact.None;
```
veronicagg: Not really related to this change, but as we're at this, should we add another Rule property for "RuleInfo" or documentation, so we can provide a link to extra documentation?

src/modeler/AutoRest.Swagger/Validation/Core/ValidationChangesImpact.cs
dsgouda: Definitely should not be ```NoServiceChangesRequired``` since the service itself needs to be changed so as to expose say a new end point for PATCH operation

src/modeler/AutoRest.Swagger/Validation/Core/Rule.cs
dsgouda: I'm open to the suggestion, how do you think these 'links' would look like, urls to the repo or relative paths to the corresponding .md file

veronicagg: I'm thinking links could be fwlinks, so we don't need to update the code if the link changes, I'd like to hear @kirthik thoughts on this as well. If not, it could also be a path to an top level page, that doesn't change frequently and would have a list of all rules with links to each specific one.

Issue #2172, opened 2017-04-18
M3027 (ListBySuscription and ListByResourceGroup) should be only applicable to only tracked top level resources


ravbhatnagar:

sarangan12:
Task completed.

PR #2233, opened 2017-05-02
Code change to exclude CRs from List By RG/Subscription rules

sarangan12:
There are atleast 3 changes suggested to Tracked Resources validation. This PR covers one of them. What does this PR do?

Currently, we apply the ListByResourceGroup and ListBySubscription rules on both parent and child resources. This is incorrect. We must apply the rules only on the parent resources. 

No regeneration changes.

@veronicagg @vishrutshah @dsgouda @salameer Please approve

Related to https://github.com/Azure/autorest/issues/2172

### DIFF
--- a/src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
+++ b/src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
@@ -505,6 +505,13 @@ namespace AutoRest.Swagger.Model.Utilities
             return pathToEvaluate.Substring(pathToEvaluate.LastIndexOf("/") + 1);
         }
 
+        /// <summary>
+        /// Gets the actual parent resource name. For example, the name in Path could be 'servers'. The actual parent name is 'server'.
+        /// </summary>
+        /// <param name="nameInPath"></param>
+        /// <param name="paths"></param>
+        /// <param name="definitions"></param>
+        /// <returns></returns>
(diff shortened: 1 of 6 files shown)


src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -457,5 +457,140 @@ public static IEnumerable<string> GetResourceProviders(Dictionary<string, Dictio
...
+        /// <param name="childResourcePaths">paths that fits the child resource criteria</param>
+        /// <returns>name of the immediate parent resource name</returns>
+        private static string GetImmediateParentResourceName(string pathToEvaluate)
+        {
+            pathToEvaluate = pathToEvaluate.Substring(0, pathToEvaluate.LastIndexOf("/{"));
+            pathToEvaluate = pathToEvaluate.Substring(0, pathToEvaluate.LastIndexOf("/{"));
+            return pathToEvaluate.Substring(pathToEvaluate.LastIndexOf("/") + 1);
+        }
+
+        private static string GetActualParentResourceName(string nameInPath, Dictionary<string, Dictionary<string, Operation>> paths, Dictionary<string, Schema> definitions)
```
dsgouda: Can you add some comments for this method

sarangan12:
Added comment. Completed dry run. In cases, where child resource is defined and the error is reported, with the current code changes, it does not report any more.

PR #2237, opened 2017-05-03
THE PIPELINE (and some fixes)

### LINKED ISSUE #2236: AutoRest crashes for `application/xml` encoded responses in combination with `x-ms-parameter-grouping`.
**Background**:

I have an operation in webapi service like follows:
  ```
[Route("books")]
        [HttpGet]
        [ResponseType(typeof(GetBooksResponse))]
        public IHttpActionResult GetBooks([FromUri] GetBooksRequest request)
        {
                ...
        }
```

```
GetBooksRequest 
{
  Genre,
  Publisher,
  ...
}
```

Swagger spec:

```
 "parameters": [
         {
            "name": "Genre",
            "in": "query",
            "required": false,
            "type": "string"
          },
          {
            "name": "Publisher",
            "in": "query",
            "required": false,
            "type": "string"
          },
...
```

I would like to use **x-ms-parameter-grouping** to group method parameters Genre,Publisher, etc in auto generated client, and wrote a ParameterGroupOperationFilter

```
public class ParameterGroupOperationFilter : IOperationFilter
    {
        public void Apply(Operation operation, SchemaRegistry schemaRegistry, ApiDescription apiDescription)
        {
            if (operation.parameters == null)
                return;

            var parameters = apiDescription.ActionDescriptor.GetParameters();
            if (parameters?.Count == 1))
            {
                foreach (var parm in operation.parameters)
                {
                    if ((parm.@in).Equals("path") || (parm.@in).Equals("query"))
                    {
                        parm.vendorExtensions.Add("x-ms-parameter-grouping",
                                            new { name = parameters[0].ParameterType.Name + "parameter" });
                    }

                }
                   
            }
        }
    }
```

When I use this operationfilter in Swagger Config, and apply AutoRest 1.0.1.0, it throws an error:
 
image

BTW, when I use AutoRest 0.17.0.0, it works fine.

**Question**: Is there anything missing in ParameterGroupOperationFilter? What is that error about?

### PR DESCRIPTION
**Pull request by olydis**
- fully abstract, configurable pipeline
- fixes #2236 (`application/xml` + `x-ms-parameter-grouping` = crash)
- legacy CLI: opt out client-side-validation by default
- allow `=` in value part of CLI param

### DIFF
--- a/Samples/3b-custom-transformations/generated/pipeline.yaml
+++ b/Samples/3b-custom-transformations/generated/pipeline.yaml
@@ -1,0 +1,88 @@
+---
+pipeline-emitter:
+  pluginName: pipeline-emitter
+  configScope:
+    - scope-pipeline-emitter
+  inputs: []
+swagger-document/loader:
+  pluginName: loader
+  configScope: []
+  inputs: []
+swagger-document/individual/transform:
+  pluginName: transform
+  configScope: []
+  inputs:
+    - swagger-document/loader
+swagger-document/compose:
+  pluginName: compose
+  configScope: []
+  inputs:
+    - swagger-document/individual/transform
+swagger-document/transform:
+  pluginName: transform
+  configScope: []
+  inputs:
+    - swagger-document/compose
+swagger-document/emitter:
+  pluginName: emitter
+  configScope:
+    - scope-swagger-document/emitter
+  inputs:
+    - swagger-document/transform
+swagger-document/azure-validator:
+  pluginName: azure-validator
+  configScope:
+    - azure-validator
+  inputs:
+    - swagger-document/transform
+csharp/modeler:
+  pluginName: modeler
+  configScope:
+    - csharp
+  inputs:
+    - swagger-document/transform
+csharp/commonmarker:
+  pluginName: commonmarker
+  configScope:
+    - csharp
+  inputs:
+    - csharp/modeler
+csharp/cm/transform:
+  pluginName: transform
+  configScope:
+    - csharp
+  inputs:
+    - csharp/commonmarker
+csharp/cm/emitter:
+  pluginName: emitter
+  configScope:
+    - csharp
+    - scope-cm/emitter
+  inputs:
+    - csharp/cm/transform
+csharp/generate:
+  pluginName: csharp
+  configScope:
+    - csharp
+  inputs:
+    - swagger-document/transform
+    - csharp/cm/transform
+csharp/simplifier:
+  pluginName: csharp-simplifier
+  configScope:
+    - csharp
+  inputs:
+    - csharp/generate
+csharp/transform:
+  pluginName: transform
+  configScope:
+    - csharp
+  inputs:
+    - csharp/simplifier
+csharp/emitter:
+  pluginName: emitter
+  configScope:
+    - csharp
+    - scope-csharp/emitter
+  inputs:
+    - csharp/transform
--- a/src/autorest-core/app.ts
+++ b/src/autorest-core/app.ts
@@ -130,7 +130,7 @@ function parseArgs(autorestArgs: string[]): CommandLineArgs {
   };
 
   for (const arg of autorestArgs) {
-    const match = /^--([^=]+)(=([^=]+))?$/g.exec(arg);
+    const match = /^--([^=]+)(=(.+))?$/g.exec(arg);
 
     // configuration file?
     if (match === null) {
--- a/src/autorest-core/lib/configuration.ts
+++ b/src/autorest-core/lib/configuration.ts
@@ -219,7 +219,9 @@ export class ConfigurationView {
 
   public * GetPluginViews(pluginName: string): Iterable<ConfigurationView> {
     for (const section of ValuesOf<any>((this.config as any)[pluginName])) {
-      yield new ConfigurationView(this.messageEmitter, this.configFileFolderUri, section, this.config);
+      if (section) {
+        yield new ConfigurationView(this.messageEmitter, this.configFileFolderUri, section === true ? {} : section, this.config);
+      }
     }
   }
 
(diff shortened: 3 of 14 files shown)

### CODE REVIEW

File: src/autorest-core/app.ts [hunk 1]
```diff
@@ -130,7 +130,7 @@ function parseArgs(autorestArgs: string[]): CommandLineArgs {
   };
 
   for (const arg of autorestArgs) {
-    const match = /^--([^=]+)(=([^=]+))?$/g.exec(arg);
+    const match = /^--([^=]+)(=(.+))?$/g.exec(arg);
```
olydis: allowing `=` in value part of CLI arg, e.g. `--input-file=https://asd/qwe.json?token=flubber` will be parsed correctly. The key/value splitting will always happen at the first `=` which should be fine as keys should never contain that character.

src/core/AutoRest.Core/Settings.cs
```diff
@@ -73,8 +73,6 @@ public Settings()
             ValidationLevel = Category.Error;
             ModelsName = "Models";
             CodeGenerationMode = "rest-client";
-
-            CustomSettings.Add("ClientSideValidation", true);
```
olydis: handled below instead (`src/core/AutoRest/Program.cs`), unfortunately the legacy CLI parsing mechanism overwrites `CustomSettings` completely, such that this line had no effect.

src/core/AutoRest.Extensions/ParameterGroupExtensionHelper.cs
```diff
@@ -35,6 +35,7 @@ private static Property CreateParameterGroupProperty(Parameter parameter)
                 //Constraints = parameter.Constraints, Omit these since we don't want to perform parameter validation
                 Documentation = parameter.Documentation,
                 ModelType = parameter.ModelType,
+                RealPath = new string[] { },
```
olydis: fixes ArgNullEx when using both `x-ms-parameter-grouping` and media type `application/xml` in the same operation (#2236)

Samples/3b-custom-transformations/generated/pipeline.yaml
```diff
@@ -0,0 +1,88 @@
+---
+pipeline-emitter:
```
olydis: pipeline representation is fully serializable, think we can create a nice diagnostic "plugin" that'll visualize this DAG and show live information about it!

src/autorest-core/lib/configuration.ts
```diff
@@ -219,7 +219,9 @@ export class ConfigurationView {
 
   public * GetPluginViews(pluginName: string): Iterable<ConfigurationView> {
     for (const section of ValuesOf<any>((this.config as any)[pluginName])) {
-      yield new ConfigurationView(this.messageEmitter, this.configFileFolderUri, section, this.config);
+      if (section) {
```
olydis: ...so `--csharp=false` doesn't count as a C# plugin scope

src/autorest-core/lib/pipeline/pipeline.ts
```diff
@@ -94,9 +102,130 @@ function CreateCommonmarkProcessor(): PipelinePlugin {
     }
   };
 }
+function CreateArtifactEmitter(inputOverride?: () => Promise<DataStoreViewReadonly>): PipelinePlugin {
+  return async (config, input, working, output) => {
```
olydis: emitting as just another pipeline step (accepting arbitrary input but emitting no output)

src/autorest-core/lib/pipeline/pipeline.ts
```diff
@@ -94,9 +102,130 @@ function CreateCommonmarkProcessor(): PipelinePlugin {
...
-  const barrier = new OutstandingTaskAwaiter();
+function* WithIndex<T>(collection: Iterable<T>): Iterable<[T, number]> {
+  let idx = 0;
+  for (const x of collection) {
+    yield [x, idx++];
+  }
+}
+
+function BuildPipeline(config: ConfigurationView): { pipeline: { [name: string]: PipelineNode }, configs: { [jsonPath: string]: ConfigurationView } } {
+  const cfgPipeline = config.GetEntry("pipeline" as any);
```
olydis: derive pipeline graph entirely from configuration!
Will add more and more convenience into that, currently one would have to rewire the graph to add an own pipeline step.
Example: Inserting a Go simplifier
``` yaml
pipeline:
  # new step:
  go/simplify:
    plugin: go-simplifier
    input: transform # formerly consumed by go/emitter
    output-artifact: source-file-go
  # rewire emitter to consume the above
  go/emitter:
    input: simplify # was: transform
    scope: scope-go/emitter
```

src/autorest-core/lib/pipeline/pipeline.ts
```diff
@@ -153,56 +308,16 @@ export async function RunPipeline(config: ConfigurationView, fileSystem: IFileSy
...
-  }
+  // schedule pipeline
+  const tasks: { [name: string]: Promise<DataStoreViewReadonly> } = {};
+  const getTask = (name: string) => name in tasks ?
+    tasks[name] :
+    tasks[name] = (async () => RunPlugin(name))();
 
-  if (config.GetEntry("model-validator")) {
-    barrier.Await(RunPlugin(config, "model-validator", scopeComposedSwaggerTransformed));
-  }
```
olydis: fixed pipeline goodbye!

src/autorest-core/lib/pipeline/pipeline.ts
```diff
@@ -121,23 +250,49 @@ export async function RunPipeline(config: ConfigurationView, fileSystem: IFileSy
...
   };
 
   // TODO: think about adding "number of files in scope" kind of validation in between pipeline steps
 
-  // to be replaced with scheduler
-  let scopeCtr = 0;
-  const RunPlugin: (config: ConfigurationView, pluginName: string, inputScope: DataStoreViewReadonly) => Promise<DataStoreViewReadonly> =
-    async (config, pluginName, inputScope) => {
+  const RunPlugin: (nodeName: string) => Promise<DataStoreViewReadonly> =
+    async (nodeName) => {
```
olydis: schedules the node with given name from the pipeline graph

src/autorest-core/resources/default-configuration.md
```diff
@@ -27,18 +27,339 @@ plugins:
...
+pipeline:
+  pipeline-emitter: # emits the pipeline graph
+    scope: scope-pipeline-emitter
+```
+
+#### Loading
+
+``` yaml
+pipeline:
+  swagger-document/loader:
```
olydis: abstract representation of the pipeline. contains some duplication, will make that more and more convenient (goes hand in hand with making pipeline customization easier)

PR #2240, opened 2017-05-03
Go test generator.

jhendrixMSFT:
This is the test generator for the Go SDK.  It uses a AST model instead of
Razor templates.  It will generate test methods to be invoked by the Go
SDK JSON-RPC server.


jhendrixMSFT:
@mcardosos @marstr FYI on the Go test generator PR.

jhendrixMSFT:
@olydis can you please take a quick look to ensure I've properly integrated this into AutoRest?  It works so I believe I have but maybe there's something I missed.

olydis:
> It works

How does it work? Looks to me as if it's only accessible using the legacy CLI, or how do you invoke this?

jhendrixMSFT:
Works via "gulp autorest", I guess that's the legacy CLI?

olydis:
https://github.com/Azure/autorest/blob/master/src/autorest-core/lib/pipeline/pipeline.ts#L178
https://github.com/Azure/autorest/blob/master/src/core/AutoRest/Plugins/Generator.cs#L55

olydis:
ah and add a line `- source-file-gotestgen` in https://github.com/Azure/autorest/blob/master/src/autorest-core/resources/default-configuration.md where you can see `- source-file-go` and such

olydis:
hope I didn't forget anything... but really, since you are the main consumers of this it's no big deal if it doesn't work with the new CLI, we can always fix that later and you can just use the CLI you're used to until we disable it -- which may still take a while ;)

jhendrixMSFT:
The name of the generator is "Go.TestGen.dll", is there anything I need to do to map this to "gotestgen"?

olydis:
nope, the change you made [here](https://github.com/Azure/autorest/pull/2240/files#diff-fcf84b1a855607cdd65c748d5220bdc8) should be sufficient

jhendrixMSFT:
Even with the '.' in the name?
```
Go.TestGen:
   TypeName: PluginGotg, AutoRest.Go.TestGen
```

olydis:
oooooohhhh shoot I missed that...
okay, basically you gotta make sure that in `Generator.cs` you get `language = "Go.TestGen"` whenever `codeGenerator == "gotestgen"`. 😄

jhendrixMSFT:
Would it be better to rename the generator to GoTestGen?  The current naming convention was based on other generators, e.g. "Azure.CSharp".  Presumably they would have the same problem.  Or is the pattern going to be to translate the names?

olydis:
well as you can see further down in `Generator.cs`, the `Azure` and `Fluent` things get appended depending on flags... wouldn't do the same for `TestGen`, so either make it a special case or indeed rename the assembly :-)

jhendrixMSFT:
Hmm well it could be the case that this pattern of a TestGen suffix proliferates to the other test generators.  Assuming this is the case could we make that appending work?

olydis:
fair point, sure thing!

Issue #2106, opened 2017-04-03
Json paths returned in rules need to be accurate to the last leaf level where the error occurs


veronicagg:
Here are some examples of rule violations, which are not accurate to the last level:
https://travis-ci.org/Azure/azure-rest-api-specs/jobs/217096049#L2664 
https://travis-ci.org/Azure/azure-rest-api-specs/jobs/217096049#L1269
https://travis-ci.org/Azure/azure-rest-api-specs/jobs/217096049#L1518
https://travis-ci.org/Azure/azure-rest-api-specs/jobs/217096049#L1734
https://travis-ci.org/Azure/azure-rest-api-specs/jobs/217096049#L2759
https://travis-ci.org/Azure/azure-rest-api-specs/jobs/217096049#L3609
https://travis-ci.org/Azure/azure-rest-api-specs/jobs/217096049#L3719
https://travis-ci.org/Azure/azure-rest-api-specs/jobs/217096049#L3741
https://travis-ci.org/Azure/azure-rest-api-specs/jobs/217096049#L4268
https://travis-ci.org/Azure/azure-rest-api-specs/jobs/217096049#L5342

veronicagg:
@sarangan12 I'm assigning this one to you, it involves many rules, so feel free to ask for help as needed. I opened only one issue to track this (at least for now) because it may make sense to update these paths in bulk. If you think it'd be better to split, feel free to do so.

olydis:
Not sure if this is tracked somewhere already but it looks like the following rules report invalid source locations:

| Rule | Example of reported path |
|-----|-----|
| `TrackedResourceListByImmediateParent` | `$.paths["#/definitions/runs"]` (whoops) |
| `SecurityDefinitionsStructureValidation ` | `$.securityDefinitions` (can't blame that location if it doesn't exist) |
| `LongRunningResponseValidationRule ` | `$.<...>.extensions["x-ms-long-running-operation"]` |
| `NextLinkPropertyMustExist ` | `$.<...>.extensions["x-ms-pageable"]` |

The last two are a bit mean, i.e. "it's not your fault": The .NET part of AutoRest parses extensions (all stuff starting with `x-`) into a dictionary `extensions`, so the object tree looks different from the original Swagger! In other words, in the last example above, it should really have complained at `$.<...>["x-ms-pageable"]` because there *is no `extensions` property* in the Swagger. I'll look into fixing that.

olydis:
[...]

PR #2242, opened 2017-05-03
Improving paths reported with validation messages

dsgouda:
Addresses issue https://github.com/Azure/autorest/issues/2106
Minor code cleanup and improved messaging


src/modeler/AutoRest.Swagger/Validation/BodyPropertiesNamesCamelCase.cs
```diff
@@ -60,7 +60,9 @@ public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<
                             {
                                 if (!ValidationUtilities.IsODataProperty(prop.Key) && !ValidationUtilities.IsNameCamelCase(prop.Key))
                                 {
-                                    yield return new ValidationMessage(new FileObjectPath(context.File, context.Path), this, prop.Key, ValidationUtilities.GetCamelCasedSuggestion(prop.Key));
+                                    yield return new ValidationMessage(new FileObjectPath(context.File, 
+                                        context.Path.AppendProperty(operation).AppendProperty("parameters").AppendIndex(path[operation].Parameters.IndexOf(param)).AppendProperty("schema").AppendProperty("properties").AppendProperty(prop.Key)),
```
olydis: I'd switch the `foreach` loop to a `for` loop rather than using `IndexOf`! That's a cheap optimization and less error prone.

src/modeler/AutoRest.Swagger/Validation/ListOperationNamingWarning.cs
```diff
@@ -40,13 +40,14 @@ public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<
                     continue;
                 }
 
+                var violatingPath = ValidationUtilities.GetOperationIdPath(opPair.Value.OperationId, entity).Key;
                 if (ValidationUtilities.IsXmsPageableResponseOperation(opPair.Value))
                 {
-                    yield return new ValidationMessage(new FileObjectPath(context.File, context.Path), this, opPair.Value.OperationId, XmsPageableViolation);
+                    yield return new ValidationMessage(new FileObjectPath(context.File, context.Path.AppendProperty(violatingPath).AppendProperty(opPair.Key).AppendProperty(opPair.Value.OperationId)), this, opPair.Value.OperationId, XmsPageableViolation);
```
olydis: think you wanna append `"operationId"`, not the actual operation ID ;-)

src/modeler/AutoRest.Swagger/Validation/ListOperationNamingWarning.cs
```diff
@@ -40,13 +40,14 @@ public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<
...
                 }
                 else if (ValidationUtilities.IsArrayTypeResponseOperation(opPair.Value, serviceDefinition))
                 {
-                    yield return new ValidationMessage(new FileObjectPath(context.File, context.Path), this, opPair.Value.OperationId, ArrayTypeViolation);
+                    yield return new ValidationMessage(new FileObjectPath(context.File, context.Path.AppendProperty(violatingPath).AppendProperty(opPair.Key).AppendProperty(opPair.Value.OperationId)), this, opPair.Value.OperationId, ArrayTypeViolation);
```
olydis: same as above

src/modeler/AutoRest.Swagger/Validation/ArmResourcePropertiesBag.cs
```diff
@@ -49,7 +49,7 @@ public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<
             foreach (var violatingModel in violatingModels)
             {
                 var violatingProperties = definitions[violatingModel].Properties.Keys.Union(definitions[violatingModel].Properties["properties"].Properties.Keys).Intersect(ArmPropertiesBag);
-                yield return new ValidationMessage(new FileObjectPath(context.File, context.Path), this, violatingModel, 
+                yield return new ValidationMessage(new FileObjectPath(context.File, context.Path.AppendProperty(violatingModel).AppendProperty("properties")), this, violatingModel,
```
veronicagg: not related to this change, but the comment for "GetValidationMessages" seems not to represent what this does, it'd be great to fix it as you're modifying this file.

src/modeler/AutoRest.Swagger/Validation/GuidValidation.cs
```diff
@@ -39,37 +39,27 @@ public class GuidValidation : TypedRule<Dictionary<string, Schema>>
...
 
-        /// <summary>
-        /// An <paramref name="definitions"/> fails this rule if one of the property has GUID,
-        /// i.e. if the type of the definition is string and the format is uuid.
-        /// </summary>
-        /// <param name="definitions">Operation Definitions to validate</param>
-        /// <param name="formatParameters">The noun to be put in the failure message</param>
-        /// <returns>true if there is no GUID. false otherwise.</returns>
-        public override bool IsValid(Dictionary<string, Schema> definitions, RuleContext context, out object[] formatParameters)
+        public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string, Schema> definitions, RuleContext context)
```
veronicagg: why are we removing the comments for the method? should they be updated?

src/modeler/AutoRest.Swagger/Validation/PutGetPatchResponseValidation.cs
```diff
@@ -39,51 +39,27 @@ public class PutGetPatchResponseValidation : TypedRule<Dictionary<string, Dictio
...
         public override Category Severity => Category.Error;
 
-        /// <summary>
-        /// Validates if the response of Put/Get/Patch are same.
-        /// </summary>
-        /// <param name="paths">paths to validate</param>
-        /// <param name="formatParameters">The noun to be put in the failure message</param>
-        /// <returns>true if the response is same for Put/Get/Patch responses.false otherwise.</returns>
-        public override bool IsValid(Dictionary<string, Dictionary<string, Operation>> paths, RuleContext context, out object[] formatParameters)
+        private static readonly IEnumerable<string> OpList = new List<string>() { "put", "get", "patch" };
```
veronicagg: could you update the comments instead of removing them?

veronicagg: is the path from using "IsValid" ever correct? I wonder how useful that method is and/or if it should be updated to receive a path.

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
```diff
@@ -55,7 +55,7 @@ public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<
                     formatParameters[0] = childResourceMapping.Key;
                     formatParameters[1] = childResourceMapping.Value;
 
-                    yield return new ValidationMessage(new FileObjectPath(context.File, context.Path.AppendProperty("#/definitions/" + childResourceMapping.Key)), this, formatParameters);
+                    yield return new ValidationMessage(new FileObjectPath(context.File, context.Parent.Path.AppendProperty("definitions").AppendProperty(childResourceMapping.Value)), this, formatParameters);
```
veronicagg: is it value or key in this case?

src/modeler/AutoRest.Swagger/Validation/XmsExamplesProvidedValidation.cs
```diff
@@ -90,7 +90,7 @@ public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<
                 foreach (var violatingOp in violatingOps)
                 {
                     var violatingPath = paths.First(pathObj => pathObj.Value.Values.Select(op => op.OperationId).Contains(violatingOp.Value.OperationId)).Key;
-                    yield return new ValidationMessage(new FileObjectPath(context.File, context.Path.AppendProperty(violatingPath).AppendProperty(violatingOp.Key).AppendProperty("operationId")), this, string.Format(OperationIdMessageSuffix, violatingOp.Value.OperationId));
+                    yield return new ValidationMessage(new FileObjectPath(context.File, context.Path.AppendProperty(violatingPath).AppendProperty(violatingOp.Key)), this, string.Format(OperationIdMessageSuffix, violatingOp.Value.OperationId));
```
veronicagg: what does the "key" return if you don't append the "operationdId"?

dsgouda: It points to the Operation node where the problem is.

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListByImmediateParent.cs
dsgouda: Whoops should be key

src/modeler/AutoRest.Swagger/Validation/PutGetPatchResponseValidation.cs
dsgouda: Not sure I understand completely, ```GetValidationMessages``` does return the exact path which causes the problem.

src/modeler/AutoRest.Swagger/Validation/XmsExamplesProvidedValidation.cs
veronicagg: ok, that's not bad either, it's quite equivalent in that case I think.

PR #2245, opened 2017-05-04
Updating Go Generator to allow UnixTime support.

marstr:
Updating the generator to take advantage of UnixTime support in go-autorest.


marstr:
I tested this by regenerating the v10 release and playing with it in the sample that I'm writing.

jhendrixMSFT:
What does the delta look like in the SDK, do you have a sample?

PR #2248, opened 2017-05-05
Add the .NET test generator.

jhendrixMSFT:
This is the .NET test generator.  It will create a mock .NET SDK that
translates API calls into JSON-RPC to be passed to test SDK servers (e.g.
an SDK created by the Go.TestGen generator).


jhendrixMSFT:
@olydis can you please take a look at the integration with CLI.
@sergey-shandar FYI here's the PR for the .NET test SDK generator.  Note that the name has changed slightly to "Azure.JsonRpcClient" (was "Azure.CSharp.JsonRpcClient").

olydis:
@jhendrixMSFT adjusted pipeline graph definition, should work like that for new CLI

jhendrixMSFT:
@olydis Boy I got that wrong, thanks for fixing it!

Issue #2232, opened 2017-05-02
climb up object path if node not found when resolving source location


olydis:
see https://github.com/Azure/autorest/issues/2106#issuecomment-298757838, thanks @dsgouda

That would improve UX in case of a buggy validation tool. I'd actually also emit a location-less warning then that the reported location wasn't found, so at least we see that something is wrong.

dsgouda:
Will try to get to this myself at some point, should be a good starting point in understanding the mysteries of AutoRest Core

PR #2249, opened 2017-05-05
Climb up the path when a node is not found

dsgouda:
Maybe a verbose function given its my maiden typescript PR


olydis:
your Jenkins job hit [this](https://github.com/Azure/autorest/issues/2217) issue, hence I removed the emoji from your PR description 😉

src/autorest-core/lib/data-store/data-store.ts
```diff
@@ -320,8 +321,28 @@ export class DataStore extends DataStoreView {
     return Object.getOwnPropertyNames(this.store);
   }
 
-  public Blame(absoluteUri: string, position: SmartPosition): BlameTree {
+  public Blame(absoluteUri: string, position: SmartPosition, message?: (m: Message) => void): BlameTree {
     const data = this.ReadStrictSync(absoluteUri);
+
+    // verify if the path exists in the document
+    if ('path' in position) {
+      let x = (<any>JSON.parse(data.ReadData()));
```
olydis: raw data can also be YAML and what not - `ReadObject` will always do the right thing so use that instead of `ReadData`

src/autorest-core/lib/data-store/data-store.ts
```diff
@@ -320,8 +321,28 @@ export class DataStore extends DataStoreView {
...
+      let keys = <string[]>(<any>position)['path'];
+      let validKeys = [];
+      let currObj = x;
+      for (let i = 0; i < keys.length; ++i) {
+        if (!(keys[i] in currObj)) {
+          if (message !== undefined) {
+            message({ Channel: Channel.Warning, Text: 'Could not find exact path: \'' + keys + '\'', Details: position });
+          }
```
olydis: would throw an equivalent Error if `message` is `undefined` (for using `Blame` in a stricter setting)

src/autorest-core/lib/data-store/data-store.ts
```diff
@@ -320,8 +321,28 @@ export class DataStore extends DataStoreView {
...
+          break;
+        }
+        validKeys.push(keys[i]);
+        currObj = currObj[keys[i]];
+      }
+      (<any>position)['path'] = validKeys;
+    }
+
     const resolvedPosition = CompilePosition(position, data);
```
olydis: verbose or not, your code looks like it would do what it should! 👍 
Howeeeeeever I'd suggest the following approach: just call out to `CompilePosition` and see whether it succeeds, if not work your way up and retry with shorter paths (and only if `message` is provided - maybe call that `onPathLookupError` to indicate that providing it means errors are no longer thrown).

This may sound less efficient, but actually it is the opposite! Here is why:
Objects we encounter here can be HUGE as in ~30k lines when JSON serialized (think fully resolved network spec meaning that even all samples are inlined!). Parsing those is noticably expensive and with your approach you do that for *EVERY* `Blame` (there are easily thousands of calls e.g. when building full source maps, as in `.map` files).
Okay, one could do smart caching (`ReadObject` does something close to caching, not that the object it returns is supposed to be mutable without changing the cached copy...) but we can do even better: the data store caches YAML ASTs of objects (those are considered immutable) and *that's* what `CompilePosition` operates on, making it faster than anything that involves parsing objects and such 😃 

Furthermore, we expect that the 99% case is that `CompilePosition` simply succeeds! So we can achieve 0 overhead for the 99% case using the "backing off on error" strategy

src/autorest-core/lib/data-store/data-store.ts
dsgouda: Really? so we do ```message!=undefined```?

olydis: you already have that `if`, don't you? Just talking about the `else` case that currently doesn't exist and hence swallows a problem ;-)

.vscode/launch.json
```diff
@@ -6,9 +6,12 @@
       "type": "node2",
       "request": "launch",
       "name": "run bootstrapper",
-      "program": "${workspaceRoot}/src/autorest/app.js",
-      "cwd": "${workspaceRoot}",
-      "preLaunchTask": "build/typescript"
+      "program": "${workspaceRoot}/src/autorest-core/app.js",
```
olydis: don't commit that or create a new task since this is not "run bootstrapper" anymore

src/autorest-core/lib/configuration.ts
```diff
@@ -232,18 +233,30 @@ export class ConfigurationView {
       // update source locations to point to loaded Swagger
       if (m.Source) {
         const blameSources = m.Source.map(s => {
-          try {
-            const blameTree = this.DataStore.Blame(s.document, s.Position);
-            const result = [...blameTree.BlameInputs()];
-            if (result.length > 0) {
-              return result.map(r => <SourceLocation>{ document: r.source, Position: Object.assign(TryDecodeEnhancedPositionFromName(r.name) || {}, { line: r.line, column: r.column }) });
+          let posClone = Clone(s.Position);
+          while (true) {
```
olydis: This may be why you're regeneration doesn't terminate, if `result` is empty this loops forever.

olydis:
fixes #2232

src/autorest-core/lib/configuration.ts
```diff
@@ -309,18 +310,32 @@ export class ConfigurationView {
...
+            try {
+              const blameTree = this.DataStore.Blame(s.document, posClone);
+              const result = [...blameTree.BlameInputs()];
```
olydis: hmmm I'm worried that this and the lines after are inside of the `try`! bugs in those lines would be swallowed (note that you also don't write the caught exception anywhere)

src/autorest-core/lib/configuration.ts
```diff
@@ -309,18 +310,32 @@ export class ConfigurationView {
       // update source locations to point to loaded Swagger
       if (m.Source) {
         const blameSources = m.Source.map(s => {
-          try {
-            const blameTree = this.DataStore.Blame(s.document, s.Position);
-            const result = [...blameTree.BlameInputs()];
-            if (result.length > 0) {
-              return result.map(r => <SourceLocation>{ document: r.source, Position: Object.assign(TryDecodeEnhancedPositionFromName(r.name) || {}, { line: r.line, column: r.column }) });
+          let posClone = Clone(s.Position);
+          while (true) {
```
olydis: control flow is super hard to analyze (at least for me) due to this loop and the numerous breaks and the return spilled across it! It does look sane to me, but I'd rather have this loop tighter and more expressive!

How about something like:
``` typescript
let blameTree: BlameTree | null = null;
while (blameTree === null) {
  try {
    blameTree = ...
  } catch (e) {
    // ~ current catch logic, message should also include `e` somehow
  }
}
const result = [...blameTree.BlameInputs()];
... // old logic
```

src/autorest-core/lib/configuration.ts
olydis: see above for tighter `try` proposal

olydis:
@dsgouda also, unit tests would be nice, so we see this does what it's supposed to do (think you can start with some existing test cases)

src/autorest-core/lib/configuration.ts
```diff
@@ -1,3 +1,5 @@
+import { BlameTree } from './source-map/blaming';
```
olydis: move below copyright header ;)

PR #2250, opened 2017-05-05
Literate Documentation Update

### LINKED ISSUE #1900: Operator '??' cannot be applied to operands of type 'Guid' and 'string'

olydis:
...and
- fixed a race condition (cached a bit too much when converting ASTs to objects)
- fixed #1900 (again, cough)
- made jsonpath library do filters right


Samples/3a-literate-swagger/Client/Models/ListQueryKeysResult.cs
```diff
@@ -29,8 +26,9 @@ public ListQueryKeysResult()
         /// <summary>
         /// Initializes a new instance of the ListQueryKeysResult class.
         /// </summary>
-        /// <param name="value">The query keys for the Azure Search
-        /// service.</param>
+        /// <param name="value">&gt; Again, shorthand for `@.properties.value`
+        ///
```
fearthecowboy: What's with the `&gt;` ? I've seen a couple of these.

Samples/3a-literate-swagger/Client/Models/ListQueryKeysResult.cs
```diff
@@ -43,7 +41,9 @@ public ListQueryKeysResult(IList<QueryKey> value = default(IList<QueryKey>))
         partial void CustomInit();
 
         /// <summary>
-        /// Gets the query keys for the Azure Search service.
+        /// Gets &amp;gt; Again, shorthand for `@.properties.value`
```
fearthecowboy: And now `&amp;gt;` ? Inception!

Samples/3a-literate-swagger/generated/swagger.md.json
```diff
@@ -49,7 +48,8 @@
             "$ref": "#/parameters/ResourceGroupName"
           },
           {
-            "$ref": "#/parameters/SearchServiceName"
+            "$ref": "#/parameters/SearchServiceName",
+            "description": "> Search is also performed relative to nodes discovered by parent headings.\n>\n> In this case, we're in the scope of the node identified by `#QueryKeys_List` which is the operation.\n>\n> Furthermore, `#SearchServiceName` is really shorthand for `@.#SearchServiceName` where `@` represents the current scope's node.\n> One could force searching the global scope with `$.#SearchServiceName`\n\nSome description."
```
fearthecowboy: `>` ? Should that be stripped by the GFM stripper?

olydis: that one runs on the code model (it was simpler that way) :/ we may indeed wanna change that since other endpoints may benefit from stripping GFM (but not force that, otherwise our "swagger resolver" endpoint may do unexpected magic to some people)

Samples/3a-literate-swagger/Client/Models/ListQueryKeysResult.cs
olydis: darn, fixed!

Issue #2067, opened 2017-03-26
[Linter] There should be rule that should ensure consistency in paging on a resource when listingbySubscription and listingByResourceGroup


amarzavery:
Either both should be pageable or both should not be pageable. It doesnt make sense to make one pageable and other not pageable.

Actual issue:
https://github.com/Azure/azure-rest-api-specs/issues/1070
/cc @ravbhatnagar @veronicagg

veronicagg:
@sarangan12 when you get to this issue, could you follow up on what this should be?

sarangan12:
Task completed

PR #2253, opened 2017-05-05
Compare x-ms-pageable between ListByResourceGroup & ListBySubscriptions operations

sarangan12:
Related to https://github.com/Azure/autorest/issues/2067 

@veronicagg @vishrutshah @dsgouda @salameer Please approve


src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -599,5 +583,88 @@ private static string GetReferencedModel(String schema, Dictionary<string, Schem
...
+        /// <param name="getOperations"></param>
+        /// <param name="regEx"></param>
+        /// <param name="definitionKey"></param>
+        /// <param name="definitions"></param>
+        /// <returns></returns>
+        public static bool ListByXCheck(Regex regEx, string definitionKey, Dictionary<string, Schema> definitions, ServiceDefinition serviceDefinition)
+        {
+            Operation operation = GetListByOperation(regEx, definitionKey, definitions, serviceDefinition);
+
+            if (operation != null)
```
dsgouda: ```return operation!=null;```

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -599,5 +583,88 @@ private static string GetReferencedModel(String schema, Dictionary<string, Schem
...
+        {
+            IEnumerable<Operation> getOperations = ValidationUtilities.GetOperationsByRequestMethod("get", serviceDefinition);
+
+            IEnumerable<Operation> operations = getOperations.Where(operation => regEx.IsMatch(operation.OperationId) &&
+                    IsXmsPageableResponseOperation(operation) &&
+                    operation.Responses.Any(
+                           response => response.Key.Equals("200") &&
+                           IsArrayOf(response.Value.Schema?.Reference, resourceModel, definitions)));
+
+            if (operations != null && operations.Count() != 0)
```
dsgouda: There is something like ```FirstOrNull``` in Linq, try using that.

src/modeler/AutoRest.Swagger/Validation/XMSPageableListByRGAndSubscriptionsValidation.cs
```diff
@@ -0,0 +1,72 @@
...
+            foreach(string resourceModel in parentTrackedResourceModels)
+            {
+                Operation listByResourceGroupOperation = ValidationUtilities.GetListByResourceGroupOperation(resourceModel, definitions, context.Root);
+                Operation listBySubscriptionOperation = ValidationUtilities.GetListBySubscriptionOperation(resourceModel, definitions, context.Root);
+
+                if(listByResourceGroupOperation != null && listBySubscriptionOperation != null)
+                {
+                    JObject rgXMSPageableExtension = (JObject)listByResourceGroupOperation.Extensions[ValidationUtilities.XmsPageable];
+                    JObject subXMSPageableExtension = (JObject)listBySubscriptionOperation.Extensions[ValidationUtilities.XmsPageable];
+                    if (!IsEqual(rgXMSPageableExtension,subXMSPageableExtension))
```
dsgouda: Do we really have to come up with a custom implementation for IsEqual here?

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -599,5 +583,88 @@ private static string GetReferencedModel(String schema, Dictionary<string, Schem
...
             => trackedResourceModels.Where(resourceModel => !childTrackedResourceModels.Any(childModel => childModel.Key.Equals(resourceModel)));
+
+        /// <summary>
+        /// For the provided resource model, it gets the operation which ends with ListByResourceGroup and returns the resource model.
+        /// </summary>
+        /// <param name="resourceModel"></param>
+        /// <param name="definitions"></param>
+        /// <param name="serviceDefinition"></param>
+        /// <returns></returns>
+        public static Operation GetListByResourceGroupOperation(string resourceModel, Dictionary<string, Schema> definitions, ServiceDefinition serviceDefinition)
```
dsgouda: Starting to wonder if we can create a separate class for listBy related methods and reduce some of the load in ValidationUtilities.

src/modeler/AutoRest.Swagger/Validation/XMSPageableListByRGAndSubscriptionsValidation.cs
sarangan12: No. we do not. That was the reason I marked it as Not ready for Review. I have used the DeepEquals logic with the latest commit.

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
sarangan12: It is FirstOrDefault method. It returns a default value around which the doc is poor. No risk approach followed here. no changes

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
sarangan12: Not sure why it is required. They are all simple methods with not a complex logic. If the logic gets difficult, I would suggest moving it out.

dsgouda: Fair enough, but let's keep an eye out for that, I don't want ```ValidationUtilities``` to grow out of hand.

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -599,5 +583,83 @@ private static string GetReferencedModel(String schema, Dictionary<string, Schem
...
+        }
+
+        /// <summary>
+        /// For the provided resource model, it gets the operation which matches with ListBySubscription and returns the resource model.
+        /// </summary>
+        /// <param name="resourceModel"></param>
+        /// <param name="definitions"></param>
+        /// <param name="serviceDefinition"></param>
+        /// <returns></returns>
+        public static Operation GetListBySubscriptionOperation(string resourceModel, Dictionary<string, Schema> definitions, ServiceDefinition serviceDefinition)
```
dsgouda: nit: not a blocker, but consider using lambda expressions for one liner implementations, makes the code compact and more readable.

src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -241,6 +241,9 @@ Modelers:
   <data name="TrackedResourceListByImmediateParentOperationMissing" xml:space="preserve">
     <value>The child tracked resource, '{0}' with immediate parent '{1}', must have a list by immediate parent operation.</value>
   </data>
+  <data name="XMSPagableListByRGAndSubscriptionsMismatch" xml:space="preserve">
+    <value>For the tracked resource '{0}', the x-ms-pageable must be same for list by resource group and subscriptions operations.</value>
```
veronicagg: Suggestion for message: "List by resource group and list by subscription operations for tracked resource `{0}` should be consistent for paging, either both should have "x-ms-pageable" extension, or both shouldn't.

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -599,5 +583,88 @@ private static string GetReferencedModel(String schema, Dictionary<string, Schem
...
+            if (operation != null)
+            {
+                return true;
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// For the provided resource model, it gets the operation which matches with specified regex and returns the resource model.
```
veronicagg: this comment indicates it returns the resource model, but it appears we're returning an operation, please review.

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -599,5 +583,88 @@ private static string GetReferencedModel(String schema, Dictionary<string, Schem
...
+        /// </summary>
+        /// <param name="regEx"></param>
+        /// <param name="resourceModel"></param>
+        /// <param name="definitions"></param>
+        /// <param name="serviceDefinition"></param>
+        /// <returns></returns>
```
veronicagg: could we fill in the proper comments on the returns?

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -599,5 +583,88 @@ private static string GetReferencedModel(String schema, Dictionary<string, Schem
...
+
+        /// <summary>
+        /// Checks if there is an operation that matches the given regex
+        /// </summary>
+        /// <param name="getOperations"></param>
+        /// <param name="regEx"></param>
+        /// <param name="definitionKey"></param>
+        /// <param name="definitions"></param>
+        /// <returns></returns>
+        public static bool ListByXCheck(Regex regEx, string definitionKey, Dictionary<string, Schema> definitions, ServiceDefinition serviceDefinition)
```
veronicagg: is this method needed? looks like it's a small wrapper for GetListByOperation, though returning boolean. Should we have a more representative name?

src/modeler/AutoRest.Swagger/Validation/TrackedResourceListBySubscription.cs
```diff
@@ -6,19 +6,16 @@
...
 {
     public class TrackedResourceListBySubscription : TypedRule<Dictionary<string, Schema>>
     {
-        private static readonly Regex listBySidRegEx = new Regex(@".+_(List|ListBySubscriptionId|ListBySubscription|ListBySubscriptions)$", RegexOptions.IgnoreCase);
-
         /// <summary>
         /// Id of the Rule.
         /// </summary>
-        public override string Id => "M3027";
+        public override string Id => "M3028";
```
veronicagg: is the id change intentional? if so, are you updating the documentation?

src/core/AutoRest.Core/Properties/Resources.resx
sarangan12: As discussed, Both the operations must have x-ms-pageable ops.

veronicagg: sure, how about something like "For tracked resource '{0}', x-ms-pageable extension values must be the same for....". I would just thinking of clarifying what we mean by  "x-ms-pageable must be the same"

veronicagg:
@sarangan12 you may want to look into https://github.com/Azure/autorest/blob/master/src/modeler/AutoRest.Swagger/Validation/Extensions/NextLinkPropertyMustExist.cs#L35 per our conversation on how we're accessing the "nextLinkName" for x-ms-pageable, so we're consistent or maybe merge the logic.

PR #2256, opened 2017-05-08
Include the model name in the JSON-RPC method request.

jhendrixMSFT:
We use the code model name as the JSON-RPC method prefix on the server
side, updating the client side generator to match.


jhendrixMSFT:
@sergey-shandar Here's the fix for the client generator.

jhendrixMSFT:
@olydis I think this failure is related to the below change.  Assuming this is the case how do I fix it?

https://github.com/Azure/azure-rest-api-specs/commit/4edca1c798d58d08294316728944b5ba5de8ff35#diff-e65ac296b8a010a5217b06d8b787c727

olydis:
you're correct, should be fixed on current master now, so rebase/merge should fix it :)

jhendrixMSFT:
Will do, thanks!

PR #2258, opened 2017-05-08
Batch mode (and a handful of other joy)

fearthecowboy:


src/autorest-core/app.ts
```diff
@@ -200,52 +215,232 @@ csharp:
...
+    }
+  }
+
+  await outstanding;
+
+  return exitcode;
+}
+
+
+async function deprecatedBatch(api: AutoRest): Promise<number> {
```
olydis: 😮

PR #2262, opened 2017-05-10
[Oav Pulgin] Update oav version to 0.4.1

vishrutshah:
Update oav plug-in to version 0.4.1


src/autorest-core/test/plugin.ts
```diff
@@ -57,7 +57,7 @@ import { LoadLiterateSwagger } from "../lib/pipeline/swagger-loader";
       }
     });
     assert.strictEqual(await autoRest.Process().finish, true);
-    const expectedNumErrors = 7;
+    const expectedNumErrors = 2;
```
olydis: do you know why this is? was oav reporting false positives?

vishrutshah: Yup we fixed some bugs in oav like `integers` in query param, `encoding` checks and added support for polymorphic models that's why some of the errors disappeared. 

Ref: https://github.com/Azure/oav/issues?q=is%3Aissue+is%3Aclosed

olydis: awesome!

vishrutshah:
Thanks for the review!!

olydis:
@vishrutshah *please* always squash and merge in the future!!!

vishrutshah:
@olydis Of course. It's my bad. I thought there was some settings we can enable so that this does not happen even by mistake. I'll be careful but we should enable that settings :)

olydis:
not sure about those settings, don't have access :) will ask Garrett when he's back

vishrutshah:
@olydis I've enabled them. So now you should see message that says "Not enabled for this repository " in " Create a merge commit " option.

olydis:
awesome, thank you! 🎉 🎈 🎆

PR #2263, opened 2017-05-10
various goodness

olydis:
- stack trace crunching in CI diff
- add and reference `azure-openapi-validator` plugin (ready to code there!)
- consolidate external plugin interface
- added `autorest-interactive` (experimental)
- fixed: exit code should be non-0 on error messages
- consistency with non-complaining in case of `--latest` and so on


Samples/2a-arm-validation/readme.md
```diff
@@ -14,14 +14,17 @@ input-file:
...
 ```
 
 ## Generation
 
 Also generate for some languages.
 
 ``` yaml 
+azure-arm: true # generate code using Azure ARM flavor
 csharp:
   output-folder: CSharp
```
dsgouda: Lets clean this up while we're at it

src/autorest-core/legacyCli.ts
```diff
@@ -60,7 +59,7 @@ export async function CreateConfiguration(baseFolderUri: string, inputScope: Dat
   const codegenerator = switches["g"] || switches["codegenerator"] || "CSharp";
   const usedCodeGenerator = codegenerator.toLowerCase().replace("azure.", "").replace(".fluent", "");
   if (codegenerator.toLowerCase() === "none") {
-    result["azure-validator"] = true;
+    (<any>result)["azure-validator"] = true;
```
dsgouda: Why?

src/autorest-core/lib/autorest-core.ts
```diff
@@ -167,9 +167,14 @@ export class AutoRest extends EventEmitter {
         }
 
         if (view.InputFileUris.length === 0) {
-          throw new Exception("No input files provided.\n\nUse --help to get help information.", 0);
+          if (view.GetEntry("allow-no-input")) {
```
dsgouda: Wait, if there is no input what does AutoRest work with

src/autorest-core/lib/outstanding-task-awaiter.ts
```diff
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { OutstandingTaskAlreadyCompletedException } from './exception';
+import { OutstandingTaskAlreadyCompletedException } from "./exception";
```
dsgouda: Out of curiosity, why did we have to rewrite our own versions of locking and stuff, pretty sure node has a bunch of reusable libraries

src/autorest-core/lib/pipeline/commonmark-documentation.ts
```diff
@@ -61,7 +61,7 @@ export async function ProcessCodeModel(codeModel: DataHandleRead, scope: DataSto
       const nodeOriginal = CloneAst(node);
       const key = nodeOriginal.key.value;
       const origKey = key + "_Original";
-      nodeOriginal.key.value = origKey
+      nodeOriginal.key.value = origKey;
```
dsgouda: This gives me an idea, do you think we could add a linting task to our gulp build which sanitizes typescript
Something like [this](https://github.com/palantir/tslint) (ofc we need to see if it really is effective)

src/autorest-core/package.json
```diff
@@ -43,6 +43,8 @@
     "jsonpath": "^0.2.11",
     "linq-es2015": "^2.4.13",
     "oav": "0.4.1",
+    "autorest-interactive": "git+https://github.com/olydis/autorest-interactive.git",
+    "azure-openapi-validator": "git+https://github.com/olydis/azure-openapi-validator.git#b45423f4560585e64f574bbffbb5cdf894cfbd53",
```
dsgouda: Should we push this change now?

Samples/2a-arm-validation/readme.md
olydis: clean up? that was the intention! :D

src/autorest-core/legacyCli.ts
olydis: no longer part of the data type since the core shouldn't care about this - this right here is the single exception. This code translates legacy CLI to an in-memory config - something that doesn't have to be over-engineered if you ask me.

src/autorest-core/lib/autorest-core.ts
olydis: nothing, it's supposed to silently exit in some cases (e.g. `autorest --latest`) but fail in others (e.g. `autorest --azure-validator`)

src/autorest-core/lib/outstanding-task-awaiter.ts
olydis: this is not for locking (which would also indicate bad design), this is merely a convenience layer over `Promise.all`. I'm no node library expert, I'd be glad to learn about such about a library!

src/autorest-core/lib/pipeline/commonmark-documentation.ts
olydis: that's exactly the tool I use myself, but its "auto-fixes" sometimes break stuff, otherwise I'd be for that!

src/autorest-core/package.json
olydis: for `azure-openapi-validator` I'd say yes, the other one I will indeed remove until Garrett has acquisition figured out.

Samples/2a-arm-validation/readme.md
dsgouda: Talking about the trailing character 😄

olydis: oh dear... do you know how that can easily be done? the thing is that my git is configured to not change line endings which is usually what you want except right now :D will see

src/autorest-core/lib/autorest-core.ts
dsgouda: Shouldn't we display help if invalid inputs are provided. Not a blocker, just curious.

src/autorest-core/lib/outstanding-task-awaiter.ts
dsgouda: ```this is not for locking ``` d'oh totally forgot we are in single threaded world.

src/autorest-core/lib/pipeline/commonmark-documentation.ts
dsgouda: Iiiiiinteresting.....

src/autorest-core/lib/autorest-core.ts
olydis: we do, the thing is that in said cases the input is infact valid, but autorest currently complains, which causes roughly 2 people per week to ask us why their scripts are "failing" when in fact they are not, AutoRest is just complaining mistakenly

Samples/2a-arm-validation/readme.md
olydis: will do that another time, would probably be easiest and best to add `.gitattributes` but wanna talk about that with the cowboy first...

PR #2264, opened 2017-05-12
Adding new validation rule categorization

dsgouda:
Had to recreate this PR since the older branch somehow got deleted
Addressing issue #2225 

Validation rules which may or may not need this categorization
M3006, M3017, S2004 and M2014

### DIFF
--- a/src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
+++ b/src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
@@ -26,6 +26,11 @@ namespace AutoRest.Swagger.Validation
         /// </summary>
         public override ValidationCategory ValidationCategory => ValidationCategory.RPCViolation;
 
+        /// <summary>
+        /// What kind of change implementing this rule can cause.
+        /// </summary>
+        public override ValidationChangesImpact ValidationChangesImpact => ValidationChangesImpact.PotentialServiceBreakingChanges;
+
         public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string, Dictionary<string, Operation>> entity, RuleContext context)
         {
             // get all operation objects that are either of get or post type
--- a/src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
+++ b/src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
@@ -29,7 +29,7 @@ namespace AutoRest.Swagger.Validation
         /// <summary>
         /// What kind of change implementing this rule can cause.
         /// </summary>
-        public override ValidationChangesImpact ValidationChangesImpact => ValidationChangesImpact.PotentialServiceBreakingChanges;
+        public override ValidationChangesImpact ValidationChangesImpact => ValidationChangesImpact.ServiceImpactingChanges;
 
         public override IEnumerable<ValidationMessage> GetValidationMessages(Dictionary<string, Dictionary<string, Operation>> entity, RuleContext context)
         {
--- a/src/modeler/AutoRest.Swagger/Validation/LROStatusCodesValidation.cs
+++ b/src/modeler/AutoRest.Swagger/Validation/LROStatusCodesValidation.cs
@@ -26,6 +26,11 @@ namespace AutoRest.Swagger.Validation
         public override ValidationCategory ValidationCategory => ValidationCategory.SDKViolation;
 
         /// <summary>
+        /// What kind of change implementing this rule can cause.
+        /// </summary>
+        public override ValidationChangesImpact ValidationChangesImpact => ValidationChangesImpact.PotentialServiceBreakingChanges;
+
+        /// <summary>
         /// The template message for this Rule. 
         /// </summary>
         /// <remarks>
--- a/src/modeler/AutoRest.Swagger/Validation/LROStatusCodesValidation.cs
+++ b/src/modeler/AutoRest.Swagger/Validation/LROStatusCodesValidation.cs
@@ -28,7 +28,7 @@ namespace AutoRest.Swagger.Validation
         /// <summary>
         /// What kind of change implementing this rule can cause.
         /// </summary>
-        public override ValidationChangesImpact ValidationChangesImpact => ValidationChangesImpact.ServiceImpactingChanges;
+        public override ValidationChangesImpact ValidationChangesImpact => ValidationChangesImpact.SDKImpactingChanges;
 
         /// <summary>
         /// The template message for this Rule. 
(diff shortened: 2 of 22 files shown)


src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
```diff
@@ -26,6 +26,11 @@ public class CollectionObjectPropertiesNamingValidation : TypedRule<Dictionary<s
         /// </summary>
         public override ValidationCategory ValidationCategory => ValidationCategory.RPCViolation;
 
+        /// <summary>
+        /// What kind of change implementing this rule can cause.
+        /// </summary>
+        public override ValidationChangesImpact ValidationChangesImpact => ValidationChangesImpact.ServiceImpactingChanges;
```
veronicagg: is this rule serviceimpacting? I'm thinking it's mostly around the modeling of pageable in the swagger, thoughts?

src/modeler/AutoRest.Swagger/Validation/LROStatusCodesValidation.cs
```diff
@@ -26,6 +26,11 @@ public class LROStatusCodesValidation : TypedRule<Dictionary<string, Dictionary<
         public override ValidationCategory ValidationCategory => ValidationCategory.SDKViolation;
 
         /// <summary>
+        /// What kind of change implementing this rule can cause.
+        /// </summary>
+        public override ValidationChangesImpact ValidationChangesImpact => ValidationChangesImpact.ServiceImpactingChanges;
```
veronicagg: this is most likely SDKImpacting, I believe, since it's the way we ask folks to model the long running operation in the swagger, including the "end result" model, so that our generated code knows what to expect after polling. Thoughts?

src/modeler/AutoRest.Swagger/Validation/NonAppJsonTypeWarning.cs
```diff
@@ -30,6 +30,11 @@ public override bool IsValid(string entity)
             => (entity.ToLowerInvariant().StartsWith(AppJsonType));
         
         /// <summary>
+        /// What kind of change implementing this rule can cause.
+        /// </summary>
+        public override ValidationChangesImpact ValidationChangesImpact => ValidationChangesImpact.ServiceImpactingChanges;
```
veronicagg: I'm not sure on the category for this one, if something other than application/json is included, we just report the warning to make sure the service does support it and if not, it gets removed from the swagger, so in terms of impact, it seems like is more of a swagger update other than a service impacting change, on the other hand. It may or may not impact the SDK depending on whether the code generator knows what to do with the other types. Thoughts?

src/modeler/AutoRest.Swagger/Validation/SecurityDefinitionsStructureValidation.cs
```diff
@@ -24,6 +24,11 @@ public class SecurityDefinitionsStructureValidation : TypedRule<Dictionary<strin
         public override ValidationCategory ValidationCategory => ValidationCategory.SDKViolation;
 
         /// <summary>
+        /// What kind of change implementing this rule can cause.
+        /// </summary>
+        public override ValidationChangesImpact ValidationChangesImpact => ValidationChangesImpact.ServiceImpactingChanges;
```
veronicagg: Is this service impacting? I'm thinking it'll be a swagger update only, do you know how this translated into the SDK generated code?

src/modeler/AutoRest.Swagger/Validation/TrackedResourceGetOperationValidation.cs
```diff
@@ -24,6 +24,11 @@ public class TrackedResourceGetOperationValidation : TypedRule<Dictionary<string
         public override ValidationCategory ValidationCategory => ValidationCategory.RPCViolation;
 
         /// <summary>
+        /// What kind of change implementing this rule can cause.
+        /// </summary>
+        public override ValidationChangesImpact ValidationChangesImpact => ValidationChangesImpact.SDKImpactingChanges;
```
veronicagg: I believe this would be ServiceImpacting if the get operation is not implemented by the service.

src/modeler/AutoRest.Swagger/Validation/TrackedResourcePatchOperationValidation.cs
```diff
@@ -26,6 +26,11 @@ public class TrackedResourcePatchOperationValidation : TypedRule<Dictionary<stri
         public override ValidationCategory ValidationCategory => ValidationCategory.RPCViolation;
 
         /// <summary>
+        /// What kind of change implementing this rule can cause.
+        /// </summary>
+        public override ValidationChangesImpact ValidationChangesImpact => ValidationChangesImpact.SDKImpactingChanges;
```
veronicagg: similarly as previous, I believe this would be service impacting if the patch operation is not implemented by the service.

veronicagg:
Regarding the rules you asked above, here are my thoughts: M3006 - ServiceImpacting; M3017 - ServiceImpacting; S2004 - SDKImpacting (I asked a question on the PR); M2014 - SDK Impacting.

src/modeler/AutoRest.Swagger/Validation/CollectionObjectPropertiesNamingValidation.cs
dsgouda: I believe renaming a model's property that is readOnly means stuff needs to change at the service end.

src/modeler/AutoRest.Swagger/Validation/NonAppJsonTypeWarning.cs
dsgouda: Do we generate SDKs in AutoRest that can support non app json types? If yes, we can probably come up with another category (None) indicating no impact on either service or sdk ends.

src/modeler/AutoRest.Swagger/Validation/SecurityDefinitionsStructureValidation.cs
dsgouda: A. Not sure, we may have to ask someone more knowledgeable 
B. @olydis do we?

olydis: `securityDefinitions` don't affect our code generation as far as I can tell, think it's mainly documentational or to make the Swagger "complete" in the sense of documenting how to access the service. At least for C#, it seems like the same credential mechanics are generated regardless.

dsgouda: So if I understand correctly, this just defines the mechanics of communicating with the service. If the service allows communication irrespective of the security definitions provided, this really shouldn't affect either the service or the SDK generated. tl;dr ValidationChangesImpact must be None.
Thoughts @veronicagg ?

veronicagg: @dsgouda agree, that seems the right category for this one.

Issue #2196, opened 2017-04-24
Convert Id and ValidationCategory fields in Rule.cs to abstract properties


dsgouda:
This requires ensuring every single ValidationRule class that inherits from Rule.cs has the ValidationCategory and Id fields overridden. We might also generate some diffs in the regenerated code here.

dsgouda:
Need a few Validation Ids here
@kirthik can you take a look
[SupportedSchemesWarning](https://github.com/Azure/autorest/blob/43a65e4eab01ef64b195d28e42556f9d0d316e7f/src/modeler/AutoRest.Swagger/Validation/SupportedSchemesWarning.cs)
[AvoidMsdnReferences](https://github.com/Azure/autorest/blob/a2d51c88366b683e1ca92cbeceb91c16b4ffdb6c/src/modeler/AutoRest.Swagger/Validation/AvoidMsdnReferences.cs)
[NoControlCharacters](https://github.com/Azure/autorest/blob/a2d51c88366b683e1ca92cbeceb91c16b4ffdb6c/src/modeler/AutoRest.Swagger/Validation/Core/NoControlCharacters.cs)


Need Validation Categories here
[NoControlCharacters](https://github.com/Azure/autorest/blob/a2d51c88366b683e1ca92cbeceb91c16b4ffdb6c/src/modeler/AutoRest.Swagger/Validation/Core/NoControlCharacters.cs)

kirthik:
For what would we check NoControlCharacters rule? Operations or model properties that are strings? or both? It seems like SDKViolation to me. Could you also tell me the impact?  Does validation failing on this rule mean SDK cannot be generated?

dsgouda:
Come to think of it I have never seen ```NoControlCharacters``` violation flagged (checked rest spec logs too), plus there is no test case for this either. Makes me wonder if we even allow control characters anywhere in our specs. @olydis does AutoRest read a document with control characters of any kind?

olydis:
@dsgouda if it's valid JSON/YAML (control characters are perfectly legal if properly escaped and such), AutoRest will read it, yes

Also never saw that rule in action, may be worth an investigation / negative test case ;-)

dsgouda:
Tested with ```NoControlCharacters``` rule control characters in specs, unescaped documents are correctly flagged by AutoRest even before the validator kicks and escaped characters are read as plain strings instead of characters.
[...]

PR #2268, opened 2017-05-15
Removing NoControlCharacters/fixing messages for path not found

dsgouda:
Removing unused validation rule ```NoControlCharacters``` and adding more context for warnings about paths not found

also addresses https://github.com/Azure/autorest/issues/2269


src/autorest-core/lib/configuration.ts
```diff
@@ -326,8 +326,8 @@ export class ConfigurationView {
                 const path = s.Position.path as string[];
                 if (path) {
                   this.Message({
-                    Channel: Channel.Warning,
-                    Text: `Could not find the exact path ${JSON.stringify(path)}`
+                    Channel: Channel.Debug,
```
olydis: this is supposed to be visible! i.e. if you see noise because of this, it's because of bugs in rules so the noise is exactly what I intended with this. The noise is supposed to be removed by fixing the bugs instead of muting the noise ;)

dsgouda: Aye, Capitan!

src/modeler/AutoRest.Swagger/Validation/Core/RulesExtensions.cs
```diff
@@ -64,6 +64,6 @@ public static IEnumerable<Rule> GetValidationRules(this Type type)
         /// <summary>
         /// The collection of rules that apply to all properties regardless of other rules.
         /// </summary>
-        public static IEnumerable<Rule> UniversalRules = new[] { new NoControlCharacters() };
+        public static IEnumerable<Rule> UniversalRules = new List<Rule>();
```
olydis: so we don't have universal rules afterwards? In that case I'd remove this entirely (and maybe some places where it's used get cleaner) since we shouldn't ever add a universal rule.
Not only because it's way simpler in TS, I'm also just super skeptical about "universal". In case of this rule, even control characters may have their place in Swaggers, e.g. in `examples` sections where you can have raw response bodies where control characters are perfectly valid.

dsgouda: A. NoControlCharacters was the only universal rule implemented from what I can tell.
B. "We shouldn't ever add a universal rule." Not sure about that, rather I don't trust the ever changing requirements for validator.
C. I agree it would be cleaner on the ts side of things, but until that is enabled let's humor this functionality, what say?

olydis: A. excellent
B. well my main point is that "universal" is a weird concept and I bet it is always just "almost universal" and really just flawed
C. sure, no problem, just thought since this project isn't really supposed to grow anymore, we can cut dead code - but we can wait until the TS project is solid

but that also reminds me: didn't you say you also tried triggering this rule? did you succeed? if no, universal rules may not be implemented correctly anyways, so there is an opportunity to fix a bug by removing the buggy code 😆

dsgouda: The rule does get triggered its just that it never received any faulty strings by the time it was invoked, which makes me believe the AutoRest core engine is rejecting control characters even before the validator is called into action

src/autorest-core/lib/configuration.ts
```diff
@@ -327,7 +327,7 @@ export class ConfigurationView {
                 if (path) {
                   this.Message({
                     Channel: Channel.Warning,
-                    Text: `Could not find the exact path ${JSON.stringify(path)}`
+                    Text: `Could not find the exact path ${JSON.stringify(path)} for ${JSON.stringify(m.Details)}`
```
veronicagg: what comes in the details here?

dsgouda: The id (name of the rule), severity and all the categorizations for that rule.

src/modeler/AutoRest.Swagger/Validation/Core/RulesExtensions.cs
olydis: the core should be doing no such thing (and doesn't seem to in the test I just did), can I have the test case? let's make sure we're not missing a bug.

olydis: just verified real quick, see https://github.com/Azure/autorest/issues/2196#issuecomment-301649410

olydis: so apparently the rule apparently does something, I'd only remove it here if we add an equivalent in the TS world (it will be more performant, easier to debug and inherently report correct paths)

PR #2297, opened 2017-05-26
[Go] Simpler names for clients

mcardosos:
Fixes one of the issues @colemickens mentions here https://github.com/Azure/azure-sdk-for-go/issues/616
Renaming the very confusing `GroupClient` to `Client`

PTAL
@jhendrixMSFT @marstr

### DIFF
--- a/src/generator/AutoRest.Go/CodeNamerGo.cs
+++ b/src/generator/AutoRest.Go/CodeNamerGo.cs
@@ -319,7 +319,7 @@ namespace AutoRest.Go
 
             // we use the base implementation here as it uses a case-insensitive comparison.
             // this is a bit of a hacky work-around for some naming changes introduced in core...
-            return EnsureNameCase(PascalCase(RemoveInvalidCharacters(base.GetEscapedReservedName(name, "Group"))));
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
While I like this change it's gonna be disruptive.  Have you looked at the diff in the generated SDK?

mcardosos:
It introduces breaking changes, yes.

src/generator/AutoRest.Go/CodeNamerGo.cs
```diff
@@ -319,7 +319,7 @@ public override string GetMethodGroupName(string name)
 
             // we use the base implementation here as it uses a case-insensitive comparison.
             // this is a bit of a hacky work-around for some naming changes introduced in core...
-            return EnsureNameCase(PascalCase(RemoveInvalidCharacters(base.GetEscapedReservedName(name, "Group"))));
+            return EnsureNameCase(PascalCase(RemoveInvalidCharacters(name)));
```
jhendrixMSFT: I think you can remove the comments preceding this line yes?

olydis:
@mcardosos okay, this one is easy to replay, I'll let you have this one 😄

PR #2308, opened 2017-05-30
level up

olydis:
- `autorest-interactive` plugin
- support index-based blaming (so stuff may complaing about `{ index: 1337 }` and it'll do the right thing)
- `x-ms-enum`
- removed compare mode (will be extracted to new tool that will be able to handle composite and such by leveraging AAAL)
- `-FANCY` by default (disable using `-LEGACY`)


src/core/AutoRest/Program.cs
```diff
@@ -91,13 +91,6 @@ private static int Main(string[] args)
                             return 0;
                         }
 
-                        // comparison?
```
fearthecowboy: Quick Question -- do we even support an execution path that runs `main() ` without `--server` (maybe for `-help`, which we should remove too.)

olydis: not sure I understand correctly: we haven't moved to `-FANCY` yet (and even then, we may still make the true legacy CLI accessible for a while). I suspect 90% of AutoRest users still run this code without `--server`, so I think we still support such an execution path.

Of course, removing the `-help` path may boost migration a bit.

fearthecowboy: Oh. I thought we had at least moved to `-FANCY` ... I totally forgot about that. 

I haven't run with the legacy cmdline in a long time at all tho' :D

Ah, right, we have to write a better cmdline help before we turn off that.

olydis: Nah we postponed that again and again... shall we move? :-D

Right, how about I turn our CLI help from the [docs](https://github.com/Azure/autorest/blob/master/docs/user/cli.md) into markdowny `--help`

Issue #2224, opened 2017-05-01
[Go] Ugly names for private variables starting with acronyms


mcardosos:
For example `aPIID is aPI identifier` when generating apideployment.

mcardosos:
Also in apideployment: `// APIOperationsClient is the composite Swagger for ApiManagement Client
`

mcardosos:
Moved last coment to https://github.com/Azure/autorest/issues/2238

PR #2310, opened 2017-05-30
Caps on acronyms

mcardosos:
Fixes https://github.com/Azure/autorest/issues/2224
PTAL
@marstr @jhendrixMSFT

### DIFF
--- a/src/generator/AutoRest.Go/Extensions.cs
+++ b/src/generator/AutoRest.Go/Extensions.cs
@@ -69,20 +69,8 @@ namespace AutoRest.Go
         /// <returns></returns>
         public static bool StartsWithAcronym(this string value)
         {
-            value = value.Trim();
-            string[] words = value.Split(' ', '-', '_');
-            if (words.First().Length > 1)
-            {
-                for (int i = 0; i < words.First().Length; i++)
-                {
-                    if (char.IsLower(words.First()[i]))
-                    {
-                        return false;
-                    }
-                }
-                return true;
-            }
-            return false;
+            string firstWord = value.Trim().Split(' ', '-', '_').First();
+            return firstWord.Length > 1 && firstWord.All(c => char.IsUpper(c));
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


mcardosos:
Generated code example
``` go
// Create adds an API to the specified product.
//
// resourceGroupName is the name of the resource group. serviceName is the name
// of the API Management service. productID is product identifier. Must be
// unique in the current API Management service instance. APIID is API
// identifier. Must be unique in the current API Management service instance.
func (client ProductApisClient) Create(resourceGroupName string, serviceName string, productID string, APIID string) (result autorest.Response, err error) {
```

src/generator/AutoRest.Go/Extensions.cs
```diff
@@ -53,10 +53,32 @@ public static string ToSentence(this string value)
...
                 value = value.Trim();
+                if (value.StartsWithAcronym())
+                {
+                    return value;
+                }                
                 return value.First().ToString().ToLowerInvariant() + (value.Length > 1 ? value.Substring(1) : "");
             }
         }
 
+        public static bool StartsWithAcronym(this string value)
```
jhendrixMSFT: Can you please add a comment explaining the algorithm this function uses?  It looks like it iterates over the string and if it's all caps then it returns true (seems reasonable however it does depend on the swagger properly casing them).

mcardosos: Yes, it depends on the correct casing in swagger.

mcardosos: Fixed

src/generator/AutoRest.Go/Extensions.cs
```diff
@@ -53,11 +53,39 @@ public static string ToSentence(this string value)
...
+                {
+                    if (char.IsLower(words.First()[i]))
+                    {
+                        return false;
+                    }
+                }
+                return true;
+            }
+            return false;
+        }
```
marstr: I think this method could be tidied up!

``` C#
public static bool StartsWithAcronym(this string value)
{
    value = value.Trim();
    var firstWord = value.Split(' ', '-','_').First();
    var lowercaseLetters = firstWord.Where(c => char.IsLower(c))
    return firstWord.Length > 1 && lowercaseLetters.Any()
}
```

mcardosos: Fixed (even fancier!)

marstr: Haha, so it is!

PR #2319, opened 2017-06-01
bring joy to: single cores, AAAL message consumers, Ruby people, CLI users

### LINKED ISSUE #2332: Keep autorest message.Text untouched
AutoRest is formatting the message.Text as follows: https://github.com/Azure/autorest/blob/master/src/autorest-core/lib/configuration.ts#L398 
For consumers of AutoRest library (ex. Vs Code extension), it'd be better to leave "text" untouched, so the consumers can put together their own formatting based on the different message fields. The way it is today, since "text" is adding together several things, it would require parsing, which is not desired. 
For command line purposes, AutoRest can still keep a "formattedMessage" property, which could be put together as desired.

olydis:
- give single cores a chance to run gulp stuff
- fixes #2332 
- fixes a bug (saw it affect Ruby CI) where a namespace like "A::B::C" is used during URI resolution (it was the default path to emit some artifacts), which got URI resolution very confused (is this relative or absolute?)
- emit a warning if both `-` and `--` arguments are mixed

### DIFF
--- a/src/autorest-core/lib/message.ts
+++ b/src/autorest-core/lib/message.ts
@@ -30,7 +30,10 @@ export interface Message {
   Key?: Iterable<string>;
   Details?: any;
   Text: string;
+
+  // injected or modified by core
   Source?: Array<SourceLocation>;
   Range?: Iterable<Range>;
   Plugin?: string;
+  FormattedMessage?: string;
 };
--- a/src/autorest-core/app.ts
+++ b/src/autorest-core/app.ts
@@ -143,15 +143,15 @@ function subscribeMessages(api: AutoRest, errorCounter: () => void) {
       case Channel.Debug:
       case Channel.Verbose:
       case Channel.Information:
-        console.log(m.FormattedMessage);
+        console.log(m.FormattedMessage || m.Text);
         break;
       case Channel.Warning:
-        console.warn(m.FormattedMessage);
+        console.warn(m.FormattedMessage || m.Text);
         break;
       case Channel.Error:
       case Channel.Fatal:
         errorCounter();
-        console.error(m.FormattedMessage);
+        console.error(m.FormattedMessage || m.Text);
         break;
     }
   });
--- a/src/autorest-core/app.ts
+++ b/src/autorest-core/app.ts
@@ -82,9 +82,23 @@ ${Stringify(config).replace(/^---\n/, "")}
     config["base-folder"] = currentDirUri;
     const api = new AutoRest(new RealFileSystem());
     await api.AddConfiguration(config);
+    const view = await api.view;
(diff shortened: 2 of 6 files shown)


marstr:
Wow, I'm shocked that this is a thing. Gulp doesn't special-case 0 to mean no threshold instead of literally do not run? I'm not immediately able to find documentation about it either.

olydis:
well... we wrote the part that interprets the threshold, and no, we take it very literal 😄

marstr:
Okay okay, fair enough :) Maybe fixing that part is the better fix?

olydis:
Well, the value kind of flows to several places (saying "the part" was a bit oversimplified), I'm always a big fan of input sanitation, that way further consumers of the value won't have to think too hard... but yeah, if it was just a, say, input to a single class that handles all throttling, then yes, that class should absolutely take care of sanitation, not the person who feeds in the value 🙂

src/autorest-core/lib/message.ts
```diff
@@ -30,7 +30,10 @@ export interface Message {
   Key?: Iterable<string>;
   Details?: any;
   Text: string;
+
+  // injected or modified by core
   Source?: Array<SourceLocation>;
   Range?: Iterable<Range>;
   Plugin?: string;
+  FormattedMessage?: string;
```
veronicagg: Thanks for making this change. Side question, would it make sense to have another property like "InfoUrl" in the Message, so it can carry a link to a wiki or similar that contains more information on the issue, or is that expected to be somehow included the Details?

olydis:
@azuresdkci test this please

src/autorest-core/lib/message.ts
olydis: interesting point! so to me, the only difference between "details" (which is essentially free form) and the main thingy is whether the property applies to most messages or not. having an info url sounds awefully useful and something one could encourage for most messages (I'd say comparable to source location and such), so I'd be cool with adding it to the main type 🙂 be aware, though, that the JSON formatted messages (used in the specs CI) just serialize out the details, so if it's supposed to show up there, we either gotta modify the formatter or also add it to the details... hmmm

veronicagg: @olydis I agree, I think it'd be applicable to all messages, so I'd vote for adding it. Yes, the formatted message should be evaluated every time something gets added, in this case, it probably should be added either to formatted message or details or both, so it's printed in CI/command line with the rest.

olydis: cool, feel free to start emitting such properties - autorest does not throw away properties on messages anyway, so adding it to `Details` from, say, the validator plugin will work out of the box (you'll see it in CI, in the VS code extension, etc.).

PR #2326, opened 2017-06-02
Consuming the azure-openapi-linter from npm feeds

dsgouda:
Renaming the `validator` to `linter` since we have multiple "validators" for openapis and there should be a way to distinguish between them.
Thanks @vishrutshah for the insight!


dsgouda:
@azuresdkci Test this please

dsgouda:
@olydis any suggestions regarding the naming here?

olydis:
@dsgouda `azure-openapi-validator` is my favorite, as originally named. These all *are* validators that together are used in a linter, so naming them consistently makes perfect sense, such as you would name stuff `childA`, `childB`, `childC` instead of `childA`, `kidB`, `kiddoC`.

> and there should be a way to distinguish between them.

The way to distinguish is the prefix - `model...` validates models using examples, `semantic...` validates semantics, `azure...` does azure specific stuff

dsgouda:
@azuresdkci Test this please

src/autorest-core/package.json
```diff
@@ -44,7 +44,7 @@
     "jsonpath": "^0.2.11",
     "linq-es2015": "^2.4.13",
     "oav": "0.4.1",
-    "azure-openapi-linter": "0.1.1",
+    "azure-openapi-linter": "0.1.3",
```
vishrutshah: Should this be "^0.1.3" as we should ideally follow semantic versioning

Samples/3b-custom-transformations/generated/pipeline.yaml
```diff
@@ -51,20 +51,18 @@ swagger-document/individual/azure-validator:
...
-    - swagger-document/azure-validator
-swagger-document/individual/azure-openapi-validator:
-  pluginName: azure-openapi-validator
+swagger-document/individual/azure-openapi-linter:
+  pluginName: azure-openapi-linter
   configScope:
     - azure-validator-individual
   inputs:
     - swagger-document/individual/transform
-    - swagger-document/individual/azure-validator
```
veronicagg: what's the reason for removing this?

dsgouda: This is autogenerated and for some reason the line gets deleted. @olydis any idea why this might happen?

dsgouda: `swagger-document/individual/azure-validator` scope used to set the `merge state` to `composed` only. By assigning proper configScopes to both `swagger-document/azure-openapi-linter` and `swagger-document/individual/azure-openapi-linter` we should be able to run the linter in both `composed` and `individual` merge states.

dsgouda:
@veronicagg @vishrutshah I tend to agree with the issues that @olydis pointed out regarding our package. Any alternative names/ideas here?

dsgouda:
@azuresdkci Test this please

vishrutshah:
@dsgouda I see his viewpoint and those are reasonable. Unfortunately, now as we already published the package I'd not try to change it unless the name is the huge concern. Thoughts?

veronicagg:
@dsgouda changes look good, regarding the name of the package, I see @olydis point, so you could change it back to that name if you prefer.

dsgouda:
@veronicagg 
I will have to rename the repo, republish the package and update this PR in that case. Will do it soon.
@vishrutshah That would require some modification to our publishing task correct?

vishrutshah:
@dsgouda Yes it'll require minor change in publishing task. Just curious what are you planning to do with the package that has been published. You need to make sure to take it down and remove from npm.

olydis:
@dsgouda @vishrutshah since the package has the wrong name (subtle reminder), taking the published one down is a side note 😉

src/autorest-core/resources/default-configuration.md
```diff
@@ -111,14 +111,12 @@ pipeline:
...
-    input:
-      - transform
-      - azure-validator # artificial predecessor in order to ensure order of messages for CI purposes
-  swagger-document/individual/azure-openapi-validator:
-    input: 
-      - transform
-      - azure-validator # artificial predecessor in order to ensure order of messages for CI purposes
+  swagger-document/azure-openapi-linter:
+    input: transform
+    scope: azure-validator-composed # artificial predecessor in order to ensure order of messages for CI purposes
```
olydis: the comment `# artificial predecessor in order to ensure order of messages for CI purposes` makes no sense here anymore! it was meant as an explanation for the 2nd input, so it doesn't accidentally get removed... but it is removed now, and the comment was just moved to a place where it has no meaning. As the comment is trying to clarify, the 2nd input is imposing a dependency which ensures *determinism* of our output! Unless CI and tests are updated to not rely on order of emitted messages, this will only pass CI with extreme luck.

Personally, I also like programs that produce the same output if run twice ;)

Samples/3b-custom-transformations/generated/pipeline.yaml
olydis: @dsgouda the input you removed was exactly ensuring the right scope, so removing this did not ensure "proper configScopes" (as you can see, lines 57 and 63 haven't changed). See below for why I think removing the additional input is also not good.

src/autorest-core/test/resources/small-input/generated/swagger
```diff
@@ -0,0 +1,1348 @@
+{
```
olydis: whoops? looks like you've run autorest manually here, this file does not belong on github

src/autorest-core/resources/default-configuration.md
veronicagg: @olydis could you point me to the documentation on this? or have a session to explain (probably again) how this works. I have a feeling the confusion is happening because it's not clear how the config file should be used.

veronicagg: @olydis similar questions as above related the file in the next comment "pipeline.yaml"

dsgouda: What @veronicagg said!

src/autorest-core/package.json
```diff
@@ -44,7 +44,7 @@
     "jsonpath": "^0.2.11",
     "linq-es2015": "^2.4.13",
     "oav": "0.4.1",
-    "azure-openapi-validator": "git+https://github.com/olydis/azure-openapi-validator.git#b45423f4560585e64f574bbffbb5cdf894cfbd53",
+    "azure-openapi-linter": "^0.1.4",
```
vishrutshah: Let's wait until we release `azure-openapi-validator` version and update this so you don't have to do back & forth.

src/autorest-core/resources/default-configuration.md
olydis: yes, we can have the session again, but: this format will likely change after Garrett solved acquisition, so I wouldn't get too comfortable with it anyways. the next file (`pipeline.yaml`) is just an instantiation of the pipeline (as configured here), which is dumped by scenario 3b

src/autorest-core/resources/default-configuration.md
```diff
@@ -114,11 +114,11 @@ pipeline:
   swagger-document/azure-openapi-validator:
     input:
       - transform
-      - azure-validator # artificial predecessor in order to ensure order of messages for CI purposes
+      - azure-validator
```
olydis: please revert, this file should *not* be changed! the comment is important

dsgouda: Kinda remember a comment saying this is not needed anymore. My bad.

olydis: With https://github.com/Azure/autorest/pull/2326#discussion_r120515648 I meant that the comment was simply ignored and moved to a different place, sort of as if you remove a "please keep door closed" from a door, open the door, and put the sign on a wall.

dsgouda:
@azuresdkci Test this please

dsgouda:
@azuresdkci Test this please

Issue #2290, opened 2017-05-24
Bug when working on a path that contains space


israellot:
It seems that AutoRest has a bug when dealing with folders that contains a space. 
I've tested on a path without spaces and it worked nicely. 
Running with high privileges it will actually create new folders with %20 replacing spaces.

`{ Error: ENOENT: no such file or directory, open 'C:\Users\Israel%20Lot\Source\Repos\yallaplay-services\autorest\Auth\IYallaPlayServicesAuthv11.cs'
    at Error (native)
    at Object.fs.openSync (fs.js:640:18)
    at Object.fs.writeFileSync (fs.js:1333:33)
    at Object.exports.writeFile (C:\Users\Israel Lot\.autorest\plugins\autorest\1.0.1-20170523-1028-preview\node_modules
\autorest-core\lib\ref\async.js:14:63)
    at C:\Users\Israel Lot\.autorest\plugins\autorest\1.0.1-20170523-1028-preview\node_modules\autorest-core\lib\ref\uri
.js:211:23
    at next (native)
    at fulfilled (C:\Users\Israel Lot\AppData\Roaming\npm\node_modules\autorest\lib\polyfill.min.js:14:62)
  errno: -4058,
  code: 'ENOENT',
  syscall: 'open',
  path: 'C:\\Users\\Israel%20Lot\\Source\\Repos\\yallaplay-services\\autorest\\Auth\\IYallaPlayServicesAuthv11.cs' }`

fearthecowboy:
Well, at least it tried!

I'm fixing that right now. Tonight's build should have that fixed.

israellot:
Testing with the new build still produces the same error.

anderslm:
It's fixed in this PR https://github.com/Azure/autorest/pull/2328

PR #2328, opened 2017-06-02
#2290 Decode the returned path to avoid %20 as spaces and such

anderslm:


fearthecowboy:
@azuresdkci test this please

Issue #2333, opened 2017-06-07
SummaryDescriptionValidation severity should be warning


veronicagg:
Please update severity to warning, instead of error. 
Also, please update documentation at https://github.com/Azure/azure-rest-api-specs/blob/master/documentation/openapi-authoring-automated-guidelines.md

veronicagg:
@dsgouda, I assigned it to you, because we talked today about it, if you don't get to it, please assign it to @sarangan12 as he wrote the rule. thanks.

Issue #2313, opened 2017-05-31
Evaluate and assign MergeState and DocumentType validation categories to all validation rules


dsgouda:

vishrutshah:
From the PR https://github.com/Azure/azure-rest-api-specs/pull/1288, we should categorize following rules as ARM specific. Just listing them here for our FYI

- [ ] `APIVersionPattern`
- [ ] `OperationsAPIImplementation`
- [ ] `LROStatusCodesReturnTypeSchema`
- [ ] `PatchInOperationName`
- [ ] `PutInOperationName`

Reviews , please make sure whether this is right and correct on the PR :)

dsgouda:
@vishrutshah please file an issue to this effect it would be easier to track that way.

dsgouda:
PR merged, closing issue.

PR #2334, opened 2017-06-07
Adding categorizations to all rules

dsgouda:
Addressing https://github.com/Azure/autorest/issues/2313

and a bit of https://github.com/Azure/autorest/issues/2333

@vishrutshah @veronicagg @sarangan12 please review whether I have correctly categorized all the rules. If certain rules that are applicable to both `arm` and `data-plane` but not necessarily a part of the `default` validation rules, do point out. Merge state is a little complicated and can probably be discussed offline if you have doubts.
@olydis need your expertise to review some of the refactoring I had to do around the swagger modeler validation tests.


src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
```diff
@@ -42,253 +42,250 @@ internal static void AssertOnlyValidationMessage(this IEnumerable<ValidationMess
...
-                    ServiceDefinitionDocumentType = serviceDefDocType,
-                    MergeState = mergeState
-                };
-                
-                return validator.GetValidationExceptions(new Uri(input, UriKind.RelativeOrAbsolute), serviceDefinition, metaData).OfType<ValidationMessage>();
+                return validator.GetValidationExceptions(new Uri(input, UriKind.RelativeOrAbsolute), serviceDefinition, metadata).OfType<ValidationMessage>();
             }
         }
 
+        private ServiceDefinitionMetadata GetMetadataForRuleTest(Type rType)
```
olydis: proposing slightly more generic solution:
``` C#
private ServiceDefinitionMetadata GetMetadataForRuleTest<TRule>() where TRule : Rule
{
    var rule = Activator.CreateInstance<TRule>();
    ...
```
Like this, `rule` will automatically have type `Rule` 🙂 also, there will be a compiler error if the type is not a rule-type

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
```diff
@@ -42,253 +42,250 @@ internal static void AssertOnlyValidationMessage(this IEnumerable<ValidationMess
...
+            return new ServiceDefinitionMetadata
+            {
+                ServiceDefinitionDocumentType = rule.ServiceDefinitionDocumentType,
+                MergeState = rule.ValidationRuleMergeState
+            };
+        }
+
         [Fact]
         public void MissingDescriptionValidation()
         {
```
olydis: given the structure of all these tests, and the fact that we're now repeating the *same type name twice* I would refactor further. It's even the same directory structure, so how about
``` C#
private IEnumerable<ValidationMessage> GetMessagesForRule<TRule>(string swaggerFileName)
{
    var messages = ValidateSwagger(Path.Combine(Core.Utilities.Extensions.CodeBaseDirectory, "Resource", "Swagger", "Validation", swaggerFileName), GetMetadataForRuleTest<TRule>());
    return messages.Where(message => message.Rule is TRule);
}
```

so that the tests look like
``` c#
var messages = GetMessagesForRule<DescriptionAndTitleMissing>("definition-missing-description.json");
Assert.Equal(2, messages.Count());
```
With this structure, the `AssertOnlyValidationMessage` functions can also be simplified (and partially removed) as they don't filter the message type anymore. Honestly, I'd just make a helper method `JustWarnings` such that you can do
``` c#
Assert.Equal(2, messages.JustWarnings().Count());
```
in places where we had `AssertOnlyValidationWarning`.

dsgouda: Thank you! It took me a good amount of time trying to figure out the best way to generalize this further.

dsgouda:
@olydis, per my understanding most validation  rules can possibly run in the individual mode without knowledge about its composed state (eg.: ```OneUnderscoreInOperationId``` or ```ControlCharacters```) however the more complicated rules like ```TrackedResourceListByImmediateParent``` should be run on the Composed state. 
Thoughts? Suggestions?

olydis:
@dsgouda 100% agree

src/modeler/AutoRest.Swagger/Validation/Extensions/XmsClientNameParameter.cs
```diff
@@ -41,6 +41,18 @@ public class XmsClientNameParameter : TypedRule<SwaggerParameter>
         public override Category Severity => Category.Error;
 
         /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
```
vishrutshah: any generated sdk can leverage this. On that bases, I'd be inclined for `default`

src/modeler/AutoRest.Swagger/Validation/Extensions/XmsClientNameProperty.cs
```diff
@@ -41,6 +41,18 @@ public class XmsClientNameProperty : TypedRule<Dictionary<string, Schema>>
         public override Category Severity => Category.Error;
 
         /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
```
vishrutshah: any generated sdk can leverage this. On that bases, I'd be inclined for `default`

src/modeler/AutoRest.Swagger/Validation/InvalidVerbUsed.cs
```diff
@@ -42,6 +42,18 @@ public class InvalidVerbUsed : TypedRule<Dictionary<string, Operation>>
         public override Category Severity => Category.Error;
 
         /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
```
vishrutshah: why not `default`

src/modeler/AutoRest.Swagger/Validation/NamePropertyDefinitionInParameter.cs
```diff
@@ -35,6 +35,18 @@ public class NamePropertyDefinitionInParameter : TypedRule<SwaggerParameter>
         public override Category Severity => Category.Error;
 
         /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
```
vishrutshah: why ARM only?

src/modeler/AutoRest.Swagger/Validation/OperationDescriptionOrSummaryRequired.cs
```diff
@@ -20,6 +20,18 @@ public class OperationDescriptionOrSummaryRequired : DescriptionRequired<Operati
         public override string MessageTemplate => Resources.MissingSummaryDescription;
 
         /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
```
vishrutshah: I'd go for `Default`

src/modeler/AutoRest.Swagger/Validation/ParameterDescriptionRequired.cs
```diff
@@ -11,6 +11,18 @@ public class ParameterDescriptionRequired : DescriptionRequired<SwaggerParameter
         private static readonly string ParameterTypeFormatter = "'{0}' parameter";
 
         /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
```
vishrutshah: `Default` for better documentation in SDK, CLI docs.microsoft.com ?

src/modeler/AutoRest.Swagger/Validation/PutRequestResponseScheme.cs
```diff
@@ -41,8 +41,20 @@ public class PutRequestResponseScheme : TypedRule<Dictionary<string, Dictionary<
...
+
+        /// <summary>
+        /// When to apply the validation rule, before or after it has been merged as a part of 
+        /// its merged document as specified in the corresponding '.md' file
+        /// By default consider all rules to be applied for After only
+        /// </summary>
+        public override ServiceDefinitionDocumentState ValidationRuleMergeState => ServiceDefinitionDocumentState.Composed;
+
+        /// Verifies if a PUT operation request and response schemas match
+        /// TODO: apply on single spec level and ARM specs only
```
vishrutshah: should we remove `TODO` ?

src/modeler/AutoRest.Swagger/Validation/SummaryAndDescriptionMustNotBeSame.cs
```diff
@@ -31,9 +31,21 @@ public class SummaryAndDescriptionMustNotBeSame: TypedRule<Operation>
         public override string MessageTemplate => Resources.SummaryDescriptionVaidationError;
 
         /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM | ServiceDefinitionDocumentType.DataPlane;
```
vishrutshah: 👍

src/modeler/AutoRest.Swagger/Validation/ValidFormats.cs
```diff
@@ -22,6 +22,18 @@ public class ValidFormats : TypedRule<string>
         public override ValidationCategory ValidationCategory => ValidationCategory.SDKViolation;
 
         /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
```
vishrutshah: what does autorest do if non supported format is provided ? If it blows up then we need ARM | Data-plane, I guess

src/modeler/AutoRest.Swagger/Validation/XmsEnumValidation.cs
```diff
@@ -42,6 +42,18 @@ public class XmsEnumValidation : TypedRule<Dictionary<string, Schema>>
         /// </summary>
         public override Category Severity => Category.Warning;
 
+        /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
```
vishrutshah: should be general?

src/autorest-core/test/end-to-end.ts
```diff
@@ -82,7 +82,7 @@ import { PumpMessagesToConsole } from './test-utility';
 
     autoRest.Message.Subscribe((_, m) => { if (m.Channel === Channel.Warning || Channel.Error) { messages.push(m); } });
     assert.equal(await autoRest.Process().finish, true);
-    assert.equal(messages.length, 0);
+    assert.notEqual(messages.length, 0);
```
olydis: if this is the fix you talk about [here](https://github.com/Azure/autorest/pull/2334/files/2992d5a101a708f6272d92ecc990c3da3b03263d#diff-f1316e08d886e44821dda7ed03690a01R73), that comment may wanna be updated too ;)

src/modeler/AutoRest.Swagger/Validation/AnonymousBodyParameter.cs
```diff
@@ -36,6 +36,18 @@ public class AnonymousBodyParameter : TypedRule<SwaggerParameter>
...
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
+
+        /// <summary>
+        /// When to apply the validation rule, before or after it has been merged as a part of 
+        /// its merged document as specified in the corresponding '.md' file
+        /// By default consider all rules to be applied for After only
+        /// </summary>
+        public override ServiceDefinitionDocumentState ValidationRuleMergeState => ServiceDefinitionDocumentState.Individual;
```
veronicagg: I think I asked a question similar to Vishrut in a prior PR, would it be helpful to define a "default" for the cases where it doesn't matter, so we can refer to it when that's the case, and explicitly call it out when it does?

src/modeler/AutoRest.Swagger/Validation/BodyPropertiesNamesCamelCase.cs
```diff
@@ -48,9 +48,21 @@ public class BodyPropertiesNamesCamelCase : TypedRule<Dictionary<string, Operati
         /// </summary>
         public override Category Severity => Category.Error;
 
-        ///// <summary>
-        ///// Validates whether property names are camelCase in body parameters.
-        ///// </summary>
+        /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.Default;
```
veronicagg: I believe this rule is ARM specific.

src/modeler/AutoRest.Swagger/Validation/DefinitionsPropertiesNamesCamelCase.cs
```diff
@@ -49,6 +49,18 @@ public class DefinitionsPropertiesNamesCamelCase : TypedRule<Dictionary<string,
         public override Category Severity => Category.Error;
 
         /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.Default;
```
veronicagg: I believe this rule is ARM specific

src/modeler/AutoRest.Swagger/Validation/Extensions/NextLinkPropertyMustExist.cs
```diff
@@ -22,6 +23,18 @@ public class NextLinkPropertyMustExist : PageableExtensionRule
         public override ValidationCategory ValidationCategory => ValidationCategory.SDKViolation;
 
         /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
```
veronicagg: if the extension is used, then this should apply, so I think it should be default

src/modeler/AutoRest.Swagger/Validation/OperationIdNounConflictingModelNames.cs
```diff
@@ -26,6 +26,18 @@ public class OperationIdNounConflictingModelNames : TypedRule<string>
         public override ValidationCategory ValidationCategory => ValidationCategory.SDKViolation;
 
         /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
```
veronicagg: is this ARM only? I had thought the issue come in due to the way AutoRest picks up these names. I know there's a recent thread on this rule in terms of being a real conflict or not, so we should figure that out.

src/modeler/AutoRest.Swagger/Validation/OperationIdSingleUnderscore.cs
```diff
@@ -37,6 +38,18 @@ public override bool IsValid(string entity)
         public override string MessageTemplate => Resources.OnlyOneUnderscoreAllowedInOperationId;
 
         /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
```
veronicagg: is this ARM only? isn't this an AutoRest limitation?

src/modeler/AutoRest.Swagger/Validation/UniqueResourcePaths.cs
```diff
@@ -37,6 +37,18 @@ public class UniqueResourcePaths : TypedRule<Dictionary<string, Dictionary<strin
         public override string MessageTemplate => Resources.UniqueResourcePaths;
 
         /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.Default;
```
veronicagg: I think this maybe ARM specific.

src/modeler/AutoRest.Swagger/Validation/AnonymousBodyParameter.cs
olydis: the default should be "composed (as described in some other comments) which is also why that is the virtual property's value, and yes, I 100% agree we should only specify *deviations* from that default. I expect "composed" to be correct 90% of the time.

src/modeler/AutoRest.Swagger/Validation/AvoidAnonymousTypes.cs
```diff
@@ -34,6 +34,18 @@ public class AvoidAnonymousTypes : TypedRule<SwaggerObject>
         public override string MessageTemplate => Resources.AnonymousTypesDiscouraged;
 
         /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
```
dsgouda: An openapi doc can have anonymous types in parameters, the restriction is ARM recommendation if I am not wrong

src/modeler/AutoRest.Swagger/Validation/AvoidMsdnReferences.cs
```diff
@@ -21,6 +21,18 @@ public class AvoidMsdnReferences : TypedRule<string>
         /// </summary>
         public override ValidationCategory ValidationCategory => ValidationCategory.SDKViolation;
 
+        /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
```
dsgouda: So the way I look at it is, 
- Default - All rules that could be applied to a simple OpenAPI document 
- ARM - All rules speced out by the ARM folks
- Data-Plane - All rules applicable to the data-plane specs

Something like this is applicable to both ARM and Data-plane but isn't necessarily a Default, in which case we could apply multiple flags to the validation rule.

src/modeler/AutoRest.Swagger/Validation/AvoidNestedProperties.cs
```diff
@@ -52,5 +52,17 @@ private static bool IsClientFlattenUsed(Dictionary<string, object> extensions)
         ///     The severity of this message (ie, debug/info/warning/error/fatal, etc)
         /// </summary>
         public override Category Severity => Category.Warning;
+
+        /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
```
dsgouda: Should we? Can we consult someone on such rules.

src/modeler/AutoRest.Swagger/Validation/BodyPropertiesNamesCamelCase.cs
dsgouda: Not really, communication over REST channels that rely on json and Json is usually camelCased.

src/modeler/AutoRest.Swagger/Validation/Core/Rule.cs
```diff
@@ -25,14 +25,14 @@ protected Rule()
         /// <summary>
         /// What kind of open api document type this rule should be applied to
         /// </summary>
-        public virtual ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
+        public abstract ServiceDefinitionDocumentType ServiceDefinitionDocumentType { get; }
```
dsgouda: A. Defaults make us lazy and open the door to rely on them when adding new rules.
B. When explicitly having to set these properties we will end up doing our due diligence and understand the impact and implications of the validation rule.

src/modeler/AutoRest.Swagger/Validation/DefinitionsPropertiesNamesCamelCase.cs
dsgouda: Same comment as before for Jsons

src/modeler/AutoRest.Swagger/Validation/DeleteMustNotHaveRequestBody.cs
```diff
@@ -43,6 +43,18 @@ public class DeleteMustNotHaveRequestBody : TypedRule<Dictionary<string, Operati
         public override Category Severity => Category.Error;
 
         /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
```
dsgouda: Same point as before

src/modeler/AutoRest.Swagger/Validation/AnonymousBodyParameter.cs
```diff
@@ -41,9 +41,12 @@ public class AnonymousBodyParameter : TypedRule<SwaggerParameter>
...
         /// <summary>
-        /// When to apply the validation rule, before or after it has been merged as a part of 
-        /// its merged document as specified in the corresponding '.md' file
-        /// By default consider all rules to be applied for After only
+        /// Whether the rule should be applied to the individual or composed context based on
+        /// the corresponding .md file
+        /// In most cases this should be composed
+        /// This is because validation rules that run in individual mode will end up
+        /// throwing multiple validation messages for the same violation if related model/property,etc 
+        /// was referenced in multiple files
```
veronicagg: I wouldn't copy the whole explanation in a specific rule, a comment for the rule IMO, should justify why the value was picked.

src/modeler/AutoRest.Swagger/Validation/AvoidAnonymousTypes.cs
veronicagg: I think this is more of an AutoRest recommendation, so that we know the name of the class and don't make one up, I would probably do ARM and Data Plane, if we wanted to leave default out.

src/modeler/AutoRest.Swagger/Validation/AvoidMsdnReferences.cs
veronicagg: should the code be updated to include data plane?

src/modeler/AutoRest.Swagger/Validation/AvoidNestedProperties.cs
veronicagg: In this case, I'd be incline to add data plane too. For other rules, sure, we can discus and figure out.

src/modeler/AutoRest.Swagger/Validation/OperationDescriptionOrSummaryRequired.cs
veronicagg: If we don't do default here, let's add data plane

src/modeler/AutoRest.Swagger/Validation/DescriptionAndTitleMissing.cs
```diff
@@ -22,6 +22,18 @@ public class DescriptionAndTitleMissing : DescriptionRequired<Dictionary<string,
         public override string MessageTemplate => Resources.MissingTitleDescription;
 
         /// <summary>
+        /// What kind of open api document type this rule should be applied to
+        /// </summary>
+        public override ServiceDefinitionDocumentType ServiceDefinitionDocumentType => ServiceDefinitionDocumentType.ARM;
```
veronicagg: I agree on Default, if there are concerns about that, let's add data plane.

src/modeler/AutoRest.Swagger.Tests/SwaggerModelerValidationTests.cs
```diff
@@ -15,619 +15,603 @@
...
+            Assert.Equal(messages.Count(), 1);
         }
 
         [Fact]
         public void Pageable200ResponseNotModeledValidation()
         {
-            var messages = ValidateSwagger(Path.Combine(Core.Utilities.Extensions.CodeBaseDirectory, "Resource", "Swagger", "Validation", "pageable-no-200-response.json"));
-            messages.Any(m => m.Rule.GetType() == typeof(PageableRequires200Response));
+            var messages = GetValidationMessagesForRule<PageableRequires200Response>("pageable-no-200-response.json");
+            Assert.True(true);
```
dsgouda: Yes, just a placeholder since this test never did anything, opened issue for this.

src/modeler/AutoRest.Swagger/Validation/AvoidMsdnReferences.cs
dsgouda: A little unsure actually, feels like something ARM recommended but also think that the documentation team were involved in requesting this rule.

src/modeler/AutoRest.Swagger/Validation/DescriptionAndTitleMissing.cs
dsgouda: Will add dataplane

src/modeler/AutoRest.Swagger/Validation/OperationIdSingleUnderscore.cs
dsgouda: Agreed, I should add DataPlane here.

src/modeler/AutoRest.Swagger/Validation/AnonymousBodyParameter.cs
dsgouda: You mean explanation per rule?

veronicagg: I wouldn't expect the general explanation to be copied in each rule file, having the general explanation where the definition of the value is should be enough. If you wanted to have a comment per rule, you could indicate why you picked individual vs composed for that rule (for example).

dsgouda: I will try to add an explanation for the special cases and leave a generic comment otherwise

PR #2350, opened 2017-06-09
Added automated guidelines

kirthik:
This is copied from azure-rest-api-specs repo. Once this is merged, I will remove the page from azure-rest-api-specs repo.

cc @veronicagg

### DIFF
--- a/docs/developer/validation-rules/openapi-authoring-automated-guidelines.md
+++ b/docs/developer/validation-rules/openapi-authoring-automated-guidelines.md
@@ -1,0 +1,264 @@
+# OpenAPI Specifications Authoring - Automated Guidelines #
+
+This document lists the set of automated rules that can be validated against OpenAPI(swagger) spec by running [validation tools](https://github.com/Azure/adx-documentation-pr/wiki/OpenAPI-Validation-tools). Please visit [here for Manual guidelines](https://github.com/Azure/azure-rest-api-specs/blob/master/documentation/openapi-authoring-manual-guidelines.md).
+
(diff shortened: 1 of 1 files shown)


docs/developer/validation-rules/openapi-authoring-automated-guidelines.md
```diff
@@ -0,0 +1,264 @@
+# OpenAPI Specifications Authoring - Automated Guidelines #
+
+This document lists the set of automated rules that can be validated against OpenAPI(swagger) spec by running [validation tools](https://github.com/Azure/adx-documentation-pr/wiki/OpenAPI-Validation-tools). Please visit [here for Manual guidelines](https://github.com/Azure/azure-rest-api-specs/blob/master/documentation/openapi-authoring-manual-guidelines.md).
```
veronicagg: I think this document is expected to contain linter rules only, not other validation tools, the main reason IMO to move the doc here is to keep it with the code. When the linter gets split out of the autorest repo (which is planned) then the documentation of the rules should move with it. 
I'd recommend to have a document in the azure-rest-api-specs repo that points to this and probably other automated tools documentation. 
Feel free to let me know if you disagree or would like me to make the change.

olydis:
@kirthik @veronicagg This is 3 months old, what's the plan? 😃

veronicagg:
@olydis we should close this PR, the document lives currently at https://github.com/Azure/azure-rest-api-specs/blob/current/documentation/openapi-authoring-automated-guidelines.md and should be now moved to https://github.com/Azure/azure-openapi-validator since that's where the rules live. We'll do the move as needed.

Issue #2363, opened 2017-06-14
AddCredentials argument is incorrect in auth.md docs


ericstj:
Was reading through this doc and noticed https://github.com/Azure/autorest/blob/master/docs/client/auth.md lists `-AddCredentials` but the actual arugument is `--add-credentials`

ericstj:
Also lots of broken links in this same document.

olydis:
thanks, on it!

PR #2364, opened 2017-06-14
fix docs

olydis:
fixes https://github.com/Azure/autorest/issues/2363


olydis:
@fearthecowboy thanks! 😃

Issue #2316, opened 2017-05-31
Resource identification logic needs to be modified to exclude models returned by a POST


ravbhatnagar:
If a model is getting returned only by a POST, it should not be considered as a resource. This should apply even when the model has “x-ms-azure-resource” extension applied. If it is getting returned by GET/PUT in addition to POST, then it should be considered as a resource.

PR #2365, opened 2017-06-14
Filtering resource models returned only by post

dsgouda:
Addressing https://github.com/Azure/autorest/issues/2316


src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -57,9 +57,16 @@ public static IEnumerable<string> GetResourceModels(ServiceDefinition serviceDef
                                                     .Where(modelName => !(IsBaseResourceModelName(modelName))
                                                                         && serviceDefinition.Definitions.ContainsKey(modelName)
                                                                         && IsAllOfOnModelNames(modelName, serviceDefinition.Definitions, xmsAzureResourceModels));
+
+            var unfilteredResourceCandidates = resourceModels.Union(modelsAllOfOnXmsAzureResources);
+
+            // Now filter all the resource models that are returned from a POST operation only
```
olydis: I don't see the "only" part reflected in the code but maybe I'm missing it.
Also, the indentation is misleading: the `.SelectMany` calls are perfectly aligned but are *not* operating on the same "level". I'd expect the second `SelectMany` to be on the same height as the inner `Where`, just break `pathObj => pathObj<HERE>.Where` and lines won't be that long.

dsgouda: d'oh (I went everywhere looking for a Homer emoji!)

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
```diff
@@ -57,10 +57,18 @@ public static IEnumerable<string> GetResourceModels(ServiceDefinition serviceDef
...
-            // return the union 
-            return resourceModels.Union(modelsAllOfOnXmsAzureResources);
 
+            var unfilteredResourceCandidates = resourceModels.Union(modelsAllOfOnXmsAzureResources);
+
+            // Now filter all the resource models that are returned from a POST operation only 
+            var postOpResourceModels = serviceDefinition.Paths.Values.SelectMany(pathObj => pathObj.Where(opObj => opObj.Key.EqualsIgnoreCase("post"))
+                                                                        .SelectMany(opObj => opObj.Value.Responses?.Select(resp => resp.Value?.Schema?.Reference?.StripDefinitionPath())))
+                                                                     .Where(model => !string.IsNullOrWhiteSpace(model))
+                                                                     .Except(putOperationsResponseModels)
```
olydis: don't implement the "only" by relying on other sets and excepting them.
1) you already missed some HTTP methods, what about PATCH?
2) it's not obvious that `putOperationsResponseModels` contains *all* models of PUT methods. Next month, someone applies a *fix* to something by changing that set, not realizing that it breaks this line. Locality 😉

veronicagg: how about looking into the models returned by operations and selecting the ones that are only returned by posts? then removing those from the previous set? I would probably not call it "unfiltered" set, since it's already quite filtered, maybe "resourceCandidates".

vishrutshah:
I do not have any other comments apart from what's already been commented. Feel free to discuss with them and I'd be fine on the decision. I am not posting "Approve" / "Request Changes" as I'll be OOF tomorrow so if it gets resolved, go ahead :)

src/modeler/AutoRest.Swagger/Model/ValidationUtilities.cs
dsgouda: I literally called it `resourceCandidates` before changing my mind 😆

dsgouda: @veronicagg that is what I am precisely trying to do
@olydis from what @ravbhatnagar indicated on the issue, I should really be ignoring anything returned by a PUT/GET operation only.
Also, what is being populated in the `putOperationsResponseModels` which are basically resource model candidates and this particular filtering is applicable only to the resource models, would I still miss out on anything here?

olydis: sounds good, how does the rule perform on the specs repo?

dsgouda: Well, this is not a rule per se, more like pre cursor step to running any arm related validation rules.
This is a one off operation and the results are stored in the context object passed to every single rule.
This makes me wonder if we should run this logic even if we do not have any arm specific rules to run, given how expensive these operations are its worth a look. what say?

veronicagg:
@dsgouda apart from updating the name to "resourceCandidates" which I'm not seeing, rest looks good to me. I'd be good to run some of the resource related rules on the repo and make sure we're accurate. Since the rule won't get released right away (not as "latest"), if you'd done validation already you could run on the repo afterwards with the nightly build. thanks!

dsgouda:
Thanks @veronicagg I will run the changes against the rest spec repo

dsgouda:
Ran this against a few specs in the rest spec repo, and I still see the resources being picked as expected, we can investigate this further if required.

src/modeler/AutoRest.Swagger/Validation/Core/RuleContext.cs
```diff
@@ -19,12 +19,16 @@ public class RuleContext
...
         /// </summary>
         /// <param name="root"></param>
-        public RuleContext(ServiceDefinition root, Uri file) : this(null)
+        public RuleContext(ServiceDefinition root, Uri file, ServiceDefinitionMetadata metadata) : this(null)
         {
             this.Root = root;
             this.Value = root;
             this.File = file;
-            PopulateResourceTypes(root);
+            if (metadata.ServiceDefinitionDocumentType == ServiceDefinitionDocumentType.ARM)
```
olydis: this change breaks CI:
https://github.com/Azure/autorest/blob/master/src/modeler/AutoRest.Swagger/Validation/BodyTopLevelProperties.cs#L52 fails because `resModels` is `null`. Since that didn't happen before, I bet that it is this guard that causes the `context.ResourceModels` to not be populated...
Now, I see that both this guard *and* the rule say that they are `ARM`, but still, this is kind of the only potential point of failure I could make out.

olydis: further observation: this `==` will fail if `metadata.ServiceDefinitionDocumentType` are multiple types at once!
Instead of duplicating the logic for document type filtering, I'd honestly just make `ResourceModels` a lazily initialized property! If someone asks for it, populate it, otherwise don't. That way, you automatically inherit the correct behavior from the fact that the rules are classified correctly.

dsgouda: hmm... starting to think if there is any value in adding the check after all.

olydis: fair point, don't think anyone severely suffers because of this... main customer is still ARM, for data plane I also assume this goes faster since there is actually not much that looks like a resource. And even then, what's the actual overhead of this? Would be surprised if it took half a second on most Swaggers.

dsgouda: Agreed, even if it does warrant a performance investigation, I think we can revisit this once we move validator to a different repo,

PR #2366, opened 2017-06-15
reenable python tests

olydis:


olydis:
lol, the problem must have been transient. We had Python tests failing consistently and across both Travis and Jenkins yesterday!

PR #2369, opened 2017-06-19
Content type overrides for streams

olydis:
As specified in more detail in `Samples/test/stream-with-content-type/readme.md`, there are 2 orthogonal fixes in this PR:
- make C# generator not fail for `Content-Type` header param, and actually use it if the body type is `Stream` (trivial to enable for other body types, but saw no point so far! If you pass JSON, it's not gonna be XML...)
- provide smart defaults for header param named `Content-Type`. Namely, if you haven't specified an enum on it, it inherits the `consumes` section. This behavior enables deduplication of data in the OpenAPI definition and is no violation of the OpenAPI standard... since `Content-Type` parameters are really to be ignored.


Samples/1a-code-generation-minimal/shell/stderr.txt
```diff
@@ -0,0 +1 @@
+
```
olydis: side effect of shelljs's `sort` on the file...

src/modeler/AutoRest.Swagger/OperationBuilder.cs
```diff
@@ -243,17 +235,30 @@ private void BuildMethodParameters(Method method)
...
+                    // enrich Content-Type header with "consumes"
+                    if (actualSwaggerParameter.Enum == null && 
+                        swaggerParameter.Extensions.GetValue<JObject>("x-ms-enum") == null &&
+                        _effectiveConsumes.Count > 1)
+                    {
+                        actualSwaggerParameter.Enum = _effectiveConsumes.ToList();
+                        actualSwaggerParameter.Extensions["x-ms-enum"] = 
+                            JObject.FromObject(new
+                            {
+                                name = "ContentType",
```
fearthecowboy: So.... what if there are two APIs that take different content types? 

Both APIs will expect their `ContentType` enum class to be called "`ContentType`" ? (and end up in `models`?)

PR #2378, opened 2017-06-20
New JSON RPC client which generates declarative code instead of imperative

sergey-shandar:


olydis:
@azuresdkci add to whitelist

olydis:
@azuresdkci test this please

Issue #2389, opened 2017-06-22
Add Message about tenant level resources to ListByResourceGroup and ListBySubcription 


sarangan12:
Related to Issue https://github.com/Azure/autorest/issues/2170

sarangan12:
Task completed

PR #2390, opened 2017-06-22
Add message regarding tenant level resources to ListByResourceGroup & ListBySubscriptions

sarangan12:
Please review and approve Related to https://github.com/Azure/autorest/issues/2389

### DIFF
--- a/src/core/AutoRest.Core/Properties/Resources.resx
+++ b/src/core/AutoRest.Core/Properties/Resources.resx
@@ -251,10 +251,10 @@ Modelers:
     <value>For the tracked resource '{0}', the x-ms-pageable extension values must be same for list by resource group and subscriptions operations.</value>
   </data>
   <data name="TrackedResourceListByResourceGroupOperationMissing" xml:space="preserve">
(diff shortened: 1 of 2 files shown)


src/core/AutoRest.Core/Properties/Resources.resx
```diff
@@ -251,10 +251,10 @@ Modelers:
     <value>For the tracked resource '{0}', the x-ms-pageable extension values must be same for list by resource group and subscriptions operations.</value>
   </data>
   <data name="TrackedResourceListByResourceGroupOperationMissing" xml:space="preserve">
-    <value>The tracked resource, '{0}', must have a list by resource group operation.</value>
+    <value>The tracked resource, '{0}', must have a list by resource group operation.(This rule does not apply for tenant level resources.)</value>
```
vishrutshah: Can we detect tenant level resource from the `path` of the operation to improve the logic of the rule?

sarangan12: Not as of now. This was decided in one of our earlier discussion

PR #2421, opened 2017-07-06
Removing CSharp validator [Do Not merge]

dsgouda:
Works with the new CSharp/typescript validator


olydis:
how? 😆 

image

dsgouda:
Disregard those, number of different branches I'm working on

src/autorest-core/package.json
```diff
@@ -1,10 +1,7 @@
 {
-  "name": "@microsoft.azure/autorest-core",
```
olydis: this file was reverted to an earlier state, important information is lost! make sure you have a clean commit history, that should resolve issues like these - manually checking all these changes here for regressions will be a nightmare.

Actually, the PR you closer earlier was clean! I last rebased it on top of `2.0` *today*. What caused you to reopen it like this? AFAIK all that was missing was moving the docs folders and adjusting the core to acquire the plugin dynamically instead of statically.

dsgouda: Let me see what I can do

PR #2436, opened 2017-07-12
UNC fix

olydis:
UNC paths `\\asd\qwe\rty` were incorrectly converted to `file:////asd/qwe/rty` instead of (according to standard) `file://asd/qwe/rty`.
Fixing that conversion magically made everything work, all other libraries we use treat those URLs appropriately.

Also improved the message when trying to write files to an UNC location (which did not fail before but also put all stuff to the local folder instead of the UNC location... now we fail with a meaningful message)


olydis:
@fearthecowboy It was python again...

PR #2438, opened 2017-07-14
fence bug fix

olydis:
Fence checking was kind of not reentrant. Sometimes it would skip code blocks although their condition is the same as that of previous ones and such. Of course it also depends on the order of code blocks. I had fun tracking this one down.

Note: This bug currently prevents some SDKs to be generated from configuration files!


src/autorest-core/lib/ref/yaml.ts
```diff
@@ -90,6 +90,9 @@ export function ResolveAnchorRef(yamlAstRoot: YAMLNode, anchorRef: string): YAML
  * Populates yamlNode.valueFunc with a function that creates a *mutable* object (i.e. no caching of the reference or such)
  */
 function ParseNodeInternal(yamlRootNode: YAMLNode, yamlNode: YAMLNode, onError: (message: string, index: number) => void): () => any {
+  if (!yamlNode) {
```
olydis: cosmetic fix: js-yaml seems to sometimes use `null` to represent `null` in YAML, although there is a corresponding `YAMLNode` instantiation for that too... Occured for me for the first array item of:
```
asd:
- 
- null
```
Note that both array items are valid and identical.

src/autorest-core/lib/source-map/merging.ts
```diff
@@ -115,7 +115,7 @@ export function resolveRValue(value: any, propertyName: string, higherPriority:
 
     // resolve the macro value for strings
     if (typeof value === "string") {
-      const match = macroRegEx.exec(value.trim());
+      const match = macroRegEx().exec(value.trim());
```
fearthecowboy: Ah. Yeah. damn stateful regexes.

olydis:
@fearthecowboy Thanks for the quick review! What the heck is going on with Travis, gets 504's for some nuget package (consistently, restarted 4x already). Jenkins is fine except for the `1.1.0` => `1.2.0` version bump. We should probably publish a fix to the stable release, so I'll leave it to you when to merge this 😉

fearthecowboy:
Yeah, I'll repub 1.2.1 after this is done. If I don't get it tonight, I'll do it first thing in the AM

olydis:
@fearthecowboy have regenerated. Note that I have also added another commit that adds a `--clear-output-folder` option so generate.cmd's no longer need to know the path of generated code. I've only activated the feature for C# for now, it is not overly engineered, but it gets the job done and is IMHO save. Note that the deletion is guarded by both `can-clear-output-folder` (to be set on the emitter's scope, see default config) and `clear-output-folder` (to be set on CLI or readme.md), so there should be absolutely no risk that this affects anything that is already out there.

Just thought I'd add this now since it'll allow me to make [this](https://github.com/Azure/azure-sdk-for-net/pull/3489/files) significantly simpler.

PR #2443, opened 2017-07-14
Batch mode

olydis:


olydis:
@fearthecowboy woohooo it works! With this, I could simplify even the complicated SDK `generate.cmd`s to a one liner and it regenerates successfully 😄 

Resource manager (formerly 5 separate calls): `autorest.cmd cognitiveservices/resource-manager`
Network (formerly 1 call with composition): `autorest.cmd network/resource-manager`

(The `autorest.cmd` just does the `npm install -g` and assembles the config file URI)

🎉 🎉 🎉 🎉 🎉

fearthecowboy:
SLICK.

PR #2455, opened 2017-07-18
Fixes to the markdown generator as requested by the doc team.

jhendrixMSFT:
Strip out embedded new-line characters in descriptions.
Place child documents in nested directories and generate appropriate
inline links to them.
Remove the "null" text for empty resource arrays.


fearthecowboy:
We don't really need to add a `skip-props` -- that can be done with configuration:

~~~ markdown

``` yaml
directive: 
  # find the SomeClass.SomeProperty property 
  where: $..SomeClass.properties.SomeProperty
  # remove the property entirely.
  transform: return undefined; 
```
~~~

It would be trivial to add a `delete: ` directive so that it wasn't a generic `transform:` as this is the kind of thing I think we'd probably do more than once or twice.

jhendrixMSFT:
@fearthecowboy Interesting, I didn't know you could do that.  Does it scale nicely for multiple properties?  The doc team has a list of 10 or so that they want to omit.

fearthecowboy:
Sure, you could do something like: 

~~~ markdown

# recommended way:

``` yaml
directive: 
  - from: myswagger.json # probably should specify what document, otherwise it's trying the transform at every step
    where: $..ClassOne.properties.SomeProperty
    transform: return undefined; 

  - from: myswagger.json
    where: $..ClassTwo.properties.SomeProperty
    transform: return undefined; 

  - from: myswagger.json
    where: $..ClassThree.properties.SomeProperty
    transform: return undefined; 
```

## not so recommended

``` yaml 
# if it is the same property in several places -
# you can just say that property name (as long as it's not too greedy!)
directive: 
  - from: myswagger.json # REALLY specify what document, otherwise it's trying the transform at every step!
    where: $..SomeProperty #selects all nodes in the swagger that were called "SomeProperty"
    transform: return undefined; 
```

~~~


The virtue with configuration transformations is that it documents what you're doing (and you can add more docs in the markdown to say 'why'

fearthecowboy:
The where clause is a [JSONPath](https://www.npmjs.com/package/JSONPath) expression, so you can do stuff like:

(quick correction -- rather than using the name of the swagger file (ie, `mywagger.json` , use `swagger-document` -- that operates on the merged swagger file if there are multiple inputs)

~~~ markdown

``` yaml
directive:
  - from: swagger-document # <- this is any 'document' or 'input-artifact' in the pipeline
    where: $..properties.prop1   # <- any JSON Path -- in this case - grab any property named `prop1`
    transform: return undefined; # any legal javascript snippet returning a value. (the selected node is passed in as $)
```
~~~

Implementation notes: 

See: 
Literate Configuration: https://github.com/Azure/autorest/blob/master/docs/user/literate-file-formats/configuration.md (you can see the possible `input-artifact`s in there)

The default 'builtin' configuration that constructs the pipeline: https://github.com/Azure/autorest/blob/aa49bc0b010963c2acf34df79cf2f8d149b4726b/src/autorest-core/resources/default-configuration.md

PR #2446, opened 2017-07-17
Generate Managers and POMs for Azure SDK for Java

RikkiGibson:
Adds some templates to generate stub Manager classes. The generation of the Manager and POM is controlled by the RegenerateManager flag. Note that the path provided to Autorest in the Azure SDK for Java project has changed so that we can reasonably write the POM to the correct location.

The ServiceName property is something that we just expect to work most of the time due to the convention for paths in Azure, as well as the fact that we expect each service we generate code for to have at least one method.

PR #2467, opened 2017-07-21
(2.0) Generate managers and POMs for Azure SDK for Java

RikkiGibson:
This is a repeat of #2446, but based on the 2.0 branch instead of master.

> Adds some templates to generate stub Manager classes. The generation of the Manager and POM is controlled by the RegenerateManager flag. Note that the path provided to Autorest in the Azure SDK for Java project has changed so that we can reasonably write the POM to the correct location.
> 
> The ServiceName property is something that we just expect to work most of the time due to the convention for paths in Azure, as well as the fact that we expect each service we generate code for to have at least one method.


RikkiGibson:
@olydis any idea why CI is failing here? Looks like it's failing similarly on Azure/2.0 branch.

olydis:
@RikkiGibson Making Travis work is WIP, Jenkins however discovered changes in generated `*.java` files. Please run `gulp regenerate`. This check is in place to prevent accidental changes to generated code - PRs are essentially forced to also contain some samples of (diffs to) generated code so they can be checked for sanity.

RikkiGibson:
Got it. That makes a lot of sense. I'll be augmenting this PR to add generated managers and project files for .NET and then committing the regenerated samples for everything.

RikkiGibson:
@olydis I've merged in the latest from 2.0 and run gulp regenerate. Does anything about the diff of gulp regenerate (last commit in the PR) seem amiss?

olydis:
@RikkiGibson yes, this seems weird, but don't worry about regeneration at this point! Once the repo's are split (working on Java as we speak), everything will make way more sense.

RikkiGibson:
I'm seeing this error from the Jenkins build:
```
17:28:01   FATAL: System.IO.IOException: File not found: src/main/java/fixtures/azurereport/implementation/AutoRestReportServiceForAzureImpl.java
```

Is there anything blocking this from being merged? Anything I can do?

olydis:
Huh, this was not the case before so I assume it is related to this PR. At this point it's probably best to reopen cleanly against https://github.com/Azure/autorest.java (will have that repo populated tomorrow!), note that your PR also changes generated `cs` files that should really not have been affected by the changes you made, looks to me as if the other languages' generators weren't the most recent ones (or not compiled) when you did `gulp regenerate`. This kind of stuff (always having to pull latest changes for everything - we have 8 languages now) is annoying pretty much everyone working on AutoRest, which is exactly why stuff is split now. I'm essentially done, just some testing and automation scripts. I'll gladly help retargeting your PR against the new repo - I suspect it'll immediately pass CI (that only does Java stuff, yay!) there 😉

RikkiGibson:
Sounds excellent. I'll follow up on that early next week.

olydis:
replayed Java parts of it at https://github.com/Azure/autorest.java/pull/2
replayed C# parts of it at TODO

PR #2468, opened 2017-07-21
Fix d.ts files for nodejs code gen

amarzavery:


amarzavery:
@balajikris please review

fearthecowboy:
@amarzavery I found out why CI is giving you trouble in Jenkins right now, I've got a fix in CI right now to clean that up. Once that is committed, I'll re-run this in CI and I'm pretty sure it'll pass too.

amarzavery:
@fearthecowboy - thanks a ton

PR #2488, opened 2017-07-31
Test ClientRuntime fix 3503

olydis:
testing https://github.com/Azure/azure-sdk-for-net/pull/3503

BEFORE MERGE:
- publish new CR
- remove local copy of CR
- point to new Nuget package


olydis:
moved to https://github.com/Azure/autorest.csharp/pull/9

PR #2490, opened 2017-08-01
[Go] client-side-validation flag support

mcardosos:
PTAL @marstr @jhendrixMSFT @olydis 
Fixes https://github.com/Azure/autorest/issues/2100 for Go generator

### DIFF
--- a/src/generator/AutoRest.Go/GeneratorSettingsGo.cs
+++ b/src/generator/AutoRest.Go/GeneratorSettingsGo.cs
@@ -12,5 +12,10 @@ namespace AutoRest.Go
         public virtual string Name => "Go";
 
         public virtual string Description => "Generic Go code generator.";
+
+        /// <summary>
+        /// Indicates whether to generate client side validation.
+        /// </summary>
+        public bool ClientSideValidation { get; set; }
     }
 }
--- a/src/generator/AutoRest.Go/Model/MethodGo.cs
+++ b/src/generator/AutoRest.Go/Model/MethodGo.cs
@@ -216,7 +216,7 @@ namespace AutoRest.Go.Model
 
         public bool ShouldValidate()
         {
-            return Singleton<GeneratorSettingsGo>.Instance.ClientSideValidation;
+            return Singleton<GeneratorSettingsGo>.Instance.ClientSideValidation && !ParameterValidations.IsNullOrEmpty();
         }
 
         public string NextMethodName => $"{Name}NextResults";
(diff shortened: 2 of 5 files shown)


src/generator/AutoRest.Go/GeneratorSettingsGo.cs
```diff
@@ -12,5 +12,10 @@ public class GeneratorSettingsGo : IsSingleton<GeneratorSettingsGo>, IGeneratorS
         public virtual string Name => "Go";
 
         public virtual string Description => "Generic Go code generator.";
+
+        /// <summary>
+        /// Indicates whether to generate client side validation.
+        /// </summary>
+        public bool ClientSideValidation { get; set; }
```
jhendrixMSFT: Where is this set?

src/generator/AutoRest.Go/Model/MethodGo.cs
```diff
@@ -213,6 +214,10 @@ public string MethodReturnSignature(bool helper)
             return $"{retVal}, {errVal}";
         }
 
+        public bool ShouldValidate() {
+            return !ParameterValidations.IsNullOrEmpty();
```
jhendrixMSFT: Nit, opening curly brace is in the wrong location.  More importantly shouldn't this return Singleton<GeneratorSettingsGo>.Instance.ClientSideValidation?  Then you can omit the changes in ParameterGo.cs.

mcardosos: I like the changes in ParameterGo.cs. Why gen code that wont be written to the file anyways. Still will change code here.

---
In reply to: [130652897](https://github.com/Azure/autorest/pull/2490#discussion_r130652897) [](ancestors = 130652897)

src/generator/AutoRest.Go/GeneratorSettingsGo.cs
mcardosos: AutoRest core

src/generator/AutoRest.Go/Model/MethodGo.cs
mcardosos: Fixed

jhendrixMSFT: But will ParameterGo.Validate() actually get called when ShouldValidate() == false?  I would think it not to get called.

olydis:
moved to https://github.com/Azure/autorest.go/pull/6

PR #2492, opened 2017-08-02
[Go] TestGetUnderflowDateTimeRFC1123 changed due to changes in Go1.8

mcardosos:
Now that Jenkins is running go 1.8 I noticed this problem.
[Go1.8 Release Notes](https://golang.org/doc/go1.8) Parsing this impossible dates now returns an error.

PTAL @marstr @jhendrixMSFT


.travis.yml
```diff
@@ -62,7 +62,7 @@ script:
   - gulp build
   - |-
     if [[ $MODE == 'main' ]]; then
-      gulp test
+      travis_wait 30  gulp test
```
jhendrixMSFT: Why is the wait necessary and how do we know that 30 is good enough?

olydis:
moved to https://github.com/Azure/autorest.go/pull/5

PR #2505, opened 2017-08-07
Add expandable string enum to template2

daschult:
Update Java EnumTemplate.cshtml to make string-based enum types extends ExpandableStringEnum.


daschult:
These checks are failing because [this pull request](https://github.com/Azure/autorest-clientruntime-for-java/pull/184) hasn't been accepted yet. Once it gets accepted and published, then these checks will start picking up the correct Maven packages and everything should work.

olydis:
moved to https://github.com/Azure/autorest.java/pull/3

PR #2508, opened 2017-08-08
PHP

sergey-shandar:
PHP generator


olydis:
@sergey-shandar Regeneration produces a diff (but it seems unrelated to upgrading YamlDotNet), @fearthecowboy any clue why Jenkins isn't even listed as CI here?

fearthecowboy:
Lemme check what group it should be using...

fearthecowboy:
@azuresdkci add to whitelist

PR #2515, opened 2017-08-10
expose traffic through inspection interface

olydis:
This:
image


fearthecowboy:
> win32-x64 foo

olydis:
moved to https://github.com/Azure/autorest/pull/2628

Issue #1355, opened 2016-08-16
Header and Option classes in C# should have a better /// comment


matthchr:
Today for example they read:

```
    /// <summary>
    /// Defines headers for Reactivate operation.
    /// </summary>
    public partial class TaskReactivateHeaders
```

```
    /// <summary>
    /// Additional parameters for the Task_Reactivate operation.
    /// </summary>
    public partial class TaskReactivateOptions
```

it would be better if they read:

```
    /// <summary>
    /// Defines headers for the Task Reactivate operation.
    /// </summary>
    public partial class TaskReactivateHeaders
```

```
    /// <summary>
    /// Additional parameters for the Task Reactivate operation.
    /// </summary>
    public partial class TaskReactivateOptions
```

dsgouda:
Addressed by the merged PR

PR #2521, opened 2017-08-12
Better commenting for addl params

dsgouda:
Addresses https://github.com/Azure/autorest/issues/1355


src/modeler/AutoRest.Swagger/SwaggerModeler.cs
```diff
@@ -293,20 +293,13 @@ public static string GetMethodName(Operation operation)
...
-                return operation.OperationId;
-            }
-
-            var parts = operation.OperationId.Split('_');
-            return parts[1];
+            return GetMethodNameFromOperationId(operation.OperationId);
         }
 
+        public static string GetMethodNameFromOperationId(string operationId) => 
+            (operationId?.IndexOf('_') != -1) ? operationId.Split('_')[1]: operationId;
```
olydis: touching that code anyways, I'd just define the method as `operationId.Split('_').Last()`

src/core/AutoRest.Extensions/ParameterGroupExtensionHelper.cs
```diff
@@ -111,8 +111,7 @@ private static string GenerateParameterGroupModelText(IEnumerable<Method> method
             if (methodList.Count == 1)
             {
                 Method method = methodList.Single();
-                return string.Format(CultureInfo.InvariantCulture, "Additional parameters for the {0} operation.",
-                    createOperationDisplayString(method.MethodGroup.Name.ToPascalCase(), method.Name));
+                return string.Format(CultureInfo.InvariantCulture, "Additional parameters for {0} operation.", method.Name);
```
olydis: For consistency, shouldn't this use `SwaggerModeler.GetMethodNameFromOperationId(method.Name)` as above?
I'd just reuse `"Additional parameters for " + SwaggerModeler.GetMethodNameFromOperationId(method.Name) + " operation."` here, `string.Format` with `CultureInfo` is complete nonsense here anyways.

src/modeler/AutoRest.Swagger/SwaggerModeler.cs
dsgouda: Do we always assume that all operationIds MUST have a single underscore? I know we have a validation rule to enforce that but still seems like a sensitive change.

olydis: Note that for multiple underscores, neither picking the 2nd nor the last one makes sense. So yes, we do assume it and making this simplification will at least be crystal clear with that (the other one only looks smarter when it really isn't 😆 )

src/core/AutoRest.Extensions/ParameterGroupExtensionHelper.cs
```diff
@@ -110,9 +110,7 @@ private static string GenerateParameterGroupModelText(IEnumerable<Method> method
             List<Method> methodList = methodsWhichUseGroup.ToList();
             if (methodList.Count == 1)
             {
-                Method method = methodList.Single();
-                return string.Format(CultureInfo.InvariantCulture, "Additional parameters for the {0} operation.",
-                    createOperationDisplayString(method.MethodGroup.Name.ToPascalCase(), method.Name));
+                return "Additional parameters for "+ methodList.Single().Name + " operation.";
```
olydis: still wondering, why is this inconsistent with the `Documentation` string you assemble above?

dsgouda: If by above you mean `AzureExtensions.cs`, for some reason looks like the `method` variable there has `name` in the form of operationId itself while in `ParameterGroupExtensionHelper.cs`, the `name` property denotes the `{1}` in `{0}_{1}` in an operationId.
We could probably investigate where these diverge but looks like overkill for such a small change.

olydis: OOOOHHH okay yeah typical AutoRest, no worries then, sorry about the nagging 👍

Samples/3f-example-generation/shell/stderr.txt
```diff
@@ -1,6 +1,26 @@
+  at ...
+  at ...
```
olydis: @dsgouda uh I think there were some errors during generation, always look at the diff ;)

olydis: only regen the samples to fix this one...

dsgouda: On it.

PR #2539, opened 2017-08-17
Tests hang on travis (WIP)

mcardosos:
Maybe http tests take too much time? That does not mean they hang/freeze


olydis:
moved to https://github.com/Azure/autorest.go/pull/3

PR #2547, opened 2017-08-21
Extension and modeler fixes.

jhendrixMSFT:
Generate parameter group postfix names according to the spec.
Preserve extensions when building header types.

### DIFF
--- a/src/core/AutoRest.Extensions/ParameterGroupExtensionHelper.cs
+++ b/src/core/AutoRest.Extensions/ParameterGroupExtensionHelper.cs
@@ -69,7 +69,7 @@ namespace AutoRest.Extensions
                     string parameterGroupName;
                     if (specifiedGroupName == null)
                     {
-                        string postfix = extensionObject.Value<string>("postfix").Capitalize() ?? "Parameters";
+                        string postfix = extensionObject.Value<string>("postfix")?.Capitalize() ?? "Parameters";
                         parameterGroupName = $"{methodGroupName}{methodName.Capitalize()}{postfix}";
                     }
                     else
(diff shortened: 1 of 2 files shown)


src/core/AutoRest.Extensions/ParameterGroupExtensionHelper.cs
```diff
@@ -69,8 +69,8 @@ private static string GetParameterGroupName(string methodGroupName, string metho
                     string parameterGroupName;
                     if (specifiedGroupName == null)
                     {
-                        string postfix = extensionObject.Value<string>("postfix") ?? "Parameters";
-                        parameterGroupName = methodGroupName + "-" + methodName + "-" + postfix;
+                        string postfix = extensionObject.Value<string>("postfix").Capitalize() ?? "Parameters";
+                        parameterGroupName = $"{methodGroupName}{methodName.Capitalize()}{postfix}";
```
olydis: This can obviously introduce breaking changes, but it seems to me that all [occurrences](https://github.com/Azure/azure-rest-api-specs/search?utf8=%E2%9C%93&q=x-ms-parameter-grouping&type=) in the specs repo *have* a `name`, so virtually nobody should be affected - does that make sense?
Also, I see that this change really just makes AutoRest comply with what we [claim](https://github.com/Azure/azure-rest-api-specs/blob/current/documentation/swagger-extensions.md#x-ms-parameter-grouping) we do...

Thinking about it, it would be really nice to have a [Sample](https://github.com/Azure/autorest/tree/master/Samples) for parameter grouping that covers everything one can do.

src/core/AutoRest.Extensions/ParameterGroupExtensionHelper.cs
```diff
@@ -69,8 +69,8 @@ private static string GetParameterGroupName(string methodGroupName, string metho
                     string parameterGroupName;
                     if (specifiedGroupName == null)
                     {
-                        string postfix = extensionObject.Value<string>("postfix") ?? "Parameters";
-                        parameterGroupName = methodGroupName + "-" + methodName + "-" + postfix;
+                        string postfix = extensionObject.Value<string>("postfix").Capitalize() ?? "Parameters";
```
olydis: `extensionObject.Value<string>("postfix")` may be null (that seems to be what makes CI fail)

PR #2552, opened 2017-08-24
split projects

olydis:


olydis:
ignoring Linux issues for now, will fix before final release, but this way we can start testing non-regression of AutoRest from 1.2.2 to now. Whatever causes the EPIPE issue won't affect code gen regression...

PR #2555, opened 2017-08-26
linux ready and updated classic validator

olydis:
Linux problems encountered:
- different layout in `.autorest` folder, specifically `.bin` folder had different nesting (looks like it's moved upwards when deeper folders can be symlinked which obviously prevents `.bin` to be inside of those)
- C# simplifier gets a `TaskCancelledException` with no apparent cause (@fearthecowboy investigation ongoing)

@fearthecowboy This PR also contains docs changes to accommodate for the changes, including docs on `use` (and the fact that no clone of this repo here is required), pointers to the new repos, etc., stuff we can point at in the announcement. BTW, everything is published and seems to work fantastic (well except the simplifier on Linux), it's accessible through `autorest@prerelease`. Note that I have published all extensions and the `autorest-core` package without a tag since it was the first publish anyways, so no point/effect.


olydis:
@fearthecowboy merged for now, so I can merge 2.0 to master...

PR #2562, opened 2017-08-30
Updating reference to the spec format to OAS from OAI

earth2marsh:
Also corrects some instances of `Autorest` to `AutoRest`.


olydis:
@earth2marsh Thanks a lot, this is very nice 🙂

PR #2583, opened 2017-09-12
Static link

fearthecowboy:
Changes bootstrapper and core to use static linked libraries.


src/autorest-core/app.ts
```diff
@@ -25,6 +25,10 @@ import { DataStore } from "./lib/data-store/data-store";
 import { EnhancedFileSystem, RealFileSystem } from './lib/file-system';
 import { Exception, OperationCanceledException } from "./lib/exception";
 
+// enable static modules for autorest-core
+if( (<any>global).StaticVolumeSet ) {
```
fearthecowboy: adds it's own static file system if the StaticVolumeSet is in the global namespace.

olydis:
@fearthecowboy This is only blocked by an empty line diff(!?) in a generated file... please merge if you're ready, so the other PRs get the benefit of having their changes to core actually tested 😉

fearthecowboy:
I've got a bit more to come with static linking, but this should unblock...

PR #2590, opened 2017-09-15
Demoting message importance for missing validator paths

dsgouda:


olydis:
Why is this PR done? We discussed about this priority multiple times already: This is a warning because it indicates **problems with linter rules**. I guarantee that these will never be fixed if we turn this into a debug message. The right thing to do is fix the problem, not mute it.

vishrutshah:
@olydis This is done to make sure we do not see duplicates in CI. 

Currently, we have two messages for one error but with different format in CI. Example: https://travis-ci.com/Azure/azure-rest-api-specs-pr/jobs/86104329#L693

One is 
```json5
{
  "type": "Warning",
  "code": "LocationMustHaveXmsMutability",
  "message": "Property 'location' must have '\"x-ms-mutability\":[\"read\", \"create\"]' extension defined. Resource Model: 'LocationServicesAccount'",
  "id": "S4002",
  "validationCategory": "SDKViolation",
  "providerNamespace": null,
  "resourceType": null,
  "sources": [
    "file:///home/travis/build/Azure/azure-rest-api-specs-pr/specification/locationservices/resource-manager/Microsoft.LocationServices/2017-01-01-preview/locationservices.json:469:2 ($.definitions)"
  ]
}
```

another is
```json5
{
  "Channel": "warning",
  "Text": "Could not find the exact path [\"definitions\",\"properties\",\"location\"] for {\"type\":\"Warning\",\"code\":\"LocationMustHaveXmsMutability\",\"message\":\"Property 'location' must have '\\\"x-ms-mutability\\\":[\\\"read\\\", \\\"create\\\"]' extension defined. Resource Model: 'LocationServicesAccount'\",\"id\":\"S4002\",\"validationCategory\":\"SDKViolation\",\"providerNamespace\":null,\"resourceType\":null}"
}
```

Let me know if that is expected from the linter :)

vishrutshah:
We are closing this PR as we discussed with @olydis that reporting is intentional for us to model our linter rules to correctly detect the paths all the time. There is no instance when we should not be able to discover the path. If we see this then that implies that rule is broken in terms of path reporting. 

Just putting notes here for I don't forget.

dsgouda:
Talked to @fearthecowboy and looks like we need to do this anyway to reduce some of the noise

olydis:
# publish job

PR #2594, opened 2017-09-20
Add default property to x-ms-enum in schema and update documentation

amarzavery:
Corresponding PR in the modeler https://github.com/Azure/autorest.modeler/pull/9
This PR updates the schema and documentation for x-ms-enum


docs/extensions/readme.md
```diff
@@ -97,6 +98,7 @@ accountType:
   x-ms-enum:
     name: AccountType
     modelAsString: false
+    default: Standard_LRS
```
fearthecowboy: I don't think we should have default on the `enum` definition itself, I think it belongs to the property.

We should have a call today and chat about it.

amarzavery:
closing this PR as we do not need default property in x-ms-enum extension

PR #2605, opened 2017-09-22
Enable Publish preview

fearthecowboy:


.gulp/common.iced
```diff
@@ -1,10 +1,103 @@
+# place an object into global namespace
```
fearthecowboy: I merged `gulp.iced` and `common.iced` into a single `common.iced`.

gulpfile.js
```diff
@@ -0,0 +1,4 @@
+// set the base folder of this project
```
fearthecowboy: makes it work nice with vscode. all logic is now in `.gulp/*.iced` files

olydis:
# publish job

PR #2628, opened 2017-09-29
more cowbell

olydis:
- small fix regarding source maps when composing swaggers that have path-level parameters
- yet better trampoline
- port of https://github.com/Azure/autorest/pull/2515:


olydis:
# publish job

PR #2632, opened 2017-10-03
Fixing bad markup in cli.md

csuich2:
Simple typo fix in cli.md. The markup was rendering incorrectly because of a missing escape character.


olydis:
# publish job

PR #2633, opened 2017-10-03
added example for method removal

olydis:


olydis:
> linux-x64 restart

olydis:
# publish job

PR #2639, opened 2017-10-05
Next gen stuff (WIP)

olydis:
- `require` config files
- `declare-directive`
- `gist://<ID>`


src/autorest-core/app.ts
```diff
@@ -257,43 +257,34 @@ async function currentMain(autorestArgs: string[]): Promise<number> {
...
-    }
-    for (const artifact of artifacts) {
-      await WriteString(artifact.uri, artifact.content);
-    }
+  // perform file system operations.
+  for (const folder of clearFolders) {
+    try { await ClearFolder(folder); } catch (e) { }
   }
-  catch (e) {
-    api.Message.Dispatch({
```
olydis: this catch made no sense. It swallowed `Exceptions` too which are handled correctly by the caller of this method... let's trust our caller.

src/autorest-core/app.ts
```diff
@@ -469,26 +460,22 @@ async function main() {
...
+      if (autorestArgs.indexOf("--debug") !== -1) {
         console.log(e);
+      } else {
+        console.log(e.message);
       }
       process.exit(e.exitCode);
     }
 
-    if (e instanceof Error) {
+    if (e !== false) {
```
olydis: if `Process` doesn't return `true`, we throw what was returned (line 277ish)... not so cool to print that in case its `false`

src/autorest-core/lib/autorest-core.ts
```diff
@@ -212,25 +212,22 @@ export class AutoRest extends EventEmitter {
         view.messageEmitter.removeAllListeners();
         return true;
       } catch (e) {
-        if (e instanceof Error) {
```
olydis: just swapped this test since `Error`s marshalled back from safeEval fail this test. `Exception`s on the other hand are not thrown within safeEval so we can rely on that

src/autorest-core/lib/configuration.ts
```diff
@@ -680,10 +741,23 @@ export class Configuration {
   public static async DetectConfigurationFile(fileSystem: IFileSystem, configFileOrFolderUri: string | null, messageEmitter?: MessageEmitter, walkUpFolders: boolean = false): Promise<string | null> {
     const originalConfigFileOrFolderUri = configFileOrFolderUri;
 
+    //
     if (!configFileOrFolderUri || configFileOrFolderUri.endsWith(".md")) {
       return configFileOrFolderUri;
     }
 
+    // try querying the Uri directly
+    if (ExistsUri(configFileOrFolderUri)) {
```
olydis: required for example in case of gists... those Uri's don't necessarily end on `.md` but on some hex code

olydis:
# publish job

PR #2641, opened 2017-10-05
Fix indents in x-ms-client-flatten example

RikkiGibson:
I think there were some tabs in the examples which rendered differently on GitHub than in the author's editor.


olydis:
# publish job

PR #2668, opened 2017-10-23
swallow leading $

olydis:
...because a certain `oav` emits paths like that which led to all paths getting reported as `$`...


olydis:
~~~ Haskell
~~~

PR #2676, opened 2017-10-29
Enable VSC Language-service

fearthecowboy:
- Switch to static-link tool
- add vscode language service
- fix config detection ( was taking files that didn't have the magic string)
- included vscode interfaces in bootstrapper


olydis:
~~~ Haskell
~~~

.gulp/typescript.iced
```diff
@@ -11,6 +11,9 @@ copyDtsFiles = (done) =>
     .on 'end', () => 
     source ["#{basefolder}/src/autorest-core/dist/**/*.d.ts","!#{basefolder}/src/autorest-core/dist/test/**" ]
       .pipe destination "#{basefolder}/src/autorest/dist/lib/core"
+    .on 'end', () =>
+     source ["#{basefolder}/src/autorest/vscode/**/*.d.ts" ]
```
fearthecowboy: I copied the vscode interfaces out of the four seperate packages that they come in. I only need the d.ts.files and didn't need the actual code.

src/autorest-core/.scripts/post-install.js
```diff
@@ -0,0 +1,3 @@
+require("child_process");
```
fearthecowboy: will delete this.

src/autorest-core/lib/configuration.ts
```diff
@@ -393,17 +394,18 @@ export class ConfigurationView {
...
                 blameTree = this.DataStore.Blame(s.document, s.Position);
               } catch (e) {
                 const path = s.Position.path as string[];
                 if (path) {
-                  this.Message({
-                    Channel: Channel.Verbose,
-                    Text: `Could not find the exact path ${JSON.stringify(path)} for ${JSON.stringify(m.Text)}`
-                  });
                   if (path.length === 0) {
+                    this.Message({
```
olydis: why? this message is there solely to raise awareness about a buggy extension, i.e. one that reports the wrong path - reporting only paths that are SO broken that you walk up to `$` seems somewhat arbitrary, muting bugs and making them go unnoticed (this was once even a warning, which was our only leverage for the tooling people to fix their rules) :P

fearthecowboy: A couple of points:

- I watched a lot of these messages go by, and a lot of them can be caused because there is a drop in fidelity of what the path can be because of the modeler. The leaf (or leaf+1) ends up being correct, but the rest is fine. I'm not so much worried about these; the correction is fixing it up fine.

- for messages that have no useful path, then I'm willing to yell, but I don't want to see a whole bunch of messages for the same one (which happens on every iteration of the loop)

- sadly, It's clear that they don't care about these anyway. The best we can do is attach the message to the less accurate ancestor node and move along.

fearthecowboy:
I seem to have a unresolved promise somewhere, because the tests are hanging; investigating...

src/autorest-core/lib/configuration.ts
olydis: > caused because there is a drop in fidelity of what the path can be because of the modeler

how is the modeler involved? 😲 

> sadly, It's clear that they don't care about these anyway. The best we can do is attach the message to the less accurate ancestor node and move along.

 well that's how we forced them to fix the rules last time and it worked great :P but okay, I guess this is causing noise since VScode subscribes to verbose messages too? Just wondered, since *verbose* messages seem to be precisely for this kind of thing, but now we're trying to not be verbose on verbose? :D

src/autorest-core/language-service/language-service.ts
```diff
@@ -0,0 +1,653 @@
...
+} from 'vscode-languageserver';
+
+
+
+//TODO: adding URL here temporarily, this should be coming either in the message coming from autorest or the plugin
+const azureValidatorRulesDocUrl = "https://github.com/Azure/azure-rest-api-specs/blob/current/documentation/openapi-authoring-automated-guidelines.md";
+
+const md5 = (content: any) => content ? createHash('md5').update(JSON.stringify(content)).digest("hex") : null;
+
+class Result {
```
veronicagg: Could you add comments as to what this represents and what each property means?

src/autorest-core/language-service/language-service.ts
```diff
@@ -0,0 +1,653 @@
...
+
+      const content = await readFile(FileUriToPath(fileUri));
+
+      return content;
+    } catch {
+    }
+    return "";
+  }
+
+  private async process(configurationUrl: string) {
```
veronicagg: could you add comments on what this does? It appears to be the main entry point to get the output of AutoRest, basically "executeAutorest"

src/autorest-core/language-service/language-service.ts
```diff
@@ -0,0 +1,653 @@
...
+      });
+    }
+
+    return configFile;
+  }
+
+  private async onDocumentChanged(document: TextDocument) {
+    this.debug(`onDocumentChanged: ${document.uri}`);
+
+    if (AutoRest.IsSwaggerExtension(document.languageId) && await AutoRest.IsSwaggerFile(document.getText())) {
```
veronicagg: in both of these cases from the vs code extension point of view, we'd like to have azure-validator: true set by default and openapi-type: arm. If they are in the values are set in the configuration already, we should take those, but if not we want to set them. Could we set configuration options here? Can we have these values also being picked up for the case where we're creating a dummy readme.md file?

src/autorest-core/language-service/language-service.ts
```diff
@@ -0,0 +1,653 @@
...
+    return content;
+  }
+
+  constructor(private connection: IConnection) {
+    super();
+
+    //  track opened, changed, saved, and closed files
+    this.onDidOpen((p) => this.onDocumentChanged(p.document));
+    this.onDidChangeContent((p) => this.onDocumentChanged(p.document));
+    this.onDidClose((p) => this.getDiagnosticCollection(p.document.uri).clear(true));
```
veronicagg: are we just clearing results here? 
In the previous version of the extension, we had it behaved:
 - if config file is left opened, we do not want to clear diagnostics for referenced files
  - if config file is closed and no other referenced files are opened, then we should clear diagnostics.
why shouldn't we maintain the same behavior?

src/autorest-core/language-service/language-service.ts
```diff
@@ -0,0 +1,653 @@
...
+      const doc = this.get(each.uri);
+      if (doc) {
+        this.onDocumentChanged(doc);
+        return;
+      }
+
+      let documentUri = each.uri;
+      const txt = await this.ReadFile(each.uri);
+      if (documentUri.startsWith("file://")) {
+        // fake out a document for us to play with
```
veronicagg: what case does this correspond to? just opening a file and start typing something?

PR #2684, opened 2017-10-31
Fix checkout/build 

fearthecowboy:
The projects need to create the static filesystems when npm install is called during development.


olydis:
~~~ Haskell
~~~

PR #2695, opened 2017-11-07
Adding extensions to handle extensible enums

dsgouda:


olydis:
~~~ Haskell
~~~

schema/swagger-extensions.json
```diff
@@ -1509,6 +1513,12 @@
               },
               "name": {
                 "type": "string"
+              },
+              "allowedValues": {
+                "type": "array",
+                "items": {
+                  "type": "string"
```
olydis: I'd not restrict the type, as we already don't restrict the type of enum values in this schema

dsgouda: AFAIK we want to restrict the types to Primary types. The modeler reads these values into a string array anyway. When generating the code we decide how the code is generated based on the type of the enum itself. (i.e. string or non-string)

olydis: Correct, but this restriction is not reflected otherwise in this schema. This is just about consistency. 😉 Also, the modeler is in motion these days.

dsgouda: relaxing the type here means authors can technically pass in values of any type and if the modeler cannot handle that its counter productive no?

olydis: yes - and that's what they already *can* do for `enum: [...]` or `x-ms-enum: values: ...`. Consistency. As a matter of fact we *have* Swaggers in the repo that contain non-string enums, they seem to be gracefully ignored by the modeler. And as I said, the modeler is in motion, supporting non-string enums is high on the list so... why make this schema more restrictive and inconsistent?

PR #2702, opened 2017-11-09
preserve object graph

olydis:
### serialization

We would use JSON as long as there are no circular data structures - but that's too agressive and can still impose data loss compared to YAML! If the object graph contains multiple instances of the same object, each of those instances would be serialized out - so the information about object identity is lost.

### deserialization

When hitting a YAML anchor, we emitted a clone of the referenced object instead of literally returning the same reference.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -3,6 +3,12 @@
   "configurations": [
     {
       "type": "node",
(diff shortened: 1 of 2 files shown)


olydis:
~~~ Haskell
~~~

PR #2722, opened 2017-11-20
nested style fix

olydis:
nested styles were rendered incorrectly, e.g.
```
`asd "qwe" rty`
```
would be rendered as if it was
```
`asd "qwe"` rty
```
for the simple reason that the ANSI codes injected by `color` are not stack-like, i.e. the "end of `"`-style" ANSI code will actually reset the color to whatever it was before the backtick style started.

I've changed the function to first inject PUSH/POP style instructions into the text and then in the very end compile that into a *flat* sequence of styled text segments.

### DIFF
--- a/src/autorest/app.ts
+++ b/src/autorest/app.ts
@@ -16,22 +16,58 @@ import { networkEnabled, rootFolder, extensionManager, availableVersions, corePa
 import { gt } from "semver";
 import chalk from "chalk"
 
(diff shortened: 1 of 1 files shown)


azuresdkci1x:
~~~ Haskell
~~~

olydis:
gracias :D yeah saw glitches in the command line help (that I'm simplifying right now so the other generators can provide their part easily)

PR #2808, opened 2018-01-19
Allows per-language readme.XXX.md files

fearthecowboy:
Adds:
  `try-include` support (includes a config file, but no message on failure)
  valueless parameters (ie,  --foo ) will do a `try-include` for `readme.foo.md` 
  known languages have `readme.<lang>.md` in a `try-include` already


olydis:
~~~ Haskell
~~~

Samples/3g-require-config/shell/stdout.txt
```diff
@@ -1,2 +1 @@
 https://aka.ms/autorest
-Including configuration file '/Samples/1a-code-generation-minimal/readme.md'
```
fearthecowboy: Is now a `--verbose` message

Issue #2819, opened 2018-01-24
Autorest does not allow swagger circular dependencies


mcardosos:
Noticed it here...
https://github.com/Azure/azure-rest-api-specs/pull/2269
Linter in CI is not running, because file a references stuff in file b and viceversa.

olydis:
Argh, in the end this issue was actually not because of multiple files referencing each other, it just looped within one file forever... will release fix soon

PR #2820, opened 2018-01-24
fixed potential SO in polymorphic resolution

olydis:
should fix https://github.com/Azure/autorest/issues/2819


olydis:
~~~ Haskell
~~~

PR #2858, opened 2018-02-23
Moving base-image of Dockerfile to use ubuntu:xenial

marstr:
This also required picking a compatible version of `libicu`
https://packages.ubuntu.com/trusty/libicu52

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
-FROM ubuntu:trusty
+FROM ubuntu:xenial 
 
 LABEL maintainer="fearthecowboy" 
 
 # Required for install
-RUN apt-get update && apt-get install -y curl libunwind8 libicu52
+RUN apt-get update && apt-get install -y curl libunwind8
 
 # NodeJS
 RUN curl -sL https://deb.nodesource.com/setup_7.x | bash - && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
-FROM ubuntu:17.04
+FROM ubuntu:trusty
 
 LABEL maintainer="fearthecowboy" 
 
 # Required for install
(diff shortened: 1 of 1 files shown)


olydis:
~~~ Haskell
~~~

Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM ubuntu:17.04
+FROM ubuntu:trusty
```
lmazuel: Trusty is super old! 2014!

I use xenial myself, and then I don't need libicu hack.

marstr: What's wrong with old? ;)

Truthfully, I picked `trusty` because it is what Travis uses, and therefore implies a little stability in my mind.

marstr: As for `libicu`, I don't know enough about why it's required or how it should be installed to do anything but pick a compatible version!

marstr: Fixed both in 4babfba

marstr:
Hey @olydis, looks to me like CI is failing due to transient errors. How do I restart the build?

lmazuel:
@marstr Do we want two docker containers? one for `autorest@preview` and one for `autorest@latest`?

olydis:
@marstr wtf indeed... typing `> win32-x64 restart` and the same for linux does the trick, but I'll quickly check if something locks those folders

olydis:
> linux-x64 restart

olydis:
> win32-x64 restart

marstr:
> win32-x64 restart

marstr:
> win32-x64 restart

marstr:
Howdy @olydis / @fearthecowboy, mind taking a look at this for me today?

fearthecowboy:
Looks good to me.

marstr:
Thank you!

PR #3095, opened 2019-02-09
Powershell config: folder update

MiYanni:
This is for: https://github.com/Azure/autorest.powershell/pull/48

- Adds api-folder and runtime-folder for use with the cshaper generator when ran when the powershell generator is loaded
- Only uses the autorest.powershell online version when the local version is not present

### DIFF
--- a/src/autorest-core/resources/plugin-powershell.md
+++ b/src/autorest-core/resources/plugin-powershell.md
@@ -2,17 +2,24 @@
 
 The beta version of the PowerShell Generator.
 
-Note: if the --powershell is mentioned, but they are using autorest.powershell, don't try to load the autorest.powershell from npm.
+``` yaml $(powershell)
 
-``` yaml $(powershell) && ( "$(requesting-extensions)".indexOf('autorest.powershell') === -1 ) 
 # requires multi-api merger
 enable-multi-api: true
+api-folder: generated/api
+runtime-folder: generated/runtime
+
+# probe from readme.powershell.md file 
+try-require: ./readme.powershell.md
+
+```
+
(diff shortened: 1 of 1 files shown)


src/autorest-core/resources/plugin-powershell.md
```diff
@@ -2,17 +2,24 @@
 
 The beta version of the PowerShell Generator.
 
-Note: if the --powershell is mentioned, but they are using autorest.powershell, don't try to load the autorest.powershell from npm.
+``` yaml $(powershell)
 
-``` yaml $(powershell) && ( "$(requesting-extensions)".indexOf('autorest.powershell') === -1 ) 
 # requires multi-api merger
 enable-multi-api: true
+api-folder: generated/api
```
fearthecowboy: I was trying to see why these are here; but I think I get it. 

They really should stay in the powerShell generator, but the trouble is, the `readme.md` file at the root of the repo is what's getting used when it's run from the cloned repo, and the one that's in the extensions/powershell is the one that gets used when it's installed via npm.

I'll go fix that so there is a single canonical source for that, and then this stuff can live there.

MiYanni: The issue was that, you need to declare the `api-folder` and `runtime-folder`, so that the CSharp generator, when ran in conjunction with the PowerShell generator, would output to the correct folders. The CSharp generator has it's own defaults (api-folder at 'current folder' and runtime-folder in `runtime`), but when the PowerShell generator is ran, we want the CSharp generator to use api-folder as `generated/api` and runtime-folder as `generated/runtime`. So, this was the best solution.

PR #3141, opened 2019-02-22
Use the `img` property. No `url` property

dfinke:


fearthecowboy:
Thanks...

I thought I got to that before I pushed it up. *sigh*

PR #3146, opened 2019-02-27
enhanced name inferring for array items and array item properties.

NelsonDaniel:


src/autorest-core/lib/pipeline/plugins/tree-shaker.ts
```diff
@@ -1,7 +1,8 @@
-import { AnyObject, DataHandle, DataSink, DataSource, Node, parseJsonPointer, Transformer, QuickDataSource } from '@microsoft.azure/datastore';
+import { AnyObject, DataHandle, DataSink, DataSource, Node, parseJsonPointer, Transformer, QuickDataSource, JsonPath } from '@microsoft.azure/datastore';
 import { ConfigurationView } from '../../configuration';
 import { PipelinePlugin } from '../common';
 import { clone } from '@microsoft.azure/linq';
+import { retries } from 'mocha-typescript';
```
fearthecowboy: ? What's `retries` in there for? that won't work at runtime...

NelsonDaniel: it probably got imported on the fly. Didn't realize that. I
 am gonna remove it.

Issue #3150, opened 2019-02-28
Update profile-filer plugin.


NelsonDaniel:
- [x] Add profile to api-version mapping inside operation x-ms-metadata.
- [x] Drop api-versions not included in the profile from metadata.

PR #3153, opened 2019-03-01
Update for profile support.

NelsonDaniel:
- update to profile-filter (profiles mapping to api version, unused apiversions drop), and tests.
- update to the way that profiles are selected.


NelsonDaniel:
Fixes: https://github.com/Azure/autorest/issues/3150

PR #3287, opened 2019-07-24
update default version of autorest.typescript to 4.1.1

amarzavery:


src/autorest-core/resources/default-configuration.md
```diff
@@ -123,7 +123,7 @@ try-require: ./readme.typescript.md
 
 ``` yaml $(typescript)
 use-extension:
-  "@microsoft.azure/autorest.typescript": "~2.0.12"
+  "@microsoft.azure/autorest.typescript": "^4.1.1"
```
fearthecowboy: You sure you want `^` ? 

> The tilde ~ matches the most recent patch version (the third number) for the specified minor version (the second number).
> ~1.2.3 will match all 1.2.x versions but will hold off on 1.3.0.
>
> The caret ^ is more relaxed. It matches the most recent minor version (the second number) for the specified major version (the first number).
> ^1.2.3 will match any 1.x.x release including 1.3.0, but will hold off on 2.0.0.

Usually we use `~` so that new patches rolled out are automatically the 'latest', but new minor versions are not.

amarzavery:
@fearthecowboy - The min version has been updated with `~` sign. Please merge this.

amarzavery:
@daviwil @ramya-rao-a - FYI, once you publish a new minor or major version of the generator to the latest tag on npm then you need to send a PR like this to update the min version of the TS generator. This will indicate autorest to use the latest version if one is not provided by the `--use` switch.

daviwil:
Thanks Amar!

PR #3290, opened 2019-07-26
Fix order of evaluation of overrides

jhendrixMSFT:
Ensure that settings passed on the command line take precedence over
settings in the config files.


fearthecowboy:
While reading your PR; I looked a bit closer into what's going on. 

I think that there is some type coercion going on where false values aren't actually getting propagated, because the code is thinking the value doesn't exist.

I'm going to test some things today and see if this is true.

jhendrixMSFT:
User error, closing.

PR #3306, opened 2019-09-10
Update to PowerShell customizations documentation

cormacpayne:
Update to the `customization.md` file to expand on information about PowerShell customizations.

_Note_: I'd recommend taking a look at the [full file](https://github.com/Azure/autorest/blob/customizations-doc/docs/powershell/customization.md) rather than the diff since the initial document is being merged into this one.

### DIFF
--- a/docs/powershell/customization.md
+++ b/docs/powershell/customization.md
@@ -1,85 +1,427 @@
-# Customization
+## PowerShell Customizations
 
-## Examples
+### What are customizations?
 
-### Replacing a parameter with multiple ones
+Customizations are a layer on top of the generated code that provide additional functionality that isn't directly exposed through the REST API specification used to generate the cmdlets. Customizations can be found in the `custom` folder at the root of every module folder, and are written in C# or PowerShell. Customizations can be brand new cmdlets, additional variants on top of existing cmdlets, and code that chains together multiple cmdlet calls within the module regardless of if the cmdlet is public or hidden.
 
-The parameter `-Filter` is used to filter the results that a GET operation will return. 
-The parameter is defined as an String and follows a particular format of 
+### How are customizations used?
 
-```
-eventTimestamp <op> <value> [and <attribute> eq <value> [and <attribute> eq <value>] ... ]
-```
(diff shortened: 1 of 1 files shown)


cormacpayne:
@fearthecowboy @markcowl @niander @MiYanni would you mind taking a look at this document when you get the chance? Let me know if there's anything I missed.

CC: @erich-wang @isra-fel @dolauli

niander:
@cormacpayne I think we should make clear that the only way to "remove a parameter" is to hide the generated cmdlet and introduce a complete new cmdlet customization that will invoke the hidden one. In the `-Filter` example wouldn't it make more sense to remove `-Filter` by hiding the original one and creating a new cmdlet that invoke the hidden one? Also, I believe that there is no example calling an internal cmdlet and having one would be nice

cormacpayne:
@niander added an example that removes a parameter and calls the internal module. I didn't want to touch the existing `-Filter` example since, in my opinion, it still makes sense to expose the `-Filter` parameter to the user as they can create a more complex query string on their own (the example I gave only used an `AND` as a separator where the user could also use `OR`) and I don't want to take away that scenario if that's the route they choose to go.

docs/powershell/customization.md
```diff
@@ -1,85 +1,454 @@
...
+{
+    # The variable $Foo has a value that was provided by the user
+
+    # The -Foo parameter is not found in the next cmdlet, so we should remove it once we're done using it
+    $null = $PSBoundParameters.Remove("Foo")
+}
 ```
-eventTimestamp <op> <value> [and <attribute> eq <value> [and <attribute> eq <value>] ... ]
+
+_Note_: for PowerShell, the removal operation is assigned to a `$null` variable to prevent the returned `bool` from being written out from the cmdlet; this should be done for the removal operation on a dictionary.
```
markcowl:
```suggestion
_Note_: for PowerShell, the removal operation is assigned to a `$null` variable to prevent the returned `bool` from being written to the Output stream from the cmdlet; this should be done for the removal operation on a dictionary, or any other command or call that returns a value.
```

Issue #1259, opened 2016-07-13
Common Headers object to be used across multiple operations


arulnithi:
Currently I have multiple operations that return somewhat similar headers in the response. Say for example a register operation that can be used to POST different assets. However since the type of asset is specified in the path, these are all modeled as different Operation objects in Swagger. 

Hence, when I receive the headers for these operations, they are all different classes modeled by AutoRest, while containing the same basic information. 

I haven't found any way to model it in Swagger for AutoRest to package the response headers into a generic Headers object

Eg.

```

"paths" : {
    "genereric_enpoint/asset_type/AssetA" : {
        "post" : {
            "responses" : {
                "200" : {
                    "headers"  :  {
                        "X"  : {  "type" : "string" },
                        "Y"  : {  "type" : "string" }
                    }
                }
            }
        }
    },  
    "genereric_enpoint/asset_type/AssetB" : {
        "post" : {
            "responses" : {
                "200" : {
                    "headers"  :  {
                        "X"  : {  "type" : "string" },
                        "Y"  : {  "type" : "string" }
                    }
                }
            }
        }
    },
}
```

Currently these two operations return different header objects, and it really adds up over different operations and response types, ( I currently have over 60+ specific header objects when I ideally only need about 4)

I am not sure if there is currently any features in AutoRest to specify a generic Header object, but I have not found any in the documentation.

fearthecowboy:
This is kindof a limitation in swagger. The `headers`  is an associative array that is used to generate the headers class that the response uses.

It is somewhat *possible* to try to compare all the generated `Headers` classes and merge them if they're identical, but I don't see it happening before 1.0 (this fall).  This may be easier to do when we make the pipeline a bit more open.

olydis:
we had this coming up during the hackathon, it was *really* painful there: lots of identical classes, that furthermore had to be processed further (hand-written convenience layer), so the code duplication in the auto generated code caused further code duplication in hand-written code.

olydis:
if we wanna fix this, I'd propose allowing a `$ref` to the `definitions` section in `headers`. AFAIK, this is not allowed as per the OpenAPI 2.0 spec so we don't collide with existing interpretations. It would also automatically allow us to customize the horrible class name chosen for header classes.

olydis:
[...]

PR #3322, opened 2019-10-21
Document x-ms-headers in extensions/readme.md

kevinoid:
Add documentation for the x-ms-headers extension for shared response headers (added in #2343 to address #1259) to `docs/extensions/readme.md` to make it easier for users to discover.

Example copied from `Samples/test/shared-response-header-types/readme.md`

Thanks for considering,
Kevin

### DIFF
--- a/docs/extensions/readme.md
+++ b/docs/extensions/readme.md
@@ -18,6 +18,7 @@ The following documents describes AutoRest specific vendor extensions for [OpenA
 * [x-ms-mutability](#x-ms-mutability) - provides insight to Autorest on how to generate code. It doesn't alter the modeling of what is actually sent on the wire.
 * [x-ms-examples](#x-ms-examples) - describes the format for specifying examples for request and response of an operation in an OpenAPI definition.
 * [x-ms-error-response](#x-ms-error-response) - indicates whether the response status code should be treated as an error response
+* [x-ms-headers](#x-ms-headers) - define shared response header types across multiple operations
 
 ### Microsoft Azure Extensions (available in most generators only when using `--azure-arm`)
 * [x-ms-odata](#x-ms-odata) - indicates the operation includes one or more [OData](http://www.odata.org/) query parameters.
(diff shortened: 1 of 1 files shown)


kevinoid:
This PR might be a bit over-eager.  I just tried using `x-ms-headers` and have been unable to get past:

    FATAL: System.ArgumentException: Reference path '#/definitions/MyHeaders' does not exist in the definition section of the Swagger document.

I couldn't find any other references to `x-ms-headers` or OpenAPI docs which use it.  Should it work, or is it just cruft?

fearthecowboy:
That's a fine question...

Oddly enough, I was just talking to someone else about how to support this, and I totally forgot that this was in here. 

I think this was to support a very specific scenario (that never actually got used). I'm not entirely sure that it actually works as intended or not. I think it set something up that wasn't actually consumed, but I gotta find out what it should be doing.

I don't think the current crop of generators support this at this point. The the new generators that are being written could if I added it to the new modeler.

Lemme get back to you.

kevinoid:
Interesting.  Thanks for the info @fearthecowboy.  No hurry on my end.  It would be useful to reduce the number of classes generated, but is not causing any significant issues for me.

fearthecowboy:
(closing since v3 is imminent anyway)

kevinoid:
That's great!

Forgive me for being dense, but how would shared response headers be defined in v3?  I can see how `/components/headers` can be used to define individual headers, but not how a reusable type for multiple headers would be defined, as `x-ms-headers` appeared to do.

fearthecowboy:
At this point, it's not supported -- I'm not sure that it ever was really actually working. It may have been that the developer was just trying something out and it was left in there.

It would be handy tho' ... the blithering repetitiveness of header declarations is terrible. 

In OAI3, you can probably `$ref ` individual headers from `/components/headers` to reuse them, but there isn't a grouping mechanism at this point.

kevinoid:
That makes sense.  Thanks @fearthecowboy.

kevinoid:
For future reference, response header groups and parameter groups, possibly as part of a trait system, are being considered in OAI/OpenAPI-Specification#445 (e.g. strawman in https://github.com/OAI/OpenAPI-Specification/issues/445#issuecomment-451452411).  They might be able to replace `x-ms-headers` and `x-ms-parameter-grouping` if included in a future OpenAPI version.

PR #3350, opened 2019-12-03
quick fixes

fearthecowboy:
- add xms-secret to schema. 
- add parameters at pathitem level, 
- trim empty inputs from pipeline 
-

### DIFF
--- a/core/lib/configuration.ts
+++ b/core/lib/configuration.ts
@@ -22,6 +22,7 @@ import { AutoRestExtension } from './pipeline/plugin-endpoint';
 import { Suppressor } from './pipeline/suppression';
 import { MergeOverwriteOrAppend, resolveRValue } from './source-map/merging';
 import { Initializer, DeepPartial } from '@azure-tools/codegen';
+import { IdentifyDocument } from './autorest-core';
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const untildify: (path: string) => string = require('untildify');
@@ -763,8 +764,7 @@ export class Configuration {
           // @<org>/<pkg>@<version>
           // <path>
           // if the entry starts with an @ it's definitely a package reference
-
-          if (await isDirectory(useEntry)) {
+          if (useEntry.endsWith('.tgz') || await isDirectory(useEntry)) {
             const pkg = await extMgr.findPackage('plugin', useEntry);
             configs['use-extension'][pkg.name] = useEntry;
           } else {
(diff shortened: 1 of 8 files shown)


core/lib/openapi/conversion.ts
```diff
@@ -9,6 +9,7 @@ import { clone } from '@azure-tools/linq';
 
 export async function convertOAI2toOAI3(input: DataHandle, sink: DataSink): Promise<DataHandle> {
   const converter = new Oai2ToOai3(input.originalFullPath, await input.ReadObject());
+  console.log(input.originalFullPath);
```
daviwil: Is this just for debugging or actual output?

core/lib/pipeline/common.ts
```diff
@@ -16,8 +17,14 @@ export function createPerFilePlugin(processorBuilder: (config: ConfigurationView
     const result: Array<DataHandle> = [];
     for (const file of files) {
       const fileIn = await input.ReadStrict(file);
-      const fileOut = await processor(fileIn, sink);
-      result.push(fileOut);
+
+      // only keep/process files that actually have content in them (ie, no empty objects, no {directive:[]} files ).
+      const pi = await fileIn.ReadObject<any>();
```
daviwil: Does `pi` stand for `plugin` in this case?

schema/swagger-extensions.json
```diff
@@ -82,9 +82,6 @@
     },
     "externalDocs": {
       "$ref": "#/definitions/externalDocs"
-    },
-    "x-ms-secondary-file": {
```
daviwil: Wasn't being used anywhere?

core/lib/configuration.ts
```diff
@@ -763,8 +764,7 @@ export class Configuration {
           // @<org>/<pkg>@<version>
           // <path>
           // if the entry starts with an @ it's definitely a package reference
-
-          if (await isDirectory(useEntry)) {
+          if (useEntry.endsWith('.tgz') || await isDirectory(useEntry)) {
```
daviwil: Any value in making the `endsWith` check a function like `isPackage` since it occurs more than once?

core/lib/openapi/conversion.ts
fearthecowboy: oops. that shouldn't be there

schema/swagger-extensions.json
fearthecowboy: huh. Pulled the file from `master`, wasn't in `master` Weird.

core/lib/configuration.ts
fearthecowboy: Meh. I dunno--they're not actually checking for the same purpose -- this one passes the path thru to the package manager as the 'location/version', the other one is making sure we don't try to physically load it from the path.  The implementation could easily diverge in the future.

daviwil: Sounds good!

fearthecowboy:
> Looks like this PR supersedes PR #3349 too?
#3349 updates the swagger-extensions.json in `master`

this one adds the same changes to the v3 branch.

PR #3356, opened 2019-12-07
allow version: to be specified in configuration

fearthecowboy:
and load the autorest-core version that way.

### DIFF
--- a/autorest/app.ts
+++ b/autorest/app.ts
@@ -34,11 +34,14 @@ import { isFile, readdir, rmdir, isDirectory } from '@azure-tools/async-io';
 import { Exception, LazyPromise } from '@azure-tools/tasks';
 import { homedir } from 'os';
 import chalk from 'chalk';
-import { join } from 'path';
+import { join, dirname } from 'path';
 import { gt } from 'semver';
-import { availableVersions, newCorePackage, oldCorePackage, ensureAutorestHome, extensionManager, installedCores, networkEnabled, pkgVersion, resolvePathForLocalVersion, rootFolder, selectVersion, tryRequire } from './autorest-as-a-service';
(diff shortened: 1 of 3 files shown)


autorest/app.ts
```diff
@@ -181,6 +187,56 @@ async function clearTempData() {
...
+        }
+      },
+      cfgfile: ResolveUri(cwd, args.configFileOrFolder || '.'),
+      switches: args
+    };
+
+    // *sigh* ... there's a bug in most versions of autorest-core that to use the API you have to 
+    // have the current directory set to the package location. We'll fix this in the future versions.
+    process.chdir(dirname(autorestApi));
+    const configSpecifiedVersion = await vm.runInNewContext(`
```
fearthecowboy: I load whatever version of the autorest-core (into a sandbox) we can get our hands on, just to let it do the parsing of the configuration. 

If the configuration contains a `version` value, we substitute out that for the version that we assumed.

autorest/app.ts
```diff
@@ -282,7 +353,15 @@ async function main() {
     if (args.debug) {
       console.log(`Starting ${newCorePackage} from ${await selectedVersion.location}`);
     }
+
+    // if they never said the version on the command line, we should make a check for the config version.
+    if (!args.version) {
```
fearthecowboy: If they used a `--version` on the command line, we never want to override that from configuration.

autorest/app.ts
```diff
@@ -81,6 +84,9 @@ function parseArgs(autorestArgs: Array<string>): any {
         value = rawValue;
       }
       result[key] = value;
+    } else {
+      result._ = result._ || [];
```
daviwil: Where does this get used?  Doesn't seem to be used by the other changes in the PR

PR #3360, opened 2019-12-12
Rebuild 

fearthecowboy:
Remove cruft, 
- propagate x-ms-examples to doc
- add x-ms-original to track source of x-ms-parameterized host parameter
- pick up changes to deduplicator/oai2-to-oai3 converter,


daviwil:
LGTM

PR #3401, opened 2020-02-04
Enable PR validation with regression tests

daviwil:
This change uses [`autorest-compare`](https://github.com/Azure/autorest.compare) to implement regression testing for `@autorest/core` when PRs are sent.  For now it only compares the output of the v3 `@autorest/typescript` generator using both previous version of `@autorest/core` and the changes in the PR, but more languages will be added later as support comes to `autorest-compare`.

One thing to note is that I'm not currently using any specs from the `azure-rest-api-specs` repo since they generally tend to fail on the (temporary) lack of multi API support in `modelerfour`.  Once that comes online I'll add tests with a subset of the specs found there.

### DIFF
--- a/.scripts/verify-pull-request.yaml
+++ b/.scripts/verify-pull-request.yaml
@@ -3,8 +3,7 @@
 # Add steps that analyze code, save build artifacts, deploy, and more:
 # https://docs.microsoft.com/azure/devops/pipelines/languages/javascript
 
-trigger:
-- master
+trigger: none
 
 pool:
   vmImage: 'ubuntu-latest'
@@ -19,7 +18,9 @@ steps:
     npm install -g npm
     npx @microsoft/rush update
     npx @microsoft/rush rebuild
-    npx @microsoft/rush test
+
+    # Tests are disabled for now, failures to be investigated (Azure/autorest#3400)
+    #npx @microsoft/rush test
 
   displayName: 'Rush install, build and test'
 
--- a/.scripts/verify-pull-request.yaml
+++ b/.scripts/verify-pull-request.yaml
@@ -66,7 +66,7 @@ steps:
                      --spec-path:validation.json \
                      --spec-path:xml-service.json \
                      --spec-path:xms-error-responses.json \
-                     --spec-root-path:./modelerfour/node_modules/@microsoft.azure/autorest.testserver/__files/ \
+                     --spec-root-path:./node_modules/@microsoft.azure/autorest.testserver/__files/ \
                      --output-path:./core/test/generated/typescript/ \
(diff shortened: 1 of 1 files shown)


daviwil:
This PR is based on the `v3-regression-tests` branch that I pushed to this repo so that I could get the DevOps pipeline working.  I'll merge it back into `v3` once it's merged.

.scripts/verify-pull-request.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+- script : |
+    npm install @microsoft.azure/autorest.testserver --no-save
+    npm install -g @autorest/compare
+  displayName: Install autorest-compare
+
+- script : |
+    export AUTOREST_TYPESCRIPT=@autorest/typescript@0.1.0-dev.20200203.1
+    autorest-compare --typescript \
+                     --spec-path:azure-report.json \
```
daviwil: I'm designing a JSON config format which will eliminate the need for such a long command line invocation.

.scripts/verify-pull-request.yaml
```diff
@@ -0,0 +1,89 @@
...
+                     --spec-path:subscriptionId-apiVersion.json \
+                     --spec-path:url-multi-collectionFormat.json \
+                     --spec-path:url.json \
+                     --spec-path:validation.json \
+                     --spec-path:xml-service.json \
+                     --spec-path:xms-error-responses.json \
+                     --spec-root-path:./node_modules/@microsoft.azure/autorest.testserver/__files/ \
+                     --output-path:./core/test/generated/typescript/ \
+                     --compare-old --v3 --package-name:test-package --version:3.0.6200 --use:$AUTOREST_TYPESCRIPT \
+                     --compare-new --v3 --package-name:test-package --version:./core --use:$AUTOREST_TYPESCRIPT
```
daviwil: Just wanted to verify: is it valid to use `--version:./core`?  Seems like it works fine but wanted to check anyway.

fearthecowboy: Yep. works nice

.scripts/verify-pull-request.yaml
fearthecowboy: Maybe use YAML; all valid JSON is valid YAML, and sometimes it's nice to have comments.  yaml-js works nice for parsing.

daviwil: Thanks, will do!

fearthecowboy: Yeah, a config file would be really nice -- with this wrapped up in a pipeline, we can't really leverage this on the dev machine without cut/paste.

daviwil: Yep.  It would also be super difficult to regenerate baseline output consistently without a config file.  I'm going to implement my plan this morning and send out a PR before lunch.

PR #3418, opened 2020-02-20
Fix Batch mode, and other refinements

fearthecowboy:
allow defining a pass-thru extension to propogate extensions, 
make the output folder relative to the current directory when the base folder is not local

### DIFF
--- a/core/app.ts
+++ b/core/app.ts
@@ -44,6 +44,7 @@ import { AutoRestConfigurationImpl, MergeConfigurations } from './lib/configurat
 import { Exception, OperationCanceledException } from './lib/exception';
 import { Channel, Message } from './lib/message';
 import { ShallowCopy } from './lib/source-map/merging';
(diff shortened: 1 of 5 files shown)


core/app.ts
```diff
@@ -179,6 +180,11 @@ function parseArgs(autorestArgs: Array<string>): CommandLineArgs {
       rawValue = join(process.cwd(), rawValue);
     }
 
+    if (rawValue.startsWith('~/')) {
```
daviwil: Feels a little wrong to hardcode this but I guess it's hard to detect that the value is a path.  Perhaps use [`path.resolve`](https://github.com/Azure/autorest/pull/3418) when the value starts with `.` or `~`?

fearthecowboy: `path.resolve` doesn't actually do that.  (tried that first)  It doesn't know `~`

PR #3430, opened 2020-02-27
Add Python regression tests to PR validation pipeline

daviwil:
This change leverages the new Python language support in`@autorest/compare` to add regression tests using the new Python generator because it has more complete support for the new CodeModel.

### DIFF
--- a/.scripts/regression-compare.yaml
+++ b/.scripts/regression-compare.yaml
@@ -53,6 +53,19 @@ specs:
      - xml-service.json
      - xms-error-responses.json
 languages:
+  - language: python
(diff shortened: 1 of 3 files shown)


daviwil:
I should mention that I looked into running the Python and TypeScript regression tests in parallel and it looks like it'll require multiple agents to do that, which would also require packing up the build output of the repo here, uploading it as an artifact, and plopping it back down into the new agents that have a clean slate.  Seemed like it would be too much extra work for that benefit so I'm holding off for now until the runtime becomes unbearable.

PR #3431, opened 2020-02-28
updating az and test

zikalino:
- az should load readme.cli.md file as well
- adding config for autorest.test (test generation plugin)

### DIFF
--- a/core/resources/plugin-test.md
+++ b/core/resources/plugin-test.md
@@ -1,0 +1,9 @@
+# Default Configuration - test
+
+``` yaml $(test)
+use-extension:
+  "@autorest/test": "latest"
+try-require:
+  - ./readme.test.md
+  - ./readme.cli.md
+```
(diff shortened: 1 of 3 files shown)


core/resources/plugin-test.md
```diff
@@ -0,0 +1,9 @@
+# Default Configuration - test
+
+``` yaml $(test)
```
fearthecowboy: `test` is a reaaaaaly generic word. I'm pretty sure I have things that use `--test` which would trigger that. 

can we find a better switch for that?

zikalino: ok

daviwil: Circling back on this, what kind of tests is this generator generating?  Maybe I can help come up with a more specific name?

daviwil:
There hasn't been any activity on this PR for a while so I'll go ahead and close it.  @zikalino, if you want to continue the discussion I'll be happy to reopen.

PR #3443, opened 2020-03-07
Fix documented parent for x-ms-parameterized-host

kevinoid:
The `x-ms-parameterized-host` extension is only accepted on the Swagger Object (aka OpenAPI Object in v3), rather than the Info Object, as currently documented.

Thanks for considering,
Kevin


fearthecowboy:
Thanks!

kevinoid:
Sure thing.  Thank you for reviewing and merging these!

PR #3456, opened 2020-03-17
Fix failing to write configuration artifacts

fearthecowboy:
(async read problem.)

### DIFF
--- a/core/lib/pipeline/plugins/emitter.ts
+++ b/core/lib/pipeline/plugins/emitter.ts
@@ -48,8 +48,6 @@ async function emitArtifact(config: ConfigurationView, uri: string, handle: Data
   if (isObject) {
     const sink = config.DataStore.getDataSink();
 
-    // const ast = await handle.ReadYamlAst();
-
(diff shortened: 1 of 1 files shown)


core/lib/pipeline/plugins/emitter.ts
```diff
@@ -47,23 +47,21 @@ async function emitArtifact(config: ConfigurationView, uri: string, handle: Data
 
   if (isObject) {
     const sink = config.DataStore.getDataSink();
-    const object = new Lazy<any>(() => handle.ReadObject<any>());
```
daviwil: Just out of curiosity, what's the difference between this and awaiting directly? Does Lazy use some other means to pull the result?

fearthecowboy: Johannes made a Lazy class to mirror .NET's lazy early in the autorest days, as a way of being able to delay the action if never needed. Lazy waits until you actually ask for it (and then if it was a promise, you'd have to await the result anyway) the idea was to never do the work if the value was never actually asked for.


it got a bit messy with promises, and really, needs to be removed. it's a blight.

PR #3478, opened 2020-04-08
preserve security info

fearthecowboy:


core/test/configuration.ts
```diff
@@ -47,9 +47,6 @@ csharp:
     // output folder should be 'foo'
     assert.equal(cfg['output-folder'], 'foo');
 
-    // my-value should not get resolved here because the dependent variable is after
-    assert.equal(cfg['my-value'], '$(sample-value)');
```
daviwil, line 51: Seems unrelated, why does it need to be removed?

fearthecowboy: It was breaking a test (configuration doesn't work like that anymore)... I dunno why I didn't catch it before.

PR #3480, opened 2020-04-13
identify anonymous schema during shaking

fearthecowboy:


core/lib/pipeline/plugins/tree-shaker.ts
```diff
@@ -606,6 +606,9 @@ export class OAI3Shaker extends Transformer<AnyObject, AnyObject> {
 
     // copy the parts of the parameter across
     visitor.bind(this)(tc, children);
+    if (isAnonymous) {
+      tc['x-anonymous-schema'] = { value: { anonymous: true }, pointer: '' };
```
daviwil, line 610: Is this a known name? Should we use an `x-autorest-` prefix on extensions we invent?

fearthecowboy: 'known' -- no, just an internal name. hmm. We could make it x-autorest-XXX -- not a bad idea

PR #3505, opened 2020-05-21
[Documentation] Add note about nextPage operation method

joheredi:
Adding a note stating that the request to get the next page from a paged operation will always use `GET`


daviwil:
Thanks!

PR #3523, opened 2020-06-19
Don't hard code max-old-space-size

jdelforno:
Windows 10

Tried to execute the app.js file directly with both NODE_OPTIONS and then of course, max-old-space-size as as an option against node.exe. Both failed.

Looked into the app.js file and found the option was hard coded. 

This effectively overrides the system configured option.

### DIFF
--- a/autorest/entrypoints/app.js
+++ b/autorest/entrypoints/app.js
@@ -2,7 +2,7 @@
 // load modules from static linker filesystem.
 // if the process was started with a low heap size (and we're not debugging!) then respawn with a bigger heap size.
 if ((require('v8').getHeapStatistics()).heap_size_limit < 8000000000 && !(require('inspector').url() || global.v8debug || /--debug|--inspect/.test(process.execArgv.join(' ')))) {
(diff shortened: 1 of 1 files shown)


RupengLiu:
Strongly agree here, I ran into this issue, please merge to unblock

vrdmr:
+ @daviwil - We are seeing this as well.

https://github.com/Azure/autorest/issues/3587 is the issue that comes up when @RupengLiu is trying to use Autorest.

daviwil:
@RupengLiu what type of machine are you running this on, is it a virtual machine?  How much RAM is available?  I'm wondering if we should just add a parameter for increasing the memory size instead of trying it automatically.  Apparently this will only matter on Node 10 and below, Node 12 now sizes the heap based on the system's available memory.

daviwil:
@jdelforno Thanks a lot for this contribution!  We decided that we do want to remove the use of `--max-old-space-size` by default but provide the ability to set it via a new parameter.  I've opened PR #3589 to make this change, closing yours in favor of the newer one.

daviwil:
Just released a new version of AutoRest with the fix (3.0.6247), let me know if it resolves the issue for you @RupengLiu.

Issue #3516, opened 2020-06-09
Directive/transform does not work as expected by using common parameters in the OpenAPI doc


tibortakacs:
**Summary**
The OpenAPI spec allows to use common parameters in the description (https://swagger.io/docs/specification/describing-parameters/ - Common Parameters section). However, if the `directive` and `transform` tags are used in the `readme.md` AutoRest config in order to transform the operationIds, the generation successfully finishes but without the expected transformation.

Example:
**readme_gwd.md**
```
# PetStore
> see https://aka.ms/autorest
This is the AutoRest configuration file for the APIs.
---
#### Basic Information
These are the global settings for APIs.
``` yaml
# list all the input OpenAPI files (may be YAML, JSON, or Literate- OpenAPI markdown)
input-file:
  - gatewayd_simple.yaml
# this allows you to programatically tweak the swagger file before it is modeled.
directive:
  from: openapi-document # do it globally
  where: $.paths.*.*
  # set each operationId to 'MyAPI<Tag>'
  transform: $.operationId = `${$.tags[0]}_${$.operationId}`
```

**gatewayd_simple.yaml**
```
openapi: "3.0.0"

info:
  version: 4.1.0
  title: Azure Sphere Device REST API
  description: REST API exposed by Azure Sphere devices via the Servicing Port

servers:
  - url: https://192.168.35.2

paths:
  /abi_versions:
    get:
      operationId: "getAbiVersions"
      tags:
        - "images"
      summary: "Get ABI versions"
      responses:
        200:
          description: "ABI version set"
          content:
            application/json:
              schema:
                type: string
  /app/status/{componentId}:
    parameters:
      - in: path
        name: componentId
        required: true
        description: "Component ID of application"
        schema:
          type: string
    get:
      tags:
        - "app"
      operationId: "getAppStatus"
      summary: "Get app status"
      responses:
        200:
          description: "App status"
          content:
            application/json:
              schema:
                type: string
```
**Autorest script**
`autorest --v3 --python --use=@autorest/python@5.1.0-preview.1 .\readme_gwd.md --debug`

The `/app/status/{componentId}` endpoint has a common parameter. The interesting point is that none of the transformation works not just there where the common parameter is used.

tibortakacs:
@daviwil I finished the investigation, please find my analysis here:
[...]

PR #3525, opened 2020-06-25
Solve common parameter vs transformation issue.

tibortakacs:
This draft PR shows the concept of a possible solution for the common parameter vs. transformation issue. This change let individual transformations fail without stopping the transformation pipeline execution. See the details in the bug ticket here: https://github.com/Azure/autorest/issues/3516


core/test/object-manipulator.ts
```diff
@@ -121,4 +125,31 @@ definitions:
       assert.strictEqual(resultObject.paths['/api/circular'].get.description, 'FUN TIME');
     }
   }
+
+  @test async 'skip'() {
```
daviwil, line 129: Was this meant to be skipped?

core/lib/pipeline/object-manipulator.ts
```diff
@@ -5,13 +5,17 @@
...
+import { Channel, Message, SourceLocation } from '../message';
 import { IdentitySourceMapping } from '../source-map/merging';
 
 export async function manipulateObject(
   src: DataHandle,
   target: DataSink,
   whereJsonQuery: string,
   transformer: (doc: any, obj: any, path: JsonPath) => any, // transforming to `undefined` results in removal
+  config?: ConfigurationView,
+  extraTransforationContext?: string,
```
daviwil, line 18: Small typo, I'd also rename it to this:

```suggestion
  transformationString?: string,
```

core/test/object-manipulator.ts
tibortakacs: Yeah, `skip` is not best name. What about `skip-transform-failure`?

core/lib/pipeline/object-manipulator.ts
```diff
@@ -46,12 +50,32 @@ export async function manipulateObject(
...
+      const oldAst = ResolveRelativeNode(ast, ast, hit.path);
+      ast = ReplaceNode(ast, oldAst, newAst) || (() => { throw new Error('Cannot remove root node.'); })();
+    } catch (err) {
+      // Background: it can happen that one transformation fails but the others are still valid. One typical use case is
+      // the common parameters versus normal HTTP operations. They are on the same level in the path, so the commonly used
+      // '$.paths.*.*' "where selection" finds both, however, most probably the transformation should and can be executed
+      // either on the parameters or on the HTTP operations, i.e. one of the transformations will fail.
+      if (config != null) {
+        let errorText = `Directive with 'where' clause '${whereJsonQuery}' failed by path '${hit.path}`;
+        if (extraTransforationContext != null) {
```
daviwil, line 68: You'll need to update references of `extraTransforationContext` to `transformationString` here and in the error text below.  Also, now that I look at this again `transformString` is probably the right name, do you mind changing it one more time?

core/test/object-manipulator.ts
```diff
@@ -121,4 +125,31 @@ definitions:
       assert.strictEqual(resultObject.paths['/api/circular'].get.description, 'FUN TIME');
     }
   }
+
+  @test async 'skip-transform-failure'() {
```
daviwil, line 129: Ahhh, somehow I blanked out on the fact that this is the test name, not an indicator that the test is skipped :)  This name is better, thanks!

core/lib/pipeline/object-manipulator.ts
tibortakacs: Ah, for sure :) I just used the Github feature to accept your recommendation, and dummily have to changed the other part of the code. Sorry for that and thank you for spotting it out. Interestingly, my local compilation did not complain that something is wrong with the name, this might a TypeScript/Javascript thing that I have to learn.

tibortakacs: By the way, this was my original naming as well. I changed it `extraTransformationContect` to avoid the confustion that it is _not_ used in the transformation, the `transformer` lambda does the job. I am happy with the `transformationString`, just meant to give you background of it.

daviwil: Sounds good!

daviwil:
Thanks for taking the time to investigate and fix this issue!  I'm merging it now despite the regression test failure since it is unrelated to this change.

PR #3538, opened 2020-07-14
Revert unintended changes from PR #3534

daviwil:
Some files with a bunch of temporary edits made it into PR #3534 which caused a lot of spurious output for @pakrym.  This change reverts that mess.


daviwil:
Merging this despite the regression test failure, I believe that was brought on by a different change.  I'll investigate a fix for that in a separate PR so that we can get past this errant output for now.

PR #3540, opened 2020-07-15
Allow all `x-` extensions on the `xml` object

daviwil:
This change enables the solution to https://github.com/Azure/autorest/issues/3535 by allowing all `x-` extensions to be applied to the `xml` object when configuring XML serialization for a schema.  Specifically, this allows an `x-ms-text` attribute to be passed through to the modelerfour CodeModel under the `extensions` sub-object.  Language generator implementors can use this information to serialize/deserialize the textual contents of an XML node into a property.

### DIFF
--- a/schema/swagger-extensions.json
+++ b/schema/swagger-extensions.json
@@ -1480,10 +1480,7 @@
         }
       },
       "patternProperties": {
-        "^x-(?!ms-).*$": {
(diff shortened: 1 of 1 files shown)


daviwil:
Regression failure was caused by a different change that merged recently, I'll investigate/fix this in another PR tomorrow.

PR #3553, opened 2020-07-30
Enable formData parameters to have arrays of files

daviwil:
This change updates our `swagger-extensions.json` schema to allow operation parameters in `formData` to specify arrays with items of type `file` for the purpose of enabling multi-file uploads for the `multipart/form-data` content type.  I basically just copied the `primitivesItems` definition and added `file` to the `type` property's enum set.

### DIFF
--- a/schema/swagger-extensions.json
+++ b/schema/swagger-extensions.json
@@ -935,7 +935,7 @@
           "type": "string"
         },
         "items": {
-          "$ref": "#/definitions/primitivesItems"
+          "$ref": "#/definitions/primitivesItemsWithFile"
         },
         "collectionFormat": {
           "$ref": "#/definitions/collectionFormatWithMulti"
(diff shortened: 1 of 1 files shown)


schema/swagger-extensions.json
```diff
@@ -1457,6 +1457,88 @@
         }
       }
     },
+    "primitivesItemsWithFile": {
```
fearthecowboy, line 1460: Just wondering -- can this be implemented using `allOf` `primitivesItems` (so we're not copy/pasting the whole definition?) - in the event the `primitivesItems` changes in the future, it'd be nice to propagate automatically

daviwil: Yep, I can give it a shot.

daviwil: Thanks for the suggestion and figuring out how to make this work behind the scenes.  I've deduplicated the common parts into `primitivesItemsBase` so that only the enum values of the two inherited schemas will be specified.

PR #3582, opened 2020-09-09
Update node js check to permit/suggest v14

fearthecowboy:


daviwil (approved):
Cool! So it works?

fearthecowboy:
I've been using node 14 for all of 2020 so far. I've had no problems at all (after we fixed the sandbox issue a long time ago, it's been great)

daviwil:
Sounds good to me!

PR #3586, opened 2020-09-11
Only turn on client-side-validation for the v2 pipeline

daviwil:
This change explicitly disables the `client-side-validation` option when the `--v3` flag is used so that it doesn't get applied for AutoRest v3 generators.

### DIFF
--- a/core/resources/plugin-validators.md
+++ b/core/resources/plugin-validators.md
@@ -47,8 +47,8 @@ pipeline:
 
 ## Client Side Validation
 
-On by default for backwards compatibility, but see https://github.com/Azure/autorest/issues/2100
+On by default in AutoRest v2 for backwards compatibility, but see https://github.com/Azure/autorest/issues/2100.
 
-``` yaml
+``` yaml $(pipeline-model) == 'v2'
 client-side-validation: true
 ```


core/resources/plugin-validators.md
```diff
@@ -47,8 +47,14 @@ pipeline:
...
+On by default in AutoRest v2 for backwards compatibility, but see https://github.com/Azure/autorest/issues/2100.
 
 ``` yaml
 client-side-validation: true
 ```
+
+For AutoRest v3 generators we turn it off by default.
+
+``` yaml $(v3)
+client-side-validation: false
```
iscai-msft, line 59: @lmazuel is still deciding whether he wants this false by default, or `None`

iscai-msft (approved):
Going to approve too, thanks @daviwil for the quick turn around!

iscai-msft (approved):

lmazuel (approved):

daviwil:
Regression test failure for Python is expected in this case because `client-side-validation` is now `false`, resulting in this output change:

```
self._serialize.client_side_validation = False
```

Fix has been verified by @iscai-msft from a local build, merging it!

PR #3590, opened 2020-09-16
Default --python to AutoRest v3 and @autorest/python

daviwil:
Since `@autorest/python` is now GA (congrats! 🎉) this change will default all requests for `--python` to use the new generator without requiring `--v3` to also be supplied as a parameter.  Note that using `--v3` will still work and we'll also support `--v2` for falling back to the old V2 generation and core pipeline.

### DIFF
--- a/.scripts/regression-compare.yaml
+++ b/.scripts/regression-compare.yaml
@@ -62,10 +62,10 @@ languages:
   - language: python
     outputPath: ../core/test/regression/python
     oldArgs:
-      - --use:@autorest/python@5.4.0
+      - --version:~3.0.6320
+      - --use:@autorest/python
     newArgs:
       - --version:../core
-      - --use:@autorest/python@5.4.0
   - language: typescript
     outputPath: ../core/test/regression/typescript
     oldArgs:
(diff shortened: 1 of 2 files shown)


.scripts/regression-compare.yaml
```diff
@@ -62,12 +62,10 @@ languages:
...
     outputPath: ../core/test/regression/python
     oldArgs:
-      - --v3
-      - --use:@autorest/python@5.1.0-preview.3
+      - --use:@autorest/python@5.3.2
     newArgs:
       - --version:../core
-      - --v3
-      - --use:@autorest/python@5.1.0-preview.3
+      - --use:@autorest/python@5.3.2
```
iscai-msft, line 68: is it possible to not tie this to the specific version 5.3.2, but just get the latest autorest python?

daviwil: Good catch!  Yep, I'll remove the version.

daviwil:
Verified with @iscai-msft that this change is working for `@autorest/python` without the `--v3` parameter and with `@microsoft.azure/autorest.python` with the `--v2` parameter.

daviwil:
@lmazuel @johanste I forgot where we landed on deciding to flip the switch for `@autorest/python` being the new default.  Are we still postponing that or should I merge this change?

daviwil:
Closing this in favor of a new PR that upgrades multiple languages to V3 by default.

PR #3617, opened 2020-10-22
add docs for x-ms-client-default

iscai-msft:


docs/extensions/readme.md
```diff
@@ -791,6 +792,59 @@ Note: The extension is not tight to this particular scenario (you could model an
...
 
+## x-ms-client-default
+Set the default value for a property or a parameter.
+
+With this extension, you can set a default value for a property or parameter that is independent of how the property / parameter's schema is handling a default
+
+**Parent element**: [Parameter Objects](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject) or [Property on the Schema Definition](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject).
+
+**Schema**:
+`string` - the default value for the property / parameter
```
daviwil, line 803: In modelerfour, `x-ms-client-default` supports `string`, `integer`, `long`, `float`, `double`, and `boolean` so we might need to add these possibilities here too

daviwil (approved):
Thanks for adding this!  Just one thing that needs to be updated, otherwise it looks good

PR #3626, opened 2020-11-04
add docs and example for tokenParamName

iscai-msft:


docs/extensions/readme.md
```diff
@@ -796,7 +796,7 @@ Note: The extension is not tight to this particular scenario (you could model an
 Set the default value for a property or a parameter.
 
 With this extension, you can set a default value for a property or parameter that is independent of how the property / parameter's schema is handling a default. This is different than the `default` value
-you can speciy
+you can specify
```
iscai-msft, line 799: unrelated spelling fix

docs/extensions/readme.md
```diff
@@ -858,6 +858,7 @@ Field Name | Type | Description
 itemName | `string` | Optional (default: `value`). Specifies the name of the property that provides the collection of pageable items.
 nextLinkName| `string` | Required. Specifies the name of the property that provides the next link (common: `nextLink`). If the model does not have a next link property then specify `null`. This is useful for services that return an object that has an array referenced by `itemName`. The object is then flattened in a way that the array is *directly* returned, no paging is used. This provides a better client side API to the end user.
 operationName | `string` | Optional (default: `<operationName>Next`). Specifies the name of the operation for retrieving the next page.
+tokenParamName | `string` | Optional (default: `nextLink`). If you are looking to pass the continuation token through an API parameter to subsequent calls, use this field to specify the name of the parameter that will take the continuation token
```
jhendrixMSFT, line 861: Does it make sense to specify a default value here?  If it does, how about `continuationToken` or something similar?

iscai-msft: `nextLink` is in a sense the current default value, because say if you want to pass a continuation token into a separate next operation: we currently can do this if the input param name is `nextLink`. I can also remove the default if you think it will confuse users, what are your thoughts?

jhendrixMSFT: Hmm, perhaps the doc for `nextLinkName` should also be updated to state that it's a next link URL or a continuation token.  For the latter, we can have a "common: `token`" or something so we can distinguish between the two cases.

For `tokenParamName` we could say something like "Contains the name of the continuation token parameter for the case where `nextLinkName` contains a continuation token instead of a URL."

iscai-msft: I like that, thanks @jhendrixMSFT !

jhendrixMSFT (approved):

docs/extensions/readme.md
```diff
@@ -856,8 +856,9 @@ The REST API guidelines define a common pattern for paging through lists of data
...
+nextLinkName| `string` | Required. Specifies the name of the property that provides the next link or a continuation token (common: `nextLink` for the former, `token` for the latter). If the model does not have a next link property then specify `null`. This is useful for services that return an object that has an array referenced by `itemName`. The object is then flattened in a way that the array is *directly* returned, no paging is used. This provides a better client side API to the end user.
 operationName | `string` | Optional (default: `<operationName>Next`). Specifies the name of the operation for retrieving the next page.
+tokenParamName | `string` | Contains the name of the continuation token parameter for the case where `nextLinkName` contains a continuation token instead of a URL.
```
pakrym, line 861: Should it be nextLinkParameter name?

docs/extensions/readme.md
```diff
@@ -996,6 +997,45 @@ Full code:
 [example3.cs](x-ms-pageable/example3.cs)
 
 
+**Example 4: Passing the continuation token through an API parameter**
```
pakrym, line 1000: Do we need a sample that has operationName  + tokenParamName?

docs/extensions/readme.md
iscai-msft: We're trying to stick more with `token` instead of `nextLink` in the name, since next link is more about using that link to make the next call, while a `token` refers to a more generic continuation token

docs/extensions/readme.md
iscai-msft: I don't think we need both at this time, and I wanted to keep it simpler (and also wanted to write less), so I went with not defining a separate next operation. I do see where it should be more clear that this will also work with a separate next operation, I'll add that to the `tokenParamName` description

pakrym (approved):

docs/extensions/readme.md
```diff
@@ -856,8 +856,9 @@ The REST API guidelines define a common pattern for paging through lists of data
...
+nextLinkName| `string` | Required. Specifies the name of the property that provides the next link or a continuation token (common: `nextLink` for the former, `token` for the latter). If the model does not have a next link property then specify `null`. This is useful for services that return an object that has an array referenced by `itemName`. The object is then flattened in a way that the array is *directly* returned, no paging is used. This provides a better client side API to the end user.
 operationName | `string` | Optional (default: `<operationName>Next`). Specifies the name of the operation for retrieving the next page.
+tokenParamName | `string` | Contains the name of the continuation token parameter for the case where `nextLinkName` contains a continuation token instead of a URL. The param needs to be present in the operation used for retrieving the next page, whether that operation has a separate swagger definition or not.
```
iscai-msft, line 861: from @lmazuel: just say "Contains the name of the continuation token parameter that is present in the next operation. Default value is `nextLink`

docs/extensions/readme.md
```diff
@@ -856,8 +856,9 @@ The REST API guidelines define a common pattern for paging through lists of data
 Field Name | Type | Description
 ---|:---:|---
 itemName | `string` | Optional (default: `value`). Specifies the name of the property that provides the collection of pageable items.
-nextLinkName| `string` | Required. Specifies the name of the property that provides the next link (common: `nextLink`). If the model does not have a next link property then specify `null`. This is useful for services that return an object that has an array referenced by `itemName`. The object is then flattened in a way that the array is *directly* returned, no paging is used. This provides a better client side API to the end user.
+nextLinkName| `string` | Required. Specifies the name of the property that provides the next link or a continuation token (common: `nextLink` for the former, `token` for the latter). If the model does not have a next link property then specify `null`. This is useful for services that return an object that has an array referenced by `itemName`. The object is then flattened in a way that the array is *directly* returned, no paging is used. This provides a better client side API to the end user.
```
iscai-msft, line 859: add that it can be "$resonse.headers" etc

jhendrixMSFT (approved):

daviwil:
Hey, anything left to be done on this one?  Seems like it can be merged now, let me know and I'll take care of it!

iscai-msft:
@daviwil docs wise it's ready to go, has there been m4 support for `tokenParamName`. I think it's best to merge it once token param name support has been added for m4

PR #3627, opened 2020-11-05
Fix releases link in autorest-versioning document

mjohnst:
Link was pointing to `/relases` it should be pointing to `/releases`


daviwil (approved):

daviwil:
Thanks!

PR #3699, opened 2020-11-17
Fix the scale of memory limit from gb to MB

isra-fel:
For example, `2048 gb` seems wrong to me, should be `2048 MB`


isra-fel:
@daviwil could you take a look at this? Thanks!

daviwil (approved):
Looks good, thanks!

PR #3713, opened 2020-12-01
Update rush and pnpm to support node 12 and 14

timotheeguerin:
Node 10 is ending its life in april 2021. [Current active LTS version is Node 14](https://nodejs.org/en/about/releases/).

This PR updates rush and pnpm which supports node 12 and 14

Tested `rush install` and `rush build` on Node `10`, `12` and `14`

Seems to be an issue with node `13` in the PR validation build(https://github.com/nodejs/node/issues/31029). It also seems like there is 3 different version of node used in the 3 piplines:
- `publish.yaml`: 10
- `push-to-npm`: 12
- `verify-pull-request`: `13 -> 14`


timotheeguerin:
I see here `common/config/rush/shrinkwrap.yaml` was in the gitignore but not in `autorest.megarepo`. Should `pnpm-lock.yaml` file be? I think it defeat the point of lock files.

daviwil:
Looks like the `azure-sdk-for-js` repo checks in the `pnpm-lock.yaml` file, I'm not sure where the shrinkwrap file comes from

timotheeguerin:
Not familiar with pnpm or rush but I think that used to be the old name for the same file, In the megarepo PR:
image

daviwil:
Interesting, I'd say let's leave it checked in for now and we can remove it if we run into problems.  CI is green so I'm going to merge this.  Thanks a bunch!

daviwil (approved):

PR #3715, opened 2020-12-01
Internal: Configure prettier and eslint and add to the CI

timotheeguerin:
Setup `prettier` and fix existing `eslint` setup to be able to run it in the CI.

Ran `prettier . --write` to fix all the existing format issues.

Files of interest that were modified not just by the formatter are:
- `.default-eslintrc.yaml`
- `.prettierignore`
- `.prettierrc.yml`
- `.scripts/verify-pull-request.yaml`
- `.vscode/settings.json`
- `common/config/rush/command-line.json`
- `core/package.json`
- `autorest/package.json`


Samples/openapi-v2/1a-code-generation-minimal/pétstöre.json
```diff
@@ -15,26 +15,16 @@
   },
   "host": "petstore.swagger.io",
   "basePath": "/v2",
```
daviwil, line 17: I think we should exclude all the files in the Samples folder from being formatted, the choices that Prettier makes seems to hurt readability and expected formatting somewhat here (and in the YAML files too)

.prettierrc.yml
```diff
@@ -0,0 +1,6 @@
+trailingComma: 'all'
+printWidth: 120
+quoteProps: 'consistent'
+endOfLine: 'auto'
+arrowParens: always
+singleQuote: true
```
daviwil, line 6: After seeing all of the quote changes accross the files, I think I'd rather stick with double quotes as the standard, it seems pretty disruptive to switch to single quotes everywhere.  What are your thoughts on that?

core/test/resources/code-model.yaml
```diff
@@ -1,656 +1,421 @@
 {
```
daviwil, line 1: I think we should avoid formatting all YAML and JSON files under the `core/test/resources` path too

daviwil (commented):
The TypeScript files look massively better!  I think we should use double-quotes instead of single quotes, though, that change might be pretty disruptive.  There's also some JSON/YAML files we shouldn't format because it makes them a little harder to scan (many Swagger/OpenAPI/CodeModel documents need some breathing room).

.prettierrc.yml
timotheeguerin: I don't really care which one we use, what do the other repo uses? I have been using double quotes in my recent project but single previously too. Here all the typescript files were using single quotes(Why I picked that), so it will make all kind of other changes in the ts files instead.

Samples/openapi-v2/1a-code-generation-minimal/pétstöre.json
timotheeguerin: ok I can exclude those folders

daviwil (approved):

daviwil:
Here we go :)

PR #3722, opened 2020-12-08
Fix: Custom name conflict

### LINKED ISSUE #3655: M4 produces duplicated model errors where it shouldn't
https://github.com/Azure/azure-rest-api-specs/blob/d9cf7c7ed3d674ebd482836e82b274014245ae67/specification/redis/resource-manager/readme.md 
and
https://github.com/Azure/azure-rest-api-specs/blob/d9cf7c7ed3d674ebd482836e82b274014245ae67/specification/storage/resource-manager/readme.md

Are affected.

Worked fine in 4.13.356 and adding the switch doesn't change the generated code.

timotheeguerin:
fix #3655

When extracting schemas tree-shaker was copying the `x-ms-client-name` property down to the model even though it was mean to describe the property name or parameter name. This could cause issues down the line with duplicate model names

### DIFF
--- a/.default-eslintrc.yaml
+++ b/.default-eslintrc.yaml
@@ -27,6 +27,7 @@ rules:
   "@typescript-eslint/no-unused-vars": "off"
   "@typescript-eslint/no-parameter-properties": "off"
   "@typescript-eslint/no-angle-bracket-type-assertion": "off"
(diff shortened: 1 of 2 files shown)


daviwil (approved):
Awesome work!

PR #3737, opened 2020-12-15
Internal: Migrate tests to jest

timotheeguerin:
Migrating to jest to be able to use some of the functionalities of jest:
- Code coverage built in
- Has all the needed dependencies for testing, no need to install many more as mocha require(Spies, expect, etc.)
- Snapshots

This also make it that test are compiled on demand making testing much easier(No need to compile before)

A change this required was to make a new `AppRoot` constant for the core package that represent the path to the root folder of the package. This ensure that we can reference files relative to the root that work regardeless of the environment it is running on(From the dist/ folder or directly from the ts file)


core/lib/constants.ts
```diff
@@ -1,7 +1,25 @@
+import { paths } from "jsonpath";
+import { resolve } from "vscode-languageserver/lib/files";
```
daviwil, line 2: Looks like these got added above the copyright header, could you move them down?

core/lib/pipeline/schema-validation.ts
```diff
@@ -8,15 +8,23 @@ import * as SchemaValidator from "z-schema";
 import { OperationAbortedException } from "../exception";
 import { Channel } from "../message";
 import { createPerFilePlugin, PipelinePlugin } from "./common";
+import * as path from "path";
+import { AppRoot } from "../constants";
+
+// TODO-TIM: Find a better way? Move schema to a package?
+const schemaFolder = process.env.JEST_WORKER_ID === undefined ? "." : path.resolve(AppRoot, "../schema");
```
daviwil, line 15: Moving the schemas to a package is an interesting idea, for sure.

daviwil (approved):

core/lib/pipeline/schema-validation.ts
timotheeguerin: yeah wondering if we can have a non published package with rush which just get statically linked

timotheeguerin: maybe just with the bundledDependencies it would work

PR #3738, opened 2020-12-15
Created internal package for schemas

timotheeguerin:
Todo: 
- [ ] Check this is working fine if the package is coming from somewhere else


autorest-schemas/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@autorest/schemas",
+  "version": "1.0.0",
+  "description": "Internal package storing schemas",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
```
daviwil, line 8: Does this need the `files` property to include those json files in the package?

timotheeguerin: nope they get included with bundledDependencies. I think npm will include everything expect what's defined in npmignore

daviwil (approved):
This is great, I love it

PR #3740, opened 2020-12-16
Move AutoRest package folders under 'packages/'

daviwil:
This change moves the AutoRest package folders under the 'packages/' path with categorized subdirectories to organize everything:

- `packages/apps/` - For now, just holds the `autorest` command line app
- `packages/libs/` - Holds any helper libraries used by AutoRest
- `packages/extensions/` - Holds any AutoRest extensions.  I'm considering `@autorest/core` an extension in this case because it is also loaded as an external process like other extensions

Eventually we'll also add `packages/tools/` when we move over `@autorest/compare` and `@autorest/testserver`.


daviwil:
I could also call it `packages/libraries` instead of `libs`, what do you think?

timotheeguerin:
I like libs, i think its an obvious enough short name that makes it a little nicer of a path.
For `extensions` should it not be `plugins`?

timotheeguerin (approved):

daviwil:
> For `extensions` should it not be `plugins`?

I think "extensions" is the standard terminology for anything that gets pulled in via `--use`:

https://github.com/Azure/autorest/blob/7ae6b36bb9c4841087bab3326114f10c84f12379/docs/autorest-versioning.md#autorest-versioning

I'll go ahead and get this merged so that it doesn't hold up your work, we can definitely revisit the folder naming again!

Issue #3606, opened 2020-10-12
TypeError: Cannot read property 'None' of undefined on V3 Swagger interface (Smarkets)


catapedamania:
```
AutoRest -CodeGenerator CSharp -Input https://api.smarkets.com/v0/control/openapi/?tags=trading_api -Namespace Smarkets --v3

AutoRest code generation utility [cli version: 3.0.6247; node: v10.16.0, max-memory: 1280 gb]
(C) 2018 Microsoft Corporation.
https://aka.ms/autorest
  TypeError: Cannot read property 'None' of undefined
```

timotheeguerin:
So there is 2 part to this issue I believe. 
1. An issue with emoji, have a fix for that.
2. You are using `oneOf` in the payload which is not supported in autorest https://github.com/Azure/autorest#openapi3-support.

PR #3741, opened 2020-12-16
Update yaml-ast-parser and uptake fix in @azure-tools/datastore

timotheeguerin:
Fix first part of the problem here  Azure/autorest#3606


daviwil (approved):
Great!  Can you wait until after I get #3742 merged and verified before merging this?  Just want to make sure a new version actually gets published with your fix

daviwil:
Publishing is working correctly now!  I'll merge your PR.

PR #3744, opened 2020-12-16
Update extension package to 3.1.0

timotheeguerin:
Update package to contain yarn -> npm migration to fix core package install

### DIFF
--- a/packages/apps/autorest/package.json
+++ b/packages/apps/autorest/package.json
@@ -55,7 +55,7 @@
     "static-link": "^0.3.0",
     "@autorest/core": "~3.0.0",
     "@azure-tools/async-io": "~3.0.0",
-    "@azure-tools/extension": "~3.0.0",
+    "@azure-tools/extension": "~3.1.0",
     "@azure-tools/tasks": "~3.0.0",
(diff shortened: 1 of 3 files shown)


daviwil (approved):
Thanks Tim!

daviwil:
Looks like the lockfile needs to be refreshed, that'll help get the CI bumped to start again.

packages/apps/autorest/package.json
```diff
@@ -55,7 +55,7 @@
     "static-link": "^0.3.0",
     "@autorest/core": "~3.0.0",
     "@azure-tools/async-io": "~3.0.0",
-    "@azure-tools/extension": "~3.0.0",
+    "@azure-tools/extension": "~3.1.0",
```
daviwil, line 58: I believe this should be a real dependnecy of AutoRest since it gets used as part of the runtime code.

timotheeguerin: yeah, i am still having issues, I think bundledDependencies is acting weird with pnpm

daviwil: Oh great...  I hadn't considered that possibility

PR #3749, opened 2020-12-17
Cleanup autorest cli and fix using core version from config file

### LINKED ISSUE #3746: @autorest/core didn't pick the pinned version "3.0.6326", instead it picked the latest version
We pinned `version: 3.0.6326` in `autorest.csharp` repo [here](https://github.com/Azure/autorest.csharp/blob/feature/v3/readme.md#L1019
) but autorest picked `'@autorest/core' - '3.0.6348'` version instead.

For reference  check [here](https://dev.azure.com/azure-sdk/public/_build/results?buildId=657655&view=logs&j=b5cdfb1e-c180-561f-918e-a169e451719e&t=aff748e3-476c-56d1-ae31-4029fde28653&l=52).

timotheeguerin:
Reorganize and cleanup a bit the CLI.
Fixes loading from a config from the current directory.

fix #3746


packages/apps/autorest/src/args.ts
```diff
@@ -0,0 +1,83 @@
...
+ * --foo=bar
+ * --foo
+ */
+const FLAG_REGEX = /^--([^=:]+)([=:](.+))?$/;
+
+/**
+ * Parse a list of command line arguments.
+ * @param argv List of cli args(process.argv)
+ */
+export const parseArgs = (argv: string[]): AutorestArgs => {
```
timotheeguerin, line 38: Seems like autorest core is doing more or less the same thing. Might be good to share the code.

packages/apps/autorest/src/core-version-utils.ts
```diff
@@ -0,0 +1,81 @@
...
+    const sandbox = {
+      require,
+      console,
+      rfs: {
+        EnumerateFileUris: async (folderUri: string): Promise<Array<string>> => {
+          return EnumerateFiles(folderUri, ["readme.md"]);
+        },
+        ReadFile: (uri: string): Promise<string> => ReadUri(uri),
+      },
+      cfgfile: ResolveUri(`file://${cwd}`, args.configFileOrFolder || ""),
```
timotheeguerin, line 43: This is pretty much the fix going from 
```
cfgfile: ResolveUri(cwd, args.configFileOrFolder || "."),
```
to
```
cfgfile: ResolveUri(`file://${cwd}`, args.configFileOrFolder || ""),
```

The previous implementation would for `cwd = "/dev/test-folder"` it would return `null:///dev`. So both setting null as the protocol and going up to the parent folder. 

Wondering if this is something that is inconsistent between windows and unix

.gitignore
```diff
@@ -160,7 +160,7 @@ Backup*/
 UpgradeLog*.XML
 
 # NuGet
-packages
+# packages
```
daviwil, line 163: Wow!  That would be a problem

daviwil (approved):
This is great!

.gitignore
timotheeguerin: yeah, it was also causing vscode not to search inside that folder

PR #3773, opened 2021-01-11
fix doc links

iscai-msft:


daviwil (approved):

timotheeguerin:
Wonder if we could use something like https://github.com/tcort/markdown-link-check to verify the links are not getting out of sync

daviwil:
That'd be a nice CI check to run on PRs

iscai-msft:
@daviwil @timotheeguerin that's fair, i wouldn't say to add it on behalf of me though. My links issues largely came bc I was adding the docs in the language geerators at the same time, so I was linkign to stuff that hadn't existed yet. Also the diff isn't as big as it seems here: I was also switching all links from relative to the full URL path since I know there's some issues with relative liks.

PR #3775, opened 2021-01-11
Internal: Lint dead links in docs.

timotheeguerin:
Related #3773

Added a new Job to the CI(New job and not step as its quite long to validate each links) and fix existing issues

### DIFF
--- a/.scripts/verify-pull-request.yaml
+++ b/.scripts/verify-pull-request.yaml
@@ -52,6 +52,6 @@ jobs:
     displayName: Lint docs
     steps:
       - script: |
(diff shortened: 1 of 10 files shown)


timotheeguerin:
@iscai-msft Looking at some of the issues there seems to be some links to corespdoning language docs which don't exists (yet). Seems like you added this as a target structure those repos should follow but are not yet? 
For example https://github.com/Azure/autorest.java/tree/v4/docs/troubleshooting.md#generation-errors

Just to confirm I should ignore those errors.

iscai-msft (approved):

PR #3776, opened 2021-01-12
add faq

iscai-msft:


timotheeguerin:
Add a link to the faq file from the readme?

timotheeguerin:
Seems like one of the link is returning a `203` which the docs link check tool doesn't accept by default. 

Can you add this to the `markdown-link-check-config.json`
```json
 "aliveStatusCodes": [200, 203]
```

iscai-msft:
> Add a link to the faq file from the readme?

you're right, I had an outtdated local version tthat already had the link

for the linting, it appears to not find any of the new autorest docs I added to master. Is it running on outdated masster?

timotheeguerin (approved):

Issue #3787, opened 2021-01-18
Could not find a declaration file for module 'autorest'


jianyexi:
**Describe the bug**
```
When import the 'autorest' package from other npm package used by typescript  , it always reports this error
```
The code is like below:

`import * as autorest from "autorest";`

I think it's caused by the incorrect 'typings' defined in the package.json

jianyexi:
The 'autorest/core' also has this issue.

jianyexi:
> **Describe the bug**
> 
> ```
> When import the 'autorest' package from other npm package used by typescript  , it always reports this error
> ```
> 
My code is like this:
`import * as autorest from "autorest";`

> I think it's caused by the incorrect 'typings' defined in the package.json

jianyexi:
@timotheeguerin  I am wondering if you have any plan to publish a new version ?

timotheeguerin:
@jianyexi I have a PR out to renable automatic publishing of `autorest` and `@autorest/core`. Should hopefully get that merged today or I'll do a manual release

PR #3788, opened 2021-01-18
Fix exported types for autorest and @autorest/core

jianyexi:
Fix https://github.com/Azure/autorest/issues/3787

### DIFF
--- a/.default-eslintrc.yaml
+++ b/.default-eslintrc.yaml
@@ -36,6 +36,7 @@ rules:
   "@typescript-eslint/consistent-type-assertions":
     - warn
     - assertionStyle: "angle-bracket"
+  "@typescript-eslint/triple-slash-reference": "off"
   no-undef: "off"
   no-unused-vars: "off"
   linebreak-style:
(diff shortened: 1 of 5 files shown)


jianyexi:
> Fix issue #3787

I looked at the v2/autorest to fix this issue, as I found the v2/autorest works well

packages/apps/autorest/src/exports.ts
```diff
@@ -3,6 +3,9 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+/// <reference path="../definitions/core.d.ts" />
```
timotheeguerin, line 6: Are those lines actually needed?

.default-eslintrc.yaml
```diff
@@ -36,6 +36,7 @@ rules:
   "@typescript-eslint/consistent-type-assertions":
     - warn
     - assertionStyle: "angle-bracket"
+  "@typescript-eslint/triple-slash-reference": "off"
```
timotheeguerin, line 39: Don't ignore that here. Add it inline if those reference are really needed

timotheeguerin (approved):

Issue #3440, opened 2020-03-06
[typescript] cannot remove parameters from operation


xirzec:
Not sure if this is a problem with the core or with the typescript generator specifically, but when I have a README transform like the below (to remove a header param)

```yaml
directive:
  - from: swagger-document
    where: $.paths..parameters[*]
    transform: >
      if ($["$ref"] == "#/parameters/ClientRequestIdParameter") {
        return undefined;
      } else {
        return $;
      }
```

I get the following error:
```
FATAL: swagger-document/individual/transform - FAILED
FATAL: Error: Error retrieving '$.paths["/docs/$count"].get.parameters[1]' (Error: Trying to retrieve item '1' from sequence with '1' items (index out of bounds))
Process() cancelled due to exception : Error retrieving '$.paths["/docs/$count"].get.parameters[1]' (Error: Trying to retrieve item '1' from sequence with '1' items (index out of bounds))
  Error: Error retrieving '$.paths["/docs/$count"].get.parameters[1]' (Error: Trying to retrieve item '1' from sequence with '1' items (index out of bounds))
```

pakrym:
Related feature request: https://github.com/Azure/autorest.modelerfour/issues/156

timotheeguerin:
@xirzec do you know if this issue is still happening, i tried reproducing and it seems to remove the parameter just fine for me.

xirzec:
Looks like the error was reproducing for 2.0.4413 but is fixed in 3.0.6322

The big downside is it just flat doesn't work for removing the param, instead I have to replace all the parameters while removing the one I don't want, like this:

```
  - from: swagger-document
    where: $.paths..get
    transform: >
      const newParameters = [];
      for (let param of $.parameters) {
        if (param["$ref"] !== "#/parameters/ClientRequestIdParameter") {
          newParameters.push(param);
        }
      }
      $.parameters = newParameters;
```

xirzec:
This is the swagger I'm modifying btw: input-file: https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/specification/search/data-plane/Azure.Search/preview/2020-06-30/searchindex.json

timotheeguerin:
@xirzec I have actually been able to repro the issue(or a slightly different one) with those specs. One problem is that the `$ref` are getting expanded to absolute paths so I have to use this to make sure it match
``` yaml
directive:
  - from: swagger-document
    where: $.paths..parameters[*]
    transform: >
[...]

PR #3796, opened 2021-01-20
[Refactor] Cleanup code for running transform directives

timotheeguerin:
Was looking into this issue https://github.com/Azure/autorest/issues/3440, which in the end I couldn't repro but still ended up doing a cleanup of this code to reduce duplicates and have a more strictly defined external code execution and available context


khmooz (approved):

daviwil (approved):
Nice improvement!

PR #3804, opened 2021-01-25
Internal: Move Codemodel package from Azure/perks

timotheeguerin:
MIgration seperated in 2 commits:

1. Moved the package without any changes
2. Renamed to `@autorest/codemodel` instead of `@azure-tools/codemodel`


packages/libs/codemodel/package.json
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "@autorest/codemodel",
+  "version": "4.14.0",
+  "patchOffset": -1219,
```
daviwil, line 4: What does the negative do in this case?

packages/libs/codemodel/package.json
```diff
@@ -0,0 +1,59 @@
...
+  "keywords": [
+    "AutoRest",
+    "codegen"
+  ],
+  "author": "",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/Azure/autorest.powershell/issues"
+  },
+  "homepage": "https://github.com/Azure/perks/tree/master/codemodel#readme",
```
daviwil, line 36: Might need to update these

packages/libs/codemodel/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "url": "git+https://github.com/Azure/autorest.powershell.git"
+  },
+  "keywords": [
+    "AutoRest",
+    "codegen"
+  ],
+  "author": "",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/Azure/autorest.powershell/issues"
```
daviwil, line 34: This was always wrong apparently!

daviwil (approved):

packages/libs/codemodel/package.json
timotheeguerin: Makes it that the next version is 4.14.0 instead of being 4.14.1220. I can revert if we don't care but it was quite a jump

PR #3808, opened 2021-01-26
Update m4 to use new @autorest/codemodel renamed package

timotheeguerin:
Had to revert to using js-yaml 3.0 as it was breaking with inconsitent versions(In codegen). Will do the migration when codegen moves


daviwil (approved):

packages/extensions/modelerfour/changelog.md
```diff
@@ -1,103 +1,11 @@
-# Changelog for ModelerFour
+# Change Log - @autorest/modelerfour
```
daviwil, line 1: Maybe we can keep the old changelog as `changelog-old.md` or something?  It has some useful details in it.

packages/libs/codemodel/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@autorest/codemodel",
-  "version": "4.14.0",
+  "version": "4.14.1",
```
daviwil: In the past we had discussed keeping codemodel's versioning in line with the versioning of modelerfour so that someone who is using a particular version of modelerfour can use the equivalent major/minor codemodel version (patch isn't important).  Maybe we can bump it to 4.15.1 (in a separate PR?)

daviwil (approved):

packages/libs/codemodel/package.json
timotheeguerin: Hhm, so we can use rush versioning strategy too for those. I think maybe after we move all related packages we can visit what version we want to have

packages/extensions/modelerfour/changelog.md
timotheeguerin: I updated the `CHANGELOG.json` to include the changes from the existing changelog

PR #3815, opened 2021-01-28
Feature: Migrate modelerfour static-link -> webpack

timotheeguerin:
Static-link is proving to create a mess regularly and doesn't work well with automatically bumped version from rush publish.


packages/extensions/modelerfour/package.json
```diff
@@ -11,14 +11,14 @@
   "main": "dist/exports.js",
```
daviwil, line 11: Does this need to point to `dist/src/main.js` now?  I don't think `dist/exports.js` exists

daviwil (approved):
Looks great!

PR #3816, opened 2021-01-29
Move @azure-tools/oai2-to-oai3 from Azure/perks

timotheeguerin:


packages/libs/oai2-to-oai3/LICENSE
```diff
@@ -0,0 +1,29 @@
+BSD 3-Clause License
+
+Copyright (c) 2016, Mermade Software
```
daviwil, line 3: I was a little surprised to see this copyright, but it seems this library might have been the starting point for oai2-to-oai3:

https://github.com/Mermade/oas-kit/blob/master/packages/swagger2openapi

daviwil (approved):

packages/libs/oai2-to-oai3/LICENSE
timotheeguerin: hhm, is that something we should change and give credit instead?

PR #3824, opened 2021-01-29
Unspecified required property for body will mark body as required

timotheeguerin:
This is reverting part of a previous bug fix where the body was ignoring the required property.
Changing to the default to optional might cause issue.


iscai-msft:
hey @timotheeguerin latest changes aren't modifying media types anymore. Looked through the testserver and it does look like there's actually a "required": false test (see [required optional](https://github.com/Azure/autorest.testserver/blob/master/swagger/required-optional.json#L789)). I don't believe the param is defined correctly, but I fixed it locally. Currently, no param is generated for that operation. Running the core fix and m4 fix don't appear to be generating it for me either, though I'm not sure if it's because my core version is not taking. Does it work for you?

PR #3834, opened 2021-02-02
Consolidate dependencies to have consistent version across packages

### LINKED ISSUE #3807: Internal: Rush re-enable consistent version check
Disabled this for migrating other packages to major package upgrade changes from the move.

In rush.json
```
"ensureConsistentVersions": false, // Change to true
```

timotheeguerin:
fix #3807 

Notable change: 
- Updated `js-yaml` from 3.x to 4.x which had some breaking changes(which caused issue previously) but that was before codegen package used it. Info https://github.com/nodeca/js-yaml/blob/master/CHANGELOG.md#400---2021-01-03. Outcome is mostly:
  - Change how the tags are generated to be `!MyTag` instead of `!<!MyTag>`


regression-tests/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@autorest/regression-test",
```
daviwil, line 2: Since this is technically a package, should it be under `packages` somewhere?  Maybe a `packages/testing` or `packages/tools`

daviwil (approved):

regression-tests/package.json
timotheeguerin: Hhm, well it isn't a package linked to rush and really isn't meant to be published or consumed

PR #3837, opened 2021-02-02
Change property redefinition error when changing type into a warning to allow polymorphism

### LINKED ISSUE #3862: Polymorphism, redefining properties with more specific definitions and code generation
Right now there is some check rules that makes sure that if a property is redefined in a child type it doesn't change the type. 
We should allow a more flexible approach

```json
"FormField": {
    "type": "object",
    "discriminator": "propertyType",
    "required": [
        "propertyType"
    ],
    "properties": {
        "value": {
        },
        "precision": {
            "type": "number",
            "minimum": 0,
            "maximum": 1
        }
    }
},
"StringFormField": {
    "x-ms-discriminator-value": "string",
    "allOf": [
        {
            "$ref": "#/definitions/FormField"
        }
    ],
    "type": "object",
    "properties": {
        "value": {
            "type": "string"
        }
    }
},
"FormFields": {
    "description": "Base model for collectoin of form fields. Has no known field names",
    "type": "object",
    "additionalProperties": {
        "$ref": "#/definitions/FormField"
    }
},
"AnalyzedPassportFormFields": {
    "allOf": [
        {
            "$ref": "#/definitions/FormFields"
        }
    ],
    "properties": {
        "issuingCountry": {
            "$ref": "#/definitions/StringFormField"
        },
        "firstName": {
            "$ref":  "#/definitions/StringFormField"
        }
    }
},
```

Here `firstName` and `issuingCountry` both redefine from `FormFields` -> `StringFormField` . This should be allowed

This is a cross-lanaguage support, current status
- [x] Autorest Core https://github.com/Azure/autorest/pull/3837
- [ ] Python
- [ ] CSharp
- [ ] Javascript
- [ ] Go
- [ ] Java
- [ ] Swift

iscai-msft:
fix #3862


iscai-msft:
rush change

iscai-msft:
rush change

packages/extensions/modelerfour/src/quality-precheck/prechecker.ts
```diff
@@ -166,7 +166,7 @@ export class QualityPreChecker {
               const details = diff
                 .map((each) => `${each.path.join(".")} => '${each.op === "delete" ? "<removed>" : each.val}'`)
                 .join(",");
-              this.session.error(
+              this.session.warning(
```
pakrym, line 169: Are we sure all languages are able to handle this scenario?

timotheeguerin: I talked with @lmazuel about this and the idea is to keep it as a warning to tell the user that this is indeed a warning and the redefinition of this type might not work. If the question is are generators able to handle the cases that should be working right now, I do think there might be some change required to support those scenarios. @kristapratico, is doing some of those changes on the python side and can confirm.

pakrym: If someone is trying to change the property type for inherited type using this feature in C# generator the generated code would plainly not work.

What are the cases we are targeting to support?

timotheeguerin: I might be mixing different language here but I think there is 2 use cases:
- Having the child field type be inheriting the parent field type
- Having the child field have more restriction(validation) 

```cs
public class Parent
{
    public FormField Field { get; set; };
}

public class Child : Parent
{
    public StringFormField Field { get; set; };
}

public class FormField
{

}

public class StringFormField : FormField
{

}
```

pakrym: Yeah, the current C# generator doesn't support this case and it might not be trivial to implement.

Not blocking this PR just FYI.

timotheeguerin (approved):

PR #3838, opened 2021-02-03
Add backlog of tests and create new @autorest/testing common test package

### LINKED ISSUE #3724: Backlog of test to add
Issue to track some test scenario to add after we merge back the repos into a single one and we can consolidate the test setup.

**Tree Shaker**
- [x] Test changes related to  `x-ms-client-name` not included in the tree shaked schema [PR 3722](https://github.com/Azure/autorest/pull/3722)
   - [x] Not included when it is refering to a property
   - [x] Still included when refering to a parameter, header, etc.
   - [x] Still included when provided directly on the schema.
 

**OAI2To3Converter**
- [x] Test changes related to cross-file references [PR 131](https://github.com/Azure/perks/pull/131)
    - [x] Test referencing a body parameter from another file
    - [x] Make sure there is enough test for other type of cross-file references(Schemas, responses, regular params)
    - [x] Check we can remove the if statement for the ref being in the same file.(Check todo in code)

- [x] Test for `x-ms-parameterized-host` cross file support [PR 133](https://github.com/Azure/perks/pull/133)

**AutorestCore**
- [ ] Check all disabled tests
- [ ] Check test testing stuff in another package shouldn't be there

timotheeguerin:
fix  #3724


packages/libs/oai2-to-oai3/src/converter.ts
```diff
@@ -817,37 +817,23 @@ export class Oai2ToOai3 {
...
+            dereferencedParameter.in === "formData"
+          ) {
+            childIterator = () => visit(dereferencedParameter, [parameterName]);
+            value = dereferencedParameter;
+            pointer = parsedRef.path;
           }
         } else {
           // TODO: Throw exception
+          // eslint-disable-next-line no-console
           console.error("### CAN'T RESOLVE $ref", value.$ref);
```
daviwil, line 837: Hmmm, this seems like it shouldn't be there anyway, looks like one of my old log lines 😬

rush.json
```diff
@@ -60,6 +60,12 @@
       "reviewCategory": "production",
       "shouldPublish": true
     },
+    {
+      "packageName": "@autorest/testing",
+      "projectFolder": "packages/libs/testing",
```
daviwil, line 65: This isn't really a lib in the sense that we don't ship it as a library, so it could also be in a `packages/testing` folder to go along with the regression testing "package".  Might have to name it `@autorest/test-utils` or something like that though.  If it's not worth the effort, I'm fine with leaving it as-is

daviwil (approved):

PR #3843, opened 2021-02-04
Cleanup @azure-tools/openapi

### LINKED ISSUE #3839: Revisit OpenAPI3 model
Seems like @azure-tools/openapi describe what is expected for m4(after tree shaking and other plugins)
Should be rethought to have an actual openapi3 schema.

The issue with what we have right now is none of the plugins running on OpenAPI3 model have typings, they all use any

timotheeguerin:
Thought there was an issue with having only reference to a normalized model(Only $ref allowed) but seems like the only issue preventing from using it correctly in test and other places was the use of this `Optional<T>` instead of typescript `?`, Changed that this also allowed removing depdendencies on linq

fix #3839

### DIFF
--- a/common/changes/@autorest/core/cleanup-openapi_2021-02-04-18-35.json
+++ b/common/changes/@autorest/core/cleanup-openapi_2021-02-04-18-35.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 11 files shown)


daviwil (approved):
Nice!

PR #3847, opened 2021-02-05
Add a new field tracking if a parameter is for a multipart request

### LINKED ISSUE #3835: add information on a parameter if it is part of a multipart request
Not blocked at all, this is an ask. Would be nice to include on the parameter's yaml data that it is part of a multipart request. Currently have code that works for this case, just wondering if m4 could add this code for me. Thanks!

timotheeguerin:
fix #3835

Add a new field on the codemodel `isInMultipart`(Name open to suggestion) to specify if a parameter is part of the multipart body request
image

### DIFF
--- a/packages/libs/codemodel/src/model/common/parameter.ts
+++ b/packages/libs/codemodel/src/model/common/parameter.ts
@@ -29,6 +29,11 @@ export interface Parameter extends Value {
 
   /** When a parameter is grouped into another, this will tell where the parameter got grouped into. */
   groupedBy?: Parameter;
+
+  /**
+   * If this parameter is to be included in a multipart request body.
+   */
+  isInMultipart?: boolean;
 }
 
 export class Parameter extends Value implements Parameter {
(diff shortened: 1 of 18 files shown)


iscai-msft (approved):
yay thanks so much @timotheeguerin !

packages/libs/codemodel/src/model/common/parameter.ts
```diff
@@ -29,6 +29,11 @@ export interface Parameter extends Value {
 
   /** When a parameter is grouped into another, this will tell where the parameter got grouped into. */
   groupedBy?: Parameter;
+
+  /**
+   * If this parameter is to be included in a multipart request body.
+   */
+  isInMultipart?: boolean;
```
daviwil, line 36: When we update any of the types in `codemodel` we also have to run the `packages/libs/codemodel/.scripts/generate-schema.js` script (no args, just run it with `node`) so that the schema files in `.resources` get updated.

daviwil (approved):
This is a great example of how having everything in the same repo makes cross-cutting changes a lot easier!

PR #3848, opened 2021-02-05
Refactoring configuration loading and creation of new @autorest/configuration package

timotheeguerin:
This is a 2nd pass at trying to reorganize the configuration loading in autorest and add stricted typing.

The end goal of this is to:
- Have a more solid configuration story(No more flags being used without documentation)
- Let the configuration be loaded directly from the cli to get rid of the need to download a version of autorest/core.

Main changes:
- Differentiate configuration(Static fix value) from the Autorest context(previously `ConfigurationView`)
- Get rid of the constructor+async init() pattern which is error prone
- Added more typings and got rid of implicit typings which let access any property on the configuration
- Created new `@autorest/configuration` package which depends

As this pr is already getting large. I decided to stop here. The issue is a lot of the configuration loading depends on shared utilities which would need to be extracted first.

Next goals:
- Extract the yaml parser tooling
- Extract the merging tools
- Extract the logging tools
- Extract the rest of the config loading
- Update autorest to load the config

Todo for this PR:
- [x] Check there isn't going to be any issue with the version loading from the cli


common/changes/autorest/feature-configuration-v2_2021-02-05-19-10.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "autorest",
+      "comment": "REthing config",
```
daviwil, line 5: Do these messages need to be updated?

```suggestion
      "comment": "Use the new @autorest/configuration package",
```

daviwil (approved):

PR #3852, opened 2021-02-08
Switch @autorest/core to use webpack

timotheeguerin:
Doing this now as adding new packages doesn't work great with `static-link`(Package needs to exists in npm to be bundled)

Todo: 
- [ ] Check work to load current version
- [x] Check no issue with importing types


daviwil (approved):
Love the config reuse, another great benefit of the monorepo!

PR #3860, opened 2021-02-09
Fix overridding config in subsequence yaml code block

### LINKED ISSUE #3625: autorest failing to override package-name
image

In this case, package-name should be overridden to `multiapiwithsubmodule#submodule`, however it's still the original (it's not just an issue bc of the `#`). Other configs, i.e. `clear-output-folder`, and `output-folder` are correctly overridden in this case

timotheeguerin:
fix #3625

Simple example of the issue is markdown as follow

````md

```yaml
output-folder: ./generated-default
```

```yaml $(python)
output-folder: ./generated-python
```
````

Bug was if the python flag was used the output-folder wasn't getting set to `./generated-python`

### DIFF
--- a/packages/extensions/core/src/lib/configuration/configuration-loader.ts
+++ b/packages/extensions/core/src/lib/configuration/configuration-loader.ts
@@ -54,7 +54,6 @@ export class ConfigurationLoader {
 
     // load config
     const hConfig = await parseCodeBlocks(contextConfig, configFile, contextConfig.DataStore.getDataSink());
-
     if (hConfig.length === 1 && hConfig[0].info === null && configFile.Description.toLowerCase().endsWith(".md")) {
       // this is a whole file, and it's a markdown file.
       return [];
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/extensions/core/src/lib/configuration/configuration-loader.ts [hunk 1]
```diff
@@ -246,7 +245,7 @@ export class ConfigurationLoader {
       configurationFiles[configFileUri] = await (await fsInputView.ReadStrict(configFileUri)).ReadData();
 
       const blocks = await this.parseCodeBlocks(await fsInputView.ReadStrict(configFileUri), await createView());
-      await addSegments(blocks, false);
+      await addSegments(blocks.reverse(), false);
```
daviwil, line 248: Seems like we shouldn't have to remember to reverse the list of blocks every time.  Is it possible that we could update `addSegments` to sort the segments based on whether they have logic fences?  How does reversing the lost of blocks help in this case?

timotheeguerin: I think the issue is that right now it counts on the list of segments to be in order of high pri -> low pri. Now if when you do addSegemnts it reverse I feel it does something you didn't asked for. I am working on reorganizing all the config loading. See the other PR. Its quite a mess in general but as the rewrite I should have a better way to load a readme file and its code blocks

daviwil (approved):

PR #3879, opened 2021-02-12
@autorest/common package

timotheeguerin:
Following on next step after pr #3848, this extract the merging and markdown/yaml parsing into its own base package.

It also remove the `composer` plugin which seems not to be used anymore and remove a bunch of merging code specific to that old plugin


packages/extensions/core/src/app.ts
```diff
@@ -572,7 +573,7 @@ async function resourceSchemaBatch(api: AutoRest): Promise<number> {
       subscribeMessages(instance, () => exitcode++);
 
       // set configuration for that item
-      instance.AddConfiguration(ShallowCopy(batchContext, "input-file"));
+      instance.AddConfiguration(omit(batchContext, "input-file"));
```
daviwil, line 576: What's omit do in this case?

daviwil (approved):

packages/extensions/core/src/app.ts
timotheeguerin: it creates a copy of the object minus the requested params(in this case ("input-file")

PR #3887, opened 2021-02-16
Feature: Add support for extension to require system dependencies to be installed with a certain version

timotheeguerin:
1. Generic dependency using 
```json
 "systemRequirements": {
    "foo": { }
  }
```
image

2. Require python version
```json
 "systemRequirements": {
    "python": {
        "version": ">=5.0"
    }
  }
```
image

Docs: https://github.com/Azure/autorest/blob/feature/system-requirements/docs/developer/writing-an-extension.md#extension-system-requirements


daviwil (approved):
This is fantastic, nice work!

PR #3888, opened 2021-02-17
Feature: Autorest cli migration to webpack

timotheeguerin:
Last package using `static-link`

Changes:
- Migration to webpack
- Updated some requires to be compatible with webpack
- Added a way for extension to call yarn cli from webpack bundle.

### DIFF
--- a/packages/apps/autorest/entrypoints/app.js
+++ b/packages/apps/autorest/entrypoints/app.js
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+/* eslint-disable no-console */
 
 global.isDebuggerEnabled =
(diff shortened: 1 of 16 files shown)


packages/apps/autorest/entrypoints/app.js
```diff
@@ -46,17 +47,7 @@ if (
       process.exit(1);
     }
 
-    if (v[0] > 14) {
```
daviwil, line 49: Should we still leave in this version check?

daviwil (approved):
Excellent work!  Glad to finally get rid of `static-link` 🙌🏻

packages/apps/autorest/entrypoints/app.js
timotheeguerin: hhm yeah, for some reason thought this was an outdated check

PR #3890, opened 2021-02-17
Internal: Publish pending changes in master as prerelease

timotheeguerin:
Automatically count the number of changes to compute the prerelease version.

Prerelease version is in this format `{next-version-by-rush}.dev-{change-count-for-this-package}`

This means if `1.2.3` is the current and the next version rush would get is `1.3.0` and there was 4 changes on that package the prerelease would be `1.3.0-dev.4`

Example of `@autorest/common`
image


eng/bump-for-prerelease.js
```diff
@@ -0,0 +1,107 @@
...
+    if (stat.isDirectory()) {
+      files.push(...(await findAllFiles(path)));
+    } else {
+      files.push(path);
+    }
+  }
+  return files;
+};
+
+const readJsonFile = async (filename) => {
```
daviwil, line 32: Can't define these as `async` functions?  Not important, just curious

daviwil (approved):
Looks good!  Looking forward to seeing how this works out

eng/bump-for-prerelease.js
timotheeguerin: not sure what you mean here, those are `async`

daviwil: I'm referring to the other way of defining the function without assigning a lambda:

```
async function readJsonFile(filename) {
```

timotheeguerin: oh, haha yeah I guess I am used to using const instead of the old function keywords. I think this is coming from doing a lot of react where this is the way to go as you often pipe function and they need the `const` way.
```js
const myComponent = observer(props => props.text);
```
so I just endedup using const all the time for consitency

daviwil: Makes sense!

timotheeguerin: now for non react, this isn't as needed so I guess could switch back to using `function`. It is a subject where I think nobody has a definite answer. THere is small techinical difference between both.

PR #3916, opened 2021-02-23
Fix: Don't automatically include Content-Type parameter if it is already defined in the spec as a parameter

### LINKED ISSUE #3914: Content-Type header shouldn't be synthesized when an explicit one exists.
If the operation already has a Content-Type parameter defined we shouldn't add another one.

### LINKED ISSUE #3917: Autorest logs debug information at Information level
```
INFORMATION: Options {"input":["prechecker"],"naming":{"override":{"cmyk":"CMYK","$host":"$host"}},"always-create-accept-parameter":true,"always-create-content-type-parameter":true,"flatten-models":true,"flatten-payloads":true,"group-parameters":true}
INFORMATION: Accept-param true
```

timotheeguerin:
fix #3914 
fix #3917

This resolve the issue where in `multipart/mixed` payload one could want to have a custom content-type provided by the consumer of the SDK(Setting the boundary)

Todo: 
- [x] Add tests

### DIFF
--- a/packages/extensions/modelerfour/src/modeler/utils.test.ts
+++ b/packages/extensions/modelerfour/src/modeler/utils.test.ts
@@ -1,0 +1,61 @@
+import { Operation, Parameter, Schema, SchemaType } from "@autorest/codemodel";
+import { isContentTypeParameterDefined } from "./utils";
+
+const stringSchema = new Schema("string", "A String", SchemaType.Any);
+
+describe("Modlerfour utils", () => {
+  describe("isContentTypeParameterDefined()", () => {
+    it("returns false if there is a no parameter in the operation", () => {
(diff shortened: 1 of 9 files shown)


packages/extensions/modelerfour/src/modeler/utils.test.ts
```diff
@@ -0,0 +1,61 @@
+import { Operation, Parameter, Schema, SchemaType } from "@autorest/codemodel";
+import { isContentTypeParameterDefined } from "./utils";
+
+const stringSchema = new Schema("string", "A String", SchemaType.Any);
+
+describe("Modlerfour utils", () => {
```
daviwil, line 6:
```suggestion
describe("Modelerfour utils", () => {
```

daviwil (approved):

PR #3932, opened 2021-02-24
Fix publishing preview version script appending change count to packages without changes.

timotheeguerin:
When computing the number of changes it wasn't excluding `type: none` which are noop changes which resulted in `rush publish --apply` not bumping the version and then the script to add an invalid 4th number(Having 4 number version without a suffix is invalid for npm`

Other issue was the dependencies where left with the version rush bumped to not the version with the appended change count which would have been invalid and prevented pacakge install(Not an issue for autorest, core and m4 that don't have npm dependencies)

### DIFF
--- a/eng/bump-for-prerelease.js
+++ b/eng/bump-for-prerelease.js
@@ -49,7 +49,8 @@ const getChangeCountPerPackage = async () => {
 
   for (const change of changes) {
     if (!(change.packageName in changeCounts)) {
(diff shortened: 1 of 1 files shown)


daviwil (approved):
Man, Rush really needs some behavior for this, I feel like we end up having to rewrite a lot of their versioning logic just for prereleases.

PR #3934, opened 2021-02-26
Bump versions and update changelogs for March 1st  release

timotheeguerin:
Main Package updates:
- `@autorest/core`: `3.0.6375` -> `3.1.0` This is the new configuration loading rewrite. Should not be a breaking change. 
- `autorest`: `3.1.1` -> `3.1.2`
- `@autorest/modelerfour`: `4.17.0` -> `4.17.1`


timotheeguerin:
@TheOnlyWei @dolauli This will release to latest the `input-file` bug fix and the powershell issue fix

Could you confirm using version `3.1.0-dev.4` with `--version:3.1.0-dev.4` everything is working as expected?

daviwil (approved):

dolauli:
@timotheeguerin

timotheeguerin:
@dolauli how about the published version? `3.1.0`? I think this was due to the configuration still forcing `~3.0.0` until I published

dolauli:
> @dolauli how about the published version? `3.1.0`? I think this was due to the configuration still forcing `~3.0.0` until I published
@timotheeguerin 
I just have a try. autorest/core@3.1.0 works. Thanks.

PR #3936, opened 2021-03-01
Feature: Move schema validator to ajv and ajv-errors for custom messages

### LINKED ISSUE #3366: Schema errors reported at the wrong location
```
ERROR: Schema violation: Additional properties not allowed: x-ms-client-name
    - file:///D:/github/azure/autorest.csharp/test/swaggers/extension-client-name.json:1:1
```

### LINKED ISSUE #3929: linter gives wrong message for path syntax error
**Describe the bug**
I was trying to generate a Swagger file that contained a syntax error. Specifically, the path did not start with `/`. However, instead, Autorest gave me the following error
```
ERROR: Schema violation: Additional properties not allowed: multivariate/models?$skip=0&$top=5 (x-ms-paths)
```
It took a long time to diagnose (thanks @lmazuel or I would probably still be pulling my hair out!). 

**Expected behavior**
It would be great if the linter called out the syntax error in the Swagger.

timotheeguerin:
fix #3929  (Better error message for invalid paths)
fix #3366 (Show the correct error line)

Using `ajv` and `ajv-errors` lets us extend the json schema with `errorMesssage` properties which allows for displaying more relevent information when certain error occur.

For example this used to be `Additional properties not allowed: multivariate/models?$skip=0&$top=5 (x-ms-paths)`
image

Other change include bumping schema to draft-07 which allows us to use `if` `then` instead of `oneOf` and give a much better error message. Specially in the parameter scenarios
Todo: 
- [x] Add tests


packages/extensions/core/src/lib/pipeline/plugins/schema-validation/swagger-schema-validator.ts
```diff
@@ -0,0 +1,22 @@
+import addFormats from "ajv-formats";
+import { JsonSchemaValidator } from "./json-schema-validator";
+
+export class SwaggerSchemaValidator extends JsonSchemaValidator {
+  public constructor() {
+    super();
+    addFormats(this.ajv);
+
+    this.ajv.addSchema(
+      require("C:/dev/azsdk/autorest/packages/libs/autorest-schemas/swagger.json"),
```
daviwil, line 10: Does this path (and the next one) need to be updated?

daviwil (approved):

packages/extensions/core/src/lib/pipeline/plugins/schema-validation/swagger-schema-validator.ts
timotheeguerin: haha completly missed that copied over from a sandbox script.

daviwil: No worries!

PR #3938, opened 2021-03-02
change trenton to terraform

dw511214992:


common/changes/@autorest/core/change-trenton-to-terraform_2021-03-02-07-03.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@autorest/core",
+      "comment": "change trenton to terraform",
+      "type": "minor"
```
timotheeguerin, line 6: would maybe just do a `patch` here instead of bumping the minor specially as this is not a public package yet.
```suggestion
      "type": "patch"
```

packages/extensions/core/resources/plugin-trenton.md
```diff
@@ -1,7 +1,7 @@
-# Default Configuration - trenton
```
timotheeguerin, line 1: This file needs to be renamed to `plugin-terraform.md`, seems to be the reason for the tests failing

timotheeguerin (approved):

PR #3950, opened 2021-03-04
Fix: Interpolating with value from the same block couldn't be overwritten

timotheeguerin:
When having for example

In the main readme
```yaml
branch: feature1
```

and this in a required readme
```yaml
branch: master
repo: https://github.com/Azure/autorest/$(branch)
```

you would end up with this config
```yaml
branch: feature1 # Branch picked the main readme value correctly
repo: https://github.com/Azure/autorest/master # But here it interpolated the local value instead.
```

This resolve the problem so the result is 
```yaml
branch: feature1 
repo: https://github.com/Azure/autorest/feature1
```

Added a bunch of tests regarding interpolation as well.


daviwil (approved):
Looks great!

Issue #3497, opened 2020-05-07
OpenAPI V3 - Cannot parse OpenAPI v3 definition file with schema containing 'additionalProperties' with boolean value


cermakp:
Hello,

I started to use the latest version of autorest (3.0.6187) which supports OpenApi v3. I have definition file generated from swagger using SwashBuckle library (using .net core application).

I am generating client with autorest using parameter `--v3`. However autorest returns following error
`FATAL: AutoRest.Core.Logging.CodeGenerationException: Error parsing swagger file. Error converting value False to type 'AutoRest.Modeler.Model.Schema'. Path 'components.schemas.BrandDto.additionalProperties', line 1, position 1439.`

![image](https://user-images.githubusercontent.com/49719500/81270542-2ec79480-904b-11ea-91e0-192571a4d955.png)

It seems that autorest does not support boolean values for `additionalProperties` as it is supported by OpenApi v3. **Anyway if I change the `additionalProperties : false` to `additionalProperties : {}` it starts to work.** 

Here is the sample json used:

```json
{
  "openapi": "3.0.1",
  "info": {
    "title": "Product Global Taxonomy Service API",
    "version": "v1"
  },
  "paths": {
    "/taxonomy/brands": {
      "get": {
        "summary": "Provides taxonomy brands",
        "operationId": "GetTaxonomyBrandsDeprecated",
        "responses": {
          "200": {
            "description": "Success",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/BrandCollection"
                }
              }
            }
          }
        },
        "deprecated": true
      }
    }
  },
  "components": {
    "schemas": {
      "BrandDto": {
        "type": "object",
        "properties": {
          "taxonomyBrandId": {
            "type": "integer",
            "description": "Surrogate Brand Id",
            "format": "int32"
          },
          "name": {
            "type": "string",
            "description": "Brand name",
            "nullable": true
          }
        },
        "additionalProperties": false
      },
      "BrandCollection": {
        "type": "object",
        "properties": {
          "collection": {
            "type": "array",
            "items": {
              "$ref": "#/components/schemas/BrandDto"
            },
            "nullable": true
          }
        },
        "additionalProperties": false
      }
    }
  }
}
```

kevinoid:
Looks like the same issue as #3439?

cermakp:
[...]

PR #4006, opened 2021-03-23
Fix: OpenAPI3 Validation of additional properties support boolean

timotheeguerin:
From  #3497

when converting to the if, then else system for better error, seems like the additionalProperties: boolean wasn't implemented correctly which prevented it from being used


ffMathy:
This is great! Thank you!

@timotheeguerin is there a way to test this out already? Has it already been released?

timotheeguerin:
Can you try with adding `--version:https://tinyurl.com/yzejgwb6` to your command line?

ffMathy:
Yeah that worked - thank you!

PR #4035, opened 2021-04-02
Added a remove-parameter directive

### LINKED ISSUE #3679: Consider adding a convenient way to remove the header.
Headers commonly defined by policies and not parameters are still defines in swagger and removing them from operation parameters seem to be pretty common.

timotheeguerin:
fix #3679

Also added documentation for all built-in directives.


pakrym (approved):
Nice!

PR #4077, opened 2021-04-15
Fix: x-ms-paths merged into path in oai2-to-oai3 converter

timotheeguerin:
Issue is that `x-ms-paths` are merged into `paths` immediately when converting from swagger to openapi 3. This is incorrect as using query param in path is still not allowed in openapi 3.
Move the merge to be latter in the pipeline.


packages/libs/oai2-to-oai3/test/resources/conversion/oai3/ApiManagementClient-openapi.json
```diff
@@ -5808,7 +5808,9 @@
         },
         "x-ms-requestBody-index": 4
       }
-    },
+    }
+  },
+  "x-ms-paths": {
     "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ApiManagement/service/{serviceName}/apis/{apiId}?export=true": {
```
pakrym, line 5814: worth adding a test that has a parameter defined in a query?

pakrym:
Works

packages/libs/oai2-to-oai3/test/resources/conversion/oai3/ApiManagementClient-openapi.json
timotheeguerin: for what this test is doing it shouldn't make a different between the content of the path. The semantic validator was never supposed to validate x-ms-paths

pakrym: Why not? SHould be fine for the path part.

timotheeguerin: well when I added the path validator it was for the openapi paths specs. I could add another one for the x-ms-paths in particular but i'll have first to double check the actual spec for that and what we should actually be validating

PR #4080, opened 2021-04-16
Prechecker duplicate schema performance improvements x40

### LINKED ISSUE #4076: prechecker deduplication is very slow
Generation times went up significantly when the deduplication was added.

We've even had to add a warning.
```
      "Checking for duplicate schemas, this could take a (long) while.  Run with --verbose for more detail.",
```

For example, when generating the https://raw.githubusercontent.com/Azure/azure-rest-api-specs/50a94ccd66f49aed78efa24ceefd0d5c1186c8db/specification/sql/resource-manager/readme.md the deduplication takes 90sec.

I think that's way too much considering we can pre-group schemas based on their names.

timotheeguerin:
Performance improvements of prechecker.

Basic benmark: `60s` to `1.5s` (x40 speed increase)

fix #4076

Todo: 
- [x] tests
- [x] docs on why we need to loop
- [x] Optmize resolving errors


pakrym:
Works and is very fast! Thank you!

PR #4086, opened 2021-04-19
Stop using items from linq to improve performance

timotheeguerin:
This improve merger performance significantly


timotheeguerin:
Unfortunately the way visit was designed make this hard to actually go through, I'll close this PR and make an alternative traverse functionally

PR #4100, opened 2021-04-26
Fix tree shaked number enum to follow same logic as string enums

timotheeguerin:


lmazuel (approved):
LGTM in intention (didn't look at the code :p)

PR #4150, opened 2021-05-27
Document x-ms-allowed-arm-id-types extension

matthchr:


matthchr:
See https://github.com/Azure/k8s-infra/pull/442 for the original discussion around this topic.

@johanste @m-nash @markcowl - would be great if you could take a look at this.

In terms of validating that Swagger authors actually set this:

1. I will send a PR to [azure-openapi-validator](https://github.com/Azure/azure-openapi-validator) adding a linter that uses a heuristic to detect things that look like references. Basically it'll find fields that are `id` or `*Id` (and possibly a few more variants) and look at the documentation to see if they are likely ARM IDs.
2. I'll work with Arthur Ning's team to use their https://aka.ms/LiveValidation pipeline to perform validation against live requests.

I _think_ between these two things we should be able to catch 90-100% of all references.

docs/extensions/readme.md
```diff
@@ -1173,6 +1174,112 @@ When set, specifies the header parameter to be used instead of `x-ms-client-requ
...
+
+### Allowed ARM ID formats
+Note: When reading the formats below, `{}` is left as-is in the `allowedTypes` entry. Named fields such as `{provider}` must be given concrete values in the `allowedTypes` entry.
+
+| Resource kind                   | Format                                                                      | Example `allowedTypes` entry                                                       |
+| ------------------------------- | --------------------------------------------------------------------------- | ---------------------------------------------------------------------------------- |
+| Resource in a tenant            | `/providers/{provider}/{resourceType}`                                      | `/providers/Microsoft.Capacity/reservationOrders`                                  |
+| Resource in a subscription      | `/subscriptions/{}/providers/{provider}/{resourceType}`                     | `/subscriptions/{}/providers/Microsoft.Resources/resourceGroups`                   |
```
matthchr, line 1196: I originally wanted to use the same format that ARM specifies in ARM templates (`Microsoft.Resources/resourceGroups` or `Microsoft.Compute/virtualMachines`, etc), but that format doesn't actually convey if the resource roots up to subscription, resourceGroup, or something else, which is required to validate resource ID validity (or generate an ID that mathces) so I think we need a more verbose format such as this one.

docs/extensions/readme.md
```diff
@@ -1173,6 +1174,112 @@ When set, specifies the header parameter to be used instead of `x-ms-client-requ
...
+
+**Example**: An `arm-id` field that must refer to a virtual network
+```json5
+"MyExampleType": {
+  "properties": {
+    "vnetId": {
+      "type": "string",
+      "format": "arm-id",
+      "x-ms-allowed-arm-id-types": {
+        "allowedTypes": [
```
timotheeguerin, line 1222: So we need this nesting for this new field?

As `x-ms-allowed-arm-id-types` and `allowedTypes` seems a bit of a duplication of names.

matthchr: Yeah agree it's a bit duplicate.

I made it a complex type rather than a raw array just because it felt more extensible if in the future we wanted like a `deniedTypes` or something too (i.e. all but this?).

If folks are confident on the shape I've proposed here I'm happy to collapse this to just a simple array

matthchr: Maybe we could change it to `x-ms-arm-id-types` with an `allowed` field?

matthchr: Changed it to `x-ms-arm-id-details`

matthchr:
@akning-ms as well

docs/extensions/readme.md
```diff
@@ -1173,6 +1174,113 @@ When set, specifies the header parameter to be used instead of `x-ms-client-requ
...
+```
+
+**Example**: An `arm-id` field that must refer to a virtual network
+```json5
+"MyExampleType": {
+  "properties": {
+    "vnetId": {
+      "type": "string",
+      "format": "arm-id",
+      "x-ms-arm-id-details": {
```
ruowan, line 1222: Is `x-ms-arm-id-details` a required property when the format is `arm-id`

docs/extensions/readme.md
```diff
@@ -1173,6 +1174,113 @@ When set, specifies the header parameter to be used instead of `x-ms-client-requ
...
+  }
+}
+```
+
+**Example (preferred)**: An `arm-id` field with no extension, referring to the common type.
+```json5
+"MyExampleType": {
+  "properties": {
+    "id": {
+      "$ref": "../../../../../common-types/resource-management/v2/types.json#/definitions/ArmId",
```
jianyexi, line 1237: what's the common type 'ArmId' for ?

docs/extensions/readme.md
PhoenixHe-msft: One id property MAY want to point to specified resource type but don't want to specify scope. For example, some property may want to point to any authorization resource. How do you think about this?

docs/extensions/readme.md
matthchr: @johanste mentioned that he thought it would be useful to have a shared type in the common defs (like we do for `Resource` and a few other types today).

docs/extensions/readme.md
matthchr: No, it's optional. Here's the snippet from the description up above:

> When set, specifies the set of resource types which can be referenced by this `arm-id`. If this extension isn't provided for a particular `arm-id`, the field can refer to any valid ARM ID.

Also, the first example shows an `arm-id` field with no extension as well.

docs/extensions/readme.md
matthchr: @PhoenixHe-msft - what does an authorization resource URL look like?

johanste: I assume that the general question is what referring to an [extension resource](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/extension-resource-types) looks like... In an ARM template, this would be done by providing the [scope](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/scope-extension-resources) for the resource.

I'm not sure how deep down the rabbit hole we'd like to run with specifying that an ARM ID can only refer to extension resource type A but only if it is scoped to resource type B. I believe that could be a recursive definition that would make the hole very deep indeed. But I could see a `scope: *` or similar added to indicate that this is an extension provider.

PhoenixHe-msft: Exactly. Suggest to referencing the design of ARM template's scope.

matthchr: From my read of the scope documentation (Thanks @johanste) it looks like scope is inferred in a large number of cases based on the other fields specified.

This seems to be confirmed by the definition of the `scope` field in the JSON Schema:
```
            "scope": {
              "type": "string",
              "description": "Scope for the resource or deployment. Today, this works for two cases: 1) setting the scope for extension resources 2) deploying resources to the tenant scope in non-tenant scope deployments"
            },
```

So that means that the valid values of `scope` are:
1. `/`
2. Something like `Microsoft.Storage/storageAccounts/myStorageAccountName` in the case of extension resources.

This list of values doesn't provide us the level of specification that we need, because it's not actually saying if it's a RG or Subscription (or region) scoped deployment. They instead infer that detail from the other template fields (which we don't have here).

So we would need to augment for our usage of scope (if we're going to use it that is).

Are you suggesting that we change this to something like the below?

```json5
"MyExampleType": {
  "properties": {
    "vnetId": {
      "type": "string",
      "format": "arm-id",
      "x-ms-arm-id-details": {
        "allowedTypes": [
          {
            "scope": "/subscriptions/resourceGroups",
            "resource": "Microsoft.Network/virtualNetworks"
          }
        ]
      }
    }
  }
}
```

We'd have to discuss what values we would allow for `scope` because as mentioned above the set that ARM supports isn't sufficient for our needs. If people do prefer this approach though we could have:

1. `/` - Tenant
2. `/subscriptions` - Subscription
3. `/subscriptions/resourceGroups` - ResourceGroup
4. `*` - For extension resources

matthchr: Or are you saying a single field string and just make it "look like a scope" on the left hand side:
`/subscriptions/providers/Microsoft.Resources/resourceGroups` - everything to the left hand side of `providers` is the scope.

matthchr: Sorry for the delay -- I've pushed an update to this PR that moved towards the recommended `scope` style.

I went with the shorter "all in a single string" format over the longer `"scope"` + `"resource"` properties above as it felt cleaner and parsing the single string to determine the two pieces (scope and resource) should be easy.

Let me know what you think @johanste and @PhoenixHe-msft

PhoenixHe-msft (approved):

akning-ms:
+ @leni-msft to comment

johanste (approved):

johanste:
@markcowl may also have some valuable feedback on this proposal. Explicitly mentioning him (and adding him as a reviewer) probably gets his attention!

docs/extensions/readme.md
leni-msft: @majastrz for comments, who may have valuable comments for the way of specifying resource types and scopes, as he is working on another proposal `x-ms-resource-scope`.

matthchr: @meni-msft great catch. I think the key differences in @majastrz proposal and this one are:

1. It supports one more scope (management groups) at `/providers/Microsoft.Management/managementGroups`. I believe I should also support this.
2. It proposes two possible scope formats that are somewhat different than the one we've landed on here. We should unify on a format.
3. It supports `AnyResource` and `AnyResourceExtension` in addition to `*`. We should decide if we want/need `AnyResource`/`AnyResourceExtension` here as well. @majastrz pointed out that we could also add these later if we just started with `*` so this isn't a hard blocker I don't think.

From this list I think the key question is on item 2 above, how the `scope` section should be specified.

timotheeguerin:
Is there anybody else that should still look at this PR or should we merge it?

matthchr:
@timotheeguerin - @leni-msft linked me to a work document from the ARM Bicep team that has a related proposal and we wanted to ensure we were using a similar `scope` for each proposal. I haven't finished doing that yet though (I've been busy on other stuff). I'll try to get that done later this week or probably more likely early next week.

It would also be great if @markcowl gets a chance to take a look probably? I am not sure what other stakeholders there are and if we've gotten their approval...

matthchr:
I wanted to unify the experience with the document being driven by @majastrz, but it looks like there hasn't been movement there. I would still like to get this merged but maybe the thing to do is just take it basically as it is, I can work with @leni-msft and team to help add detectors to find Swaggers that are missing (via the Swagger completeness workstream that flows to S360 probably?)

@leni-msft is there a wiki or something on how to add rules to detect things for stuff like this? What's the process there. I'm basically happy to merge this assuming people are still OK with it but a merged doc doesn't do much to move the needle on actually getting teams to adopt this. I can actually go annotate some major services Swaggers myself and send PRs for that and it would be great to get some linting as well if I can get help on how to do it...

leni-msft:
@matthchr To add new rules, please submit form via https://aka.ms/lintrulecandidate , and our team will review and discuss how to implement the rule in PR static validation and live validation.

docs/extensions/readme.md
leni-msft: @matthchr For item 2, to unify the scope format with @majastrz 's proposal, shall we change the extension like below?
```
"MyExampleType": {
  "properties": {
    "id": {
      "type": "string",
      "format": "arm-id",
      "x-ms-arm-id-details": {
        "allowedResources": [
          {
            "scopes": ["Tenant", "Subscription", "ResourceGroup"],
            "type": "Microsoft.Network/virtualNetworks"
          },
          {
            "scopes": ["Tenant", "Subscription", "ResourceGroup"],
            "type": "Microsoft.Compute/disks"
          },
          {
            "type": "Microsoft.Compute/virtualMachines"
          }
        ]
      }
    }
  }
}
```
The possible values of `scopes` are [ `Tenant`, `Subscription`, `ResourceGroup`, `ManagementGroup`, `*` ], and make `scopes` optional and default to `["ResourceGroup"]`.

So the changes include:
* `allowedTypes` -> `allowedResources`
* add `scopes`
* `resource` -> `type`

Since this is Azure-specific swagger extension, it should be fair to use these and `resourceType`.

matthchr: - `allowedTypes` -> `allowedResources`: I'm happy to make this change.
- add `scopes`: Agreed
- `resource` -> `type`: The other document calls it `paths` (or `path` for singular here), and it uses the full URL like `/subscriptions/{subscriptionId}/resourceGroups/{rg}/providers/Microsoft.Compute/virtualMachines/{vm}` whereas your example us just using `Microsoft.Compute/virtualMachines`. Just wanted to highlight that difference to the proposal that @majastrz made. I do think what you're suggesting here makes a lot of sense though as it doesn't make sense to put the full path here if it can possibly apply to many scopes (since the full path would be different for each scope).

I'll go ahead and make these changes because they seem good to me.

matthchr: Oh one more thing, I think `type` probably should support `*` too. That way you can do something like:
```
// This can be any ARM ID  
"id": {
  "type": "string",
  "format": "arm-id",
}
```

```
// This can be any ARM ID as long as it's RG scoped
"id": {
  "type": "string",
  "format": "arm-id",
  "x-ms-arm-id-details": {
    "allowedResources": [
      {
        "scopes": ["ResourceGroup"],
        "type": "*"
      },
    ]
}
```

I don't actually know if we have any resources that need this functionality today but I don't know that we don't so we might as well support it?

matthchr: One other thought... I think we should drop `*` and replace it with `Extension`.

`*`'s meaning is ambiguous right now. Is `"scope": "*"` == `"scope": ["Tenant", "Subscription", "ResourceGroup", "ManagementGroup"]`? Not really, because `*` also supports extension resources. It's confusing if `"scope": "*"` just means "this resource is an extension resource" though - if we want it to mean that we should call it `"scope": "Extension"` or `"scope": "AnyResource"` instead.

`*` meaning "All scopes" doesn't seem that useful because `scopes` is already a collection (as proposed by @leni-msft), and it seems unlikely to me that if Azure comes up with a new scope that all the old resources that said `"scope": "*"` with actually jump to immediately supporting references at the new scope. It seems better to just be explicit and make resources that refer to something like `Microsoft.Resources/deployments` say:
```
"deploymentId": {
  "type": "string",
  "format": "arm-id",
  "x-ms-arm-id-details": {
    "allowedResources": [
      {
        "scopes": ["Tenant", "Subscription", "ResourceGroup", "ManagementGroup"],
        "type": "Microsoft.Resources/deployments"
      },
    ]
}
```

Thoughts?

leni-msft: Thanks @matthchr for the quick updates.

> * `resource` -> `type`: The other document calls it `paths` (or `path` for singular here) ... 

No problem, the change has also been agreed by @majastrz in the doc comment.

> I think type probably should support `*` too

Agree. One further step, the type should support `Microsoft.Compute/*` - to limit the resource providers but allow any resource types under it.

> I think we should drop `*` and replace it with `Extension`

Agree to add `Extension`. But I suggest to keep `*`(any valid scope) just for easier writing - it's unlikely that Azure will have a new scope.

docs/extensions/readme.md
```diff
@@ -1192,6 +1193,143 @@ When set, specifies the header parameter to be used instead of `x-ms-client-requ
...
+| `Subscription`    | `/subscriptions`                         | The resource is deployed into a subscription                                                                                                                                       |
+| `ResourceGroup`   | `/subscriptions/resourceGroups`          | The resource is deployed into a resource group                                                                                                                                     |
+| `ManagementGroup` | `/Microsoft.Management/managementGroups` | The resource is deployed into a management group                                                                                                                                   |
+| `Extension`       | Any                                      | The resource is an extension resource and may be deployed as a subresource of another resource in any of the above scopes.                                                         |
```
leni-msft, line 1225: Do you intend to define the actual URL prefix? if so, they should be like following:

> Tenant
> `/`
> 
> Subscription
> `/subscriptions/{subscriptionId}/`
> 
> Resource Group
> `/subscriptions/{subscriptionId}/resourceGroups/{groupName}/`
> 
> Management Group
> `/providers/Microsoft.Management/managementGroups/{managementGroupName}/`
> 
> Extension
> `{parentScope}/providers/{extensionNamespace}/{extensionType}/{extensionName}/`
> 
>

matthchr: Fixed

docs/extensions/readme.md
matthchr: > Agree. One further step, the type should support Microsoft.Compute/* - to limit the resource providers but allow any resource types under it.

Done.

> Agree to add Extension as a valid scope value. But I suggest to keep *(any valid scope) just for easier writing - it's unlikely that Azure will have a new scope.

Ok, as long as it's not likely for Azure to be adding new scopes I'm fine with `*`. Have added it back and documented it.

leni-msft (approved):

docs/extensions/readme.md
```diff
@@ -1192,6 +1193,145 @@ When set, specifies the header parameter to be used instead of `x-ms-client-requ
...
+
+**AllowedResource schema**:
+
+| Field Name  | Type       | Description                                                                                                                                                                                                         |
+| ----------- | ---------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| scopes      | `[string]` | An array of scopes. See [Allowed Scopes](#allowed-scopes). If not specified, the default scope is `["ResourceGroup"]`.                                                                                              |
+| type        | `string`   | **Required** The type of resource that is being referred to. For example `Microsoft.Network/virtualNetworks` or `Microsoft.Network/virtualNetworks/subnets`. See [Example Types](#example-types) for more examples. |
+
+#### Allowed Scopes
```
markcowl, line 1217: What are the thoughts on how scopes should be interpreted?  Does this apply a restriction to the same tenant / subscription / resource group / management group as the referring resource, or something else?

I worry about making this too deployment-specific.  It would be nice to apply this in general to encode actual 'locality' restrictions that could also be used both for smart tools to validate an input reference or for for smart tools to deploy the resource.  One such restriction would be 'location', for resources that must be in the same region as the referring resource.

markcowl (changes requested):

docs/extensions/readme.md
markcowl: So, management groups typically sit 'above' subscriptions, so, unless I am wildly mistaken, you can *only* deploy a virtual network to a resource group.  Management Groups are a good deployment location for extension resources, but don't apply to resources that are Tracked Resources or Subscription resources, or resource group resources.

I guess here, you want to be able to deploy the resource without having to know anything about the properties of the resource type, is that correct?

matthchr: > I guess here, you want to be able to deploy the resource without having to know anything about the properties of the resource type, is that correct?

Yes, exactly. Maybe I need to do some documentation improvements (or we need to pick a different name than `Scope`?). As discussed over teams the issue is that this is a reference to a say `Microsoft.Network/virtualnetworks`, and _at the reference site_ there's no knowledge of what that thing looks like. The objective for this annotation would be that client tools could give better hints/warnings about ARM references, so ideally given this annotation we can look at an ARM ID and confirm if it's valid. Without a `scope` section though we couldn't confirm that:
`/subscriptions/<id>/Microsoft.Network/virtualNetworks/myVnet` was an invalid reference, because it's not at the right scope.

markcowl: I would say, this is good for the first version, and perhaps we can augment this later with some restrictions around the resource (for example, that it has to be in the same region,

markcowl (approved):

PR #4152, opened 2021-06-02
Allow x-ms-paths not to start with /

timotheeguerin:


kasobol-msft (approved):
Works pretty well here https://github.com/Azure/azure-sdk-for-net/pull/21424

PR #4214, opened 2021-07-15
Feature: Telemetry

### LINKED ISSUE #4081: Telemetry for autorest
The same way we have telemetry of most of our products like CLI, the idea would be to have an appinsights dedicated to autorest that will collect the set of parameters people are inputting to autorest, so we can get an idea of the usage of it.

timotheeguerin:
fix #4081 

- [ ] Check doc describe all telemetry
- [ ] Replace AI prod Key 
- [ ] Add Privacy notice


NowRNever2234:
Get the fuck off my phone bitch I'm seeing it now￼

PR #4257, opened 2021-08-11
Fix: Dictionaries don't get xml serialization property added in codemodel

### LINKED ISSUE #4256: xmlName is not sent to the generator from Autorest
The Storage Team has reported an issue with TS Generator. You can find the issue at https://github.com/Azure/autorest.typescript/issues/1141. 

In the Storage Blob Data Plane SDK - [file](https://github.com/Azure/azure-rest-api-specs/blob/storage-dataplane-preview/specification/storage/data-plane/Microsoft.BlobStorage/preview/2020-10-02/blob.json), we have the following swagger definition:

```
"ObjectReplicationMetadata": {
      "type": "object",
      "xml": {
        "name": "OrMetadata"
      },
      "additionalProperties": {
        "type": "string"
      }
}
```

With the above definition, the generated SDK looks like:

```
objectReplicationMetadata: {
        serializedName: "ObjectReplicationMetadata",
        xmlName: "ObjectReplicationMetadata",
        type: {
          name: "Dictionary",
          value: { type: { name: "String" } }
        }
},
```

As you can see with the above code, the `xmlName` must be `OrMetadata`. But, the name is incorrect. 

But, the `xmlName` property is not sent to the generator in the code model. This has to be fixed.

@timotheeguerin FYI

timotheeguerin:
fix #4256

Todo: 
- [ ] add tests

### DIFF
--- a/common/changes/@autorest/modelerfour/fix-dict-serialization_2021-08-11-19-31.json
+++ b/common/changes/@autorest/modelerfour/fix-dict-serialization_2021-08-11-19-31.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@autorest/modelerfour",
+      "comment": "**Fix** Dictionaries don't get xml serialization information passed into codemodel",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@autorest/modelerfour",
+  "email": "tiguerin@microsoft.com"
+}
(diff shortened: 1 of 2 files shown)


timotheeguerin:
@sarangan12 can you verify this fixes it

sarangan12:
@timotheeguerin Thanks for the quick turn around. I have verified the fix. The `xmlName` is getting generated correctly. 

Kindly let me know when this can be shipped. Also, to start using the latest versions, we should just use `autorest --reset`. Am I correct?

timotheeguerin:
The version will be published as preview as soon as the main build completes and should be as version `4.21.0-dev.3`
I most likely not publish a new version for a while. I am going to be off the next 2 weeks and current `main` branch has a lot of new changes which could potentially need hotfixes. You can keep using the preview version however.

PR #4284, opened 2021-09-09
Docker publishing base

timotheeguerin:
fix #4284


jongio:
Thanks @timotheeguerin  - When should we expect this change to propagate to : https://hub.docker.com/r/azuresdk/autorest

timotheeguerin:
Hey @jongio, all images described in this doc here should be published https://github.com/Azure/autorest/blob/main/docs/docker.md

Please file another issue if there is some problems.

PR #4287, opened 2021-09-13
use autorest.java main branch in docs

weidongxu-microsoft:
autorest.java recently switched to "main" branch (from "v4" branch).


timotheeguerin (approved):
Thanks for doing this!

PR #4343, opened 2021-10-19
Fix: Nullable dropped on certain properties

timotheeguerin:
if the properties was tree shaked then it would drop the nullable attribute


aelij:
@timotheeguerin I can confirm the fix works for my scenario. Thanks!

PR #4357, opened 2021-11-10
Oai2 oai3 parameter extensions flexible

timotheeguerin:


packages/libs/oai2-to-oai3/src/converter.ts
```diff
@@ -11,7 +11,15 @@ import {
   createMappingTree,
 } from "@azure-tools/datastore";
 import { JsonPointer, getFromJsonPointer, appendJsonPointer } from "@azure-tools/json";
-import oai3, { EncodingStyle, HttpOperation, JsonType, PathItem, SecurityType } from "@azure-tools/openapi";
+import oai3, {
+  EncodingStyle,
+  HttpOperation,
+  includeXDashKeys,
+  JsonType,
+  omitXDashProperties,
```
AlexanderSher, line 19: Do you use `omitXDashProperties` anywhere?

AlexanderSher:
Solves the problem with `x-ms-` attributes

PR #4382, opened 2021-12-01
Fix: Loading configuration nested under the same as a flag 

timotheeguerin:
Plugins that depends on nested configuration under the same key as the flag wouldn't get merged.

### DIFF
--- a/common/changes/@autorest/common/fix-merging-config_2021-12-01-16-29.json
+++ b/common/changes/@autorest/common/fix-merging-config_2021-12-01-16-29.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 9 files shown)


niander:
Has this change been released? I am still facing the same issue when using latest @autorest/core

timotheeguerin:
Hi, @niander this  change should have been released. Can you share what config and cli args you are using?

niander:
@timotheeguerin I did `autorest --reset` and tried again, which seems to have fixed my issue. Nonetheless, answering your question, I was using `autorest --az` and it wasn't reading from the `readme.az.md` the configuration `az.extensions`. It is working now. Thanks

PR #4383, opened 2021-12-06
Fix: flattened properties causing duplicates

timotheeguerin:
Change caused by #4378 to resolve duplicate issues when deduplicating consecutive words in properties.
This remove the deduplication of properties with `AutoGenerated` which was not supported before anyway and shouldn't be.


AlexanderSher:
I can confirm that it fixed the issue:
image

Issue #4390, opened 2021-12-21
x-ms-identifiers not allowed when generating ARM template schemas


leni-msft:
**Before filling a bug**

- [x] have you checked the [faq](https://github.com/Azure/autorest/tree/main/docs/faq.md) for known issues.
- [x] have you checked existing issues

**Describe the bug**

Please provide as much information as you can. This would be:
- OpenAPI files having the issues: https://github.com/leni-msft/azure-rest-api-specs/blob/d76689fc4d/specification/oep/resource-manager/readme.md
- Autorest command used:
```
➜  azure-rest-api-specs git:(pr-17105) autorest --version=3.0.6375 --azureresourceschema --multiapi ~/workspace/azure-rest-api-specs/specification/oep/resource-manager/readme.md --debug
SourceMapConsumer.initialize is a no-op when running in node.js
AutoRest code generation utility [cli version: 3.5.1; node: v14.15.5]
(C) 2018 Microsoft Corporation.
https://aka.ms/autorest
info    | AutoRest core version selected from configuration: 3.0.6375.
debug   | [0.67 s] Network Enabled: true
debug   | [0.74 s] The most recent installed version is 3.7.3
debug   | [0.74 s] Starting @autorest/core from /home/leni/.autorest/@autorestcore@3.0.6375/nodemodules/@autorest/core
   Loading AutoRest core      '/home/leni/.autorest/@autorestcore@3.0.6375/nodemodules/@autorest/core/dist' (3.0.6375)
   Loading AutoRest extension '@autorest/azureresourceschema' (~3.0.45->3.0.103)
Processing batch task - {"tag":"schema-2021-06-01-preview"} .
DEBUG [1.51 s]: swagger-document/loader-swagger - START inputs = 0
DEBUG [1.51 s]: openapi-document/loader-openapi - START inputs = 0
DEBUG [1.51 s]: configuration-emitter - START inputs = 0
DEBUG [1.51 s]: pipeline-emitter - START inputs = 0
DEBUG [1.55 s]: configuration-emitter - END [0.03 s]
DEBUG [1.55 s]: pipeline-emitter - END [0.03 s]
DEBUG [1.59 s]: openapi-document/loader-openapi - END [0.07 s]
DEBUG [1.59 s]: openapi-document/individual/transform - SKIPPING
DEBUG [1.59 s]: openapi-document/individual/schema-validator-openapi - SKIPPING
DEBUG [1.59 s]: openapi-document/individual/identity - SKIPPING
[...]

PR #4391, opened 2021-12-21
define x-ms-identifiers in swagger-extensions.json

leni-msft:
fixes https://github.com/Azure/autorest/issues/4390


leni-msft:
@timotheeguerin @daviwil would you be able to review this? And schemas generation is using `--version=v3`, is `main` the correct branch to fix?

leni-msft:
hotfix this issue by using `3.5.1` autorest/core

PR #4431, opened 2022-02-18
Fabric Bot Configuration

jsquire:
The focus of these changes is to add the Fabric Bot rules into the repository.


timotheeguerin (approved):
Does this means that this config will now control the fabric bot automation?

jsquire:
> Does this means that this config will now control the fabric bot automation?

It does.   All rule changes are now able to be tracked and history is available.   😄

PR #4447, opened 2022-03-03
python doc: add --default-optional-constants-to-none

changlong-liu:
For issue https://github.com/Azure/autorest.python/issues/893, and should be merge after https://github.com/Azure/autorest.python/pull/1171


iscai-msft (approved):
looks great!

PR #4486, opened 2022-03-22
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/packages/extensions/core/src/lib/pipeline/plugin-endpoint.ts
+++ b/packages/extensions/core/src/lib/pipeline/plugin-endpoint.ts
@@ -366,7 +366,7 @@ export class AutoRestExtension extends EventEmitter {
         const t = context.config.outputFolderUri.length;
         return (
           await context.fileSystem.list(ensureIsFolderUri(`${context.config.outputFolderUri}${artifactType || ""}`))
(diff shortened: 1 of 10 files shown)


timotheeguerin:
hey @CommanderRoot thanks for the contribution, looks like the CI is failing on some formatting, `npx prettier --write .` should run it or `rush format` if you installed the dependencies.

CommanderRoot:
`npx prettier --write .` and `rush format` generate different outputs, at least in the file https://github.com/Azure/autorest/compare/134196877e11dac7633687a6586d22a3c764d0d7..a6a3bd676629e7f13329f033f4e1428bb30f2de0
Lets hope it works now

timotheeguerin:
hhm interesting, might be an change in prettier version , will take a look.

PR #4490, opened 2022-03-28
Clarify method to get nextLink URL for x-ms-pageable

### LINKED ISSUE #4454: Support POST call to get next page for Paging API
**Is your feature request related to a problem? Please describe.**
When working on data plane codegen for Metrics Adviser, there is a new pattern of use POST call to get next page.
However, according to https://azure.github.io/autorest/extensions/#x-ms-pageable,  the request to get next page is always a 'GET' call.

**Describe the solution you'd like**
Extend the definition in x-ms-pageable to support both GET and POST.

**Describe alternatives you've considered**
Need to write complex directive to workaround this problem which i don't think this is a good solution.

**Additional context**
N/A

mikekistler:
Clarifies how the `operationName` in `x-ms-pageable` can be used to specify that request to nextLink URL should use `POST`.

Fixes #4454

### DIFF
--- a/docs/extensions/readme.md
+++ b/docs/extensions/readme.md
@@ -926,7 +926,7 @@ With this extension, you can set a default value for a property or parameter tha
 
 The REST API guidelines define a common pattern for paging through lists of data. The operation response is modeled in OpenAPI as a list of items (a "page") and a link to the next page, effectively resembling a singly linked list. Tag the operation as `x-ms-pageable` and the generated code will include methods for navigating between pages.
 
-**Note:** The request to get the nextPage will always be a `GET` request
+**Note:** The request to the nextLink URL will be a `GET` request unless `operationName` is specified. If `operationName` is specified, the request will use the HTTP method for that operation.
 
 **Schema**:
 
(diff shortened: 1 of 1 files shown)


timotheeguerin (approved):
Just need to fix the formatting. `rush format`

mikekistler:
I'm not certain there's a problem with the formatting.  It looks to me like a failure in the CI.
```
Starting "rush check-format"


> /Users/mikekistler/Projects/Azure/autorest/packages/extensions/core/node_modules/.bin/prettier --list-different --config .prettierrc.yml --ignore-path .prettierignore **/*.{ts,js,cjs,mjs,json,yml,yaml,cadl,md}
/Users/mikekistler/Projects/Azure/autorest/eng/scripts/check-format.js:5
  if (err instanceof CommandFailedError) {
                     ^

ReferenceError: CommandFailedError is not defined
    at Object.<anonymous> (/Users/mikekistler/Projects/Azure/autorest/eng/scripts/check-format.js:5:22)
    at Module._compile (node:internal/modules/cjs/loader:1103:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1157:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:77:12)
    at node:internal/main/run_main_module:17:47

The script failed with exit code 1
```

timotheeguerin:
Yeah that's a bug I fixed in this PR https://github.com/Azure/autorest/pull/4487 but the formatting issue is still there. It crash with this error when trying to give a a good error message that the formatting is incorrect.

mikekistler:
I think I have fixed the formatting issues.

PR #4504, opened 2022-04-06
Update rush and pnpm

timotheeguerin:
Using newer verson of PNPM should allow using VS2022 to build the native dependencies


DeagleGross:
Worked for me:
```
Linking finished successfully. (1.95 seconds)
Next you should probably run "rush build" or "rush rebuild"
Rush update finished successfully. (47.67 seconds)
```

PR #4550, opened 2022-06-02
Make Go v4 code generator the default

jhendrixMSFT:
Use v2 only when --legacy is specified.


timotheeguerin (approved):

timotheeguerin:
@jhendrixMSFT can you run `rush change` to create the changelog for this. Should be a `minor` change

PR #4572, opened 2022-06-24
config, fix extension java, update min codegen version to 4.1.0

weidongxu-microsoft:
Set default codegen of Java to DPG GA 1.0 at https://github.com/Azure/autorest.java/releases/tag/v4.1.0


packages/libs/configuration/resources/plugin-java.md
```diff
@@ -6,7 +6,7 @@ The V3 version of the Java Generator.
 version: ^3.7.0
```
weidongxu-microsoft, line 6: Should we set this at `^3.8.4` as this is the recommended `@autorest/core` version?

PR #4577, opened 2022-06-24
Update configuration to remove reliance on docker.io

scbedd:
@kashifkhan @iscai-msft 

Here are the changes. I will work through the build changes that are necessary as well.

### DIFF
--- a/docker/all/ubuntu/Dockerfile
+++ b/docker/all/ubuntu/Dockerfile
@@ -1,4 +1,4 @@
-FROM azuresdk/autorest
+FROM azsdkengsys.azurecr.io/azuresdk/autorest
 
 # DOTNET
 ENV DOTNET_VERSION=5.0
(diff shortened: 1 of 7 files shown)


scbedd:
@iscai-msft I will have an update to your publishing coming sooner rather than later. I will probably end up _duplicating_ your release code and then disabling the docker one via variable.

benbp (approved):

scbedd:
Unfortunately it looks like the docker publishing is already in a bad state. I am going to disable the failing stage, and allow the owners to follow-up on the deployment stages and decisions as to how they want the builds to work. The goal with this PR is AVOID having the repo crash across the board for all builds just for docker.io references.

scbedd:
The failing link check is also not something I've added. Going to ignore that change with check enforcer override and merge.

PR #4592, opened 2022-07-07
Avoid pointless warning

softworkz:
The warning is pointless for two reasons:

1. Using type=string and format=binary is the correct way to specify blob data in openapi
2. When type=string and format=binary, the block is effectively a no-op


timotheeguerin (commented):
Good catch @softworkz, thanks for fixing this. Could you run `rush format` as this is making the CI fail right now.

Issue #4602, opened 2022-07-26
remove-operation{-match} directive removes non-matching operations


andrewsomething:
**Before filling a bug**

- [x] have you checked the [faq](https://github.com/Azure/autorest/tree/main/docs/faq.md) for known issues.
- [x] have you checked existing issues

**Describe the bug**

According to the documentation, the `remove-operation` built-in directive can accept a regex:

```yaml
directive:
  - remove-operation: <operationId>  | <regex>
```
https://github.com/Azure/autorest/blob/66c3f70a8ad9ba1c5c8fdedd0378c921c06a32ce/docs/generate/built-in-directives.md#remove-operation

This does not seem to be true. After some digging, `remove-operation` does not  accept a regex, but `remove-operation-match` does:

https://github.com/Azure/autorest/blob/main/packages/libs/configuration/resources/directives.md#removal

Unfortunately, that does not work as expected.  `remove-operation-match` seems to be over matching. Generating a client using it removes all operations, not just ones matching the regex. 

**Please provide as much information as you can. This would be:**
- OpenAPI files having the issues: https://github.com/Azure/autorest/blob/d021dadda293256d3bf18f6c057ba474aaab37af/packages/extensions/core/test/resources/tiny.yaml
- Autorest command used: `autorest --python --input-file=https://raw.githubusercontent.com/Azure/autorest/d021dadda293256d3bf18f6c057ba474aaab37af/packages/extensions/core/test/resources/tiny.yaml --output-folder=generated/  --directive="remove-operation-match: /^cowbell_g.*/i"`

**Expected behavior**

A client would be generated with a `cowbell.add()` method but not a `cowbell.get()`. Instead, no operations are generated at all.

**Additional context**

The following integration test demonstrates the issue. `/^cowbell_g.*/i` should only match and remove `Cowbell_Get`, but the test fails as both operations are undefined.

```diff
diff --git a/packages/extensions/core/test/modifiers.e2e.ts b/packages/extensions/core/test/modifiers.e2e.ts
index 357205098..865724720 100644
--- a/packages/extensions/core/test/modifiers.e2e.ts
+++ b/packages/extensions/core/test/modifiers.e2e.ts
@@ -97,6 +97,17 @@ describe("Modifiers", () => {
     expect(findOperation(code, "Cowbell_Get")).toBe(undefined);
   });
 
+  it("remove an operation using a regex", async () => {
+    const code = await generate({
+      directive: {
[...]

PR #4604, opened 2022-07-27
Document where-operation-match directive.

andrewsomething:
As a follow up to https://github.com/Azure/autorest/issues/4602, this adds `remove-operation-match` to the built-in directives documentation. It also fixes a small formatting issue on that page.


timotheeguerin (approved):
Thanks for this @andrewsomething

PR #4647, opened 2022-09-30
Support final-state-schema LRO option

### LINKED ISSUE #4646: Add final-result-schema property to lro-options extension
**Is your feature request related to a problem? Please describe.**

Autorest currently uses the schema of the "200" response as the "final result schema" of LRO operations, but this requires services to document a 200 response even when the operation never actually returns a 200 -- only 202.

**Describe the solution you'd like**

We should add a `final-result-schema` property to the `x-ms-long-running-operation-options` to specify a ref to the final result schema.

To avoid requiring changes to all the language generators, we should make a change to M4 to artificially add in the 200 response when it finds an LRO that doesn’t have one but has the `final-result-schema`.  With this approach none of the generators would even know it happened, because they would be presented with the same information as before.
 
**Additional context**

I have a branch with the proposed documentation changes for `final-result-schema` [here](https://github.com/mikekistler/autorest/tree/final-state-schema/docs/extensions#x-ms-long-running-operation-options).

mikekistler:
This PR adds support for a new property in x-ms-long-running-operation-options to define the schema for the result of the final-state-via operation.

Includes test and doc updates.

Fixes #4646


packages/extensions/modelerfour/test/modeler/modelerfour.requests.test.ts
```diff
@@ -234,4 +234,70 @@ describe("Modelerfour.Request", () => {
...
+
+      const newSpec: oai3.Model = createTestSpec();
+      addOperation(newSpec, "/products/{id}", {
+        put: {
+          operationId: "product_create",
+          description: "Create a Product",
+          "x-ms-long-running-operation": true,
+          "x-ms-long-running-operation-options": {
+            "final-state-via": "location",
+            "final-state-schema": "#/components/schemas/Product",
```
m-nash, line 283: Can we add a test that validates we still error out if there is no 200 AND there is no `final-state-schema`?
Maybe another one that validates we get an error if we both are set?

mikekistler: Good suggestions @m-nash! I have added these two tests.

But note that there is no error for an LRO without a 200 response -- modeler4 allows this. The test now confirms this behavior.

mikekistler:
@timotheeguerin Could you please review and provide feedback or approve? Thx.

packages/extensions/modelerfour/src/modeler/modelerfour.ts
```diff
@@ -2125,8 +2125,27 @@ export class ModelerFour {
...
+    // If the operation has final-state-schema lro option, add a "200" response if there isn't one already
+    if (httpOperation["x-ms-long-running-operation-options"]?.["final-state-schema"]) {
+      const finalStateSchema = httpOperation["x-ms-long-running-operation-options"]?.["final-state-schema"] as string;
+      if (!Object.keys(modelerResponses).some((k) => k === "200")) {
+        modelerResponses["200"] = {
+          description: "Success",
+          content: {
+            "application/json": {
+              schema: {
+                $ref: finalStateSchema,
```
timotheeguerin, line 2139: Have you tested that from e2e including in the swagger -> openapi3 conversion?

I think for that change we need to make sure the $ref gets converted in every plugin(it might just work, because a lot will just recursively look for $ref)

mikekistler: For swagger -> openapi3 I think all we want is for the extension to be carried forward faithfully ... and it appears that is the case.  It does not -- and IMHO should not -- manufacture a "fake" 200 response in the openapi3.  I'm hoping that we do not need to do that in order for the SDK generators to act as if a 200 response was present.  How can we verify that?

mikekistler: Ah! I understand better now. Yes ... we need to fix the ref. I will try your suggestion to see how that works.

docs/extensions/readme.md
```diff
@@ -1089,7 +1089,10 @@ See [Azure RPC Spec](https://github.com/Azure/azure-resource-manager-rpc/blob/ma
 **Schema**:
 Field Name | Type | Description
 ---|:---:|---
-final-state-via | `string` - one of `azure-async-operation` or `location` or `original-uri` or `operation-location` | `final-state-via` SHOULD BE one of
+final-state-via | `string` - one of `azure-async-operation` or `location` or `original-uri` or `operation-location` | see below
+final-state-schema | A `ref` to the schema of the final result | For languages that make this the result of the LRO
```
heaths, line 1093: Thanks for looping me in, @mikekistler. This seems tangential to what I was actually asking about, though: while I like it, I was actually hoping that Cadl could, by default (or as often as possible), emit `final-state-via` (or `final-state-schema` as seen here) - or equivalent in DPG - so that generators know *definitively* what the final schema would be so they can generate method signatures accordingly.

This came up recently when we realized .NET's final state algorithm was wrong (and some test assertions along with it): https://github.com/Azure/autorest.csharp/pull/2728. Looking across other languages, we found they were all a little different. While I asked @lmazuel to consider normalizing this while leading the DPG efforts, I can't help but wonder if Cadl could emit more by default to not require "guesswork" like this to determine final state.

/cc @bterlson

mikekistler:
@timotheeguerin I tried your suggestion of using `final-state-schema: { "$ref": "#/definitions/Foo" }`, but that didn't "just work" (at least not the way I tried it), so I added the support to convert it. This was pretty easy since we apparently already do this for "x-ms-odata".  Added a test as well.

packages/extensions/modelerfour/src/modeler/modelerfour.ts
```diff
@@ -2146,8 +2146,27 @@ export class ModelerFour {
   processResponses(httpOperation: OpenAPI.HttpOperation, operation: Operation) {
     const acceptTypes = new Set<string>();
 
+    const modelerResponses = httpOperation.responses;
+
+    // If the operation has final-state-schema lro option, add a "200" response if there isn't one already
+    if (httpOperation["x-ms-long-running-operation-options"]?.["final-state-schema"]) {
+      const finalStateSchema = httpOperation["x-ms-long-running-operation-options"]?.["final-state-schema"] as string;
+      if (!Object.keys(modelerResponses).some((k) => k === "200")) {
+        modelerResponses["200"] = {
```
rkmanda, line 2155: Its possible a 200 is defined but without a response body - in which case we would want to use the final-state-schema?

For long running PUTs, 200 and 201 are most likely always expected but it would be desirable to have the final schema specified through final-state-schema?

packages/extensions/modelerfour/test/modeler/modelerfour.requests.test.ts
```diff
@@ -234,4 +235,141 @@ describe("Modelerfour.Request", () => {
...
+      const oldModel = await runModeler(oldSpec);
+      const newModel = await runModeler(newSpec);
+      const oldResponses = oldModel.operationGroups[0].operations[0].responses;
+      const newResponses = newModel.operationGroups[0].operations[0].responses;
+      expect(oldResponses).toBeDefined();
+      expect(newResponses).toBeDefined();
+      expect(newResponses).toEqual(oldResponses);
+    });
+
+    // Modelerfour allows an LRO to have a 202 response without a 200 response.
```
rkmanda, line 342: >202 [](http://example.com/codeflow?start=43&length=3)

applies to 201 as well right?

rkmanda (changes requested):
:clock1:

packages/extensions/modelerfour/src/modeler/modelerfour.ts
mikekistler: > Its possible a 200 is defined but without a response body - in which case we would want to use the final-state-schema?

I think it's unclear what the intent is here so I'd be reluctant to modify an existing 200 response. If the intent is to say that the operation _could_ return a 200 with no body then adding the final response schema is wrong. But if the operation does not actually return 200, then the right fix is to remove the 200 response and then the final-state-schema will work as intended.

> For long running PUTs, 200 and 201 are most likely always expected but it would be desirable to have the final schema specified through final-state-schema?

Long-running PUTs should return a 200 or 201 _AND_ a response body, and I think this should be done with standard OpenAPI. No extension needed.

> The other option would be that we never allow a 200 without a response schema defined. Not sure if there is a linter rule for this.

We do flag this with the data-plane linter.

packages/extensions/modelerfour/test/modeler/modelerfour.requests.test.ts
mikekistler: Right ... neither 200 nor 201 is needed. I'll fix.

packages/extensions/modelerfour/src/modeler/modelerfour.ts
heaths: Opened https://github.com/Azure/cadl-azure/issues/2266 to track.

timotheeguerin (approved):

PR #4664, opened 2022-11-03
Merge openapi to cadl plugin

joheredi:
Add the OpenApi to CADL autorest plugin to the autorest repository


common/config/rush/common-versions.json
```diff
@@ -37,5 +37,8 @@
     // "typescript": [
     //   "~2.4.0"
     // ]
+    "@cadl-lang/compiler": ["~0.31.0"],
+    "ts-node": ["^10.9.1"],
+    "typescript": ["^4.8.4"]
```
timotheeguerin, line 42: any reason why not using the same version here?

packages/extensions/openapi-to-cadl/.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
```
timotheeguerin, line 1: we should already have a prettier config at the root

common/config/rush/common-versions.json
timotheeguerin: I guess autorest is using the older version. Maybe good to update in another pr

common/config/rush/common-versions.json
```diff
@@ -37,5 +37,8 @@
     // "typescript": [
     //   "~2.4.0"
     // ]
+    "@cadl-lang/compiler": ["~0.31.0"],
+    "@azure-tools/cadl-autorest": ["^0.22.0"],
+    "ts-node": ["^10.9.1"]
```
timotheeguerin, line 42: Forgot to update that one but I think we should be able to use the same one for every package

timotheeguerin (approved):

PR #4676, opened 2022-11-18
General pacakges updates

### LINKED ISSUE #4639: Use alternate parser as node-tree-sitter
Seems to not be maintained anymore and can't be installed on node 18 https://github.com/tree-sitter/node-tree-sitter/issues/106

Might be able to use tree sitter directly as they provide wasm binding https://github.com/tree-sitter/tree-sitter

### LINKED ISSUE #4662: [M4] Test and enable CI for node18
Node12 has reached EOL and the azure-sdk-for-js repository has officially dropped support for it which prevents yarn from installing @autorest/typescript.

We should validate and enable support for Node18 for our pipeline validation.

timotheeguerin:
Notable: 
- typescript `4.7` -> `4.9`

fix #4662 
fix #4639


packages/extensions/core/src/lib/plugins/merger.ts
```diff
@@ -135,7 +135,7 @@ export class MultiAPIMerger extends Transformer<any, oai.Model> {
         case "info":
           if (!this.isSecondaryFile) {
             const info = this.getOrCreateObject(target, "info", pointer);
```
joheredi, line 137: maybe you can type info as AnyObject instead of the any cast below?

```ts
const info: AnyObject = this.getOrCreateObject(target, "info", pointer);

```

joheredi (approved):

packages/extensions/core/src/lib/plugins/merger.ts
timotheeguerin: actually was able to update the type of getOrCreateObject to be accurate

PR #4679, opened 2022-11-22
Add @projectedName and update service version

joheredi:
This PR adds @friendlyName on object properties when the serialized name is different to the schema name.

Also moves the service version to the @service decorator

### DIFF
--- a/packages/extensions/openapi-to-cadl/src/utils/decorators.ts
+++ b/packages/extensions/openapi-to-cadl/src/utils/decorators.ts
@@ -72,8 +72,8 @@ export function getPropertyDecorators(property: Property): CadlDecorator[] {
 
   if (property.serializedName !== property.language.default.name) {
     decorators.push({
-      name: "friendlyName",
-      arguments: [property.language.default.name],
+      name: "projectedName",
+      arguments: ["json", property.serializedName],
     });
   }
 
--- a/packages/extensions/openapi-to-cadl/src/utils/decorators.ts
+++ b/packages/extensions/openapi-to-cadl/src/utils/decorators.ts
@@ -70,6 +70,13 @@ export function getPropertyDecorators(property: Property): CadlDecorator[] {
     });
   }
 
+  if (property.serializedName !== property.language.default.name) {
+    decorators.push({
+      name: "friendlyName",
+      arguments: [property.language.default.name],
+    });
+  }
+
   return decorators;
 }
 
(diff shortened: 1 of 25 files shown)


timotheeguerin (approved):

packages/extensions/openapi-to-cadl/test/complexObject/cadl-output/models.cadl
```diff
@@ -16,12 +16,14 @@ Unique identifier representing a specific product for a given latitude &
 longitude. For example, uberX in San Francisco will have a different product_id
 than uberX in Los Angeles.
 """)
+  @friendlyName("productId")
```
catalinaperalta, line 19: I'm seeing an error locally when this decorator is added over model properties. Asking offline about it.

packages/extensions/openapi-to-cadl/test/nh/cadl-output/models.cadl
```diff
@@ -170,39 +170,78 @@ model NotificationHubEntryModel {
...
   type?: NotificationHubContentModelType;
+
+  @friendlyName("notificationHubDescription")
   NotificationHubDescription?: NotificationHubDescriptionModel;
 }
 
 model NotificationHubDescriptionModel {
+  @friendlyName("xmlnsI")
   "xmlns:i"?: NotificationHubDescriptionModelXmlnsI;
+
```
catalinaperalta, line 181: Based off of the offline discussion, I think the output here should look like:
```suggestion
  @projectedNames("json", "xmlns:i")
  "xmlnsI"?: NotificationHubDescriptionModelXmlnsI;
 
```
cc @timotheeguerin

catalinaperalta: where the serialized name is in projected names

packages/extensions/openapi-to-cadl/src/utils/decorators.ts
```diff
@@ -72,8 +72,8 @@ export function getPropertyDecorators(property: Property): CadlDecorator[] {
 
   if (property.serializedName !== property.language.default.name) {
     decorators.push({
-      name: "friendlyName",
-      arguments: [property.language.default.name],
+      name: "projectedName",
+      arguments: ["json", property.serializedName],
```
catalinaperalta, line 76: Should `json` be hardcoded here? Or could there be cases where it's `xml` for instance? I think that info probably shows up somewhere in the tree

timotheeguerin: I think cadl is lacking full xml support for now so I'm fine keeping it like that until we want to convert an xml service and get the bigger picture

PR #4691, opened 2022-12-12
[openAPI-to-cadl] Remove cadl-python from dependencies

tjprescott:
Can't see what purpose this would have since the tool is all typescript. This dependency makes it impossible to use the plugin without Python installed. It should be unnecessary.


tjprescott:
@timotheeguerin not sure why it was there. Our partner team reported that she was unable to run the plugin unless she installed Python, and this would be why. But maybe there is a need for it.

packages/extensions/openapi-to-cadl/package.json
```diff
@@ -46,7 +46,6 @@
     "@cadl-lang/prettier-plugin-cadl": "^0.5.17",
```
timotheeguerin, line 46: Need to make a changelog`rush change` so we can release a new version. O instead bump the version in pacakge.json

tjprescott: I'll wait for Jose to weigh-in. Maybe there's a reason to have cadl-python as a dependency but I suspect not. It's not an urgent PR.

joheredi: I think I used it for testing and forgot to remove it. Thanks for sending out the change

joheredi (approved):

PR #4715, opened 2023-02-22
[Openapi to Cadl] Enum improvements

joheredi:
Some improvements in enum:

- Honor docs for enum elements
- Simplify Extensible enums (Now using @fixed for non-extensible enums)
- Removed the "friendly names" for enum elements that the tool was making up.

Other improvements
- Add missing host parameter
- Update to cadl compiler version 0.40.0

Note: a change to update references to typespec will be following this PR


packages/extensions/openapi-to-cadl/src/utils/decorators.ts
```diff
@@ -80,6 +82,18 @@ export function getPropertyDecorators(property: Property): CadlDecorator[] {
   return decorators;
 }
 
+export function getEnumlDecorators(enumeration: SealedChoiceSchema | ChoiceSchema): CadlDecorator[] {
```
tjprescott, line 85: Is this a typo?  Should it be `getEnumDecorators`?

packages/extensions/openapi-to-cadl/src/transforms/transform-service-information.ts
```diff
@@ -4,6 +4,7 @@ import { getFirstEndpoint } from "../utils/get-endpoint";
 import { isConstantSchema } from "../utils/schemas";
 
 export function transformServiceInformation(model: CodeModel): ServiceInformation {
+  const foo = transformBaseUrl(model);
```
tjprescott, line 7: Assuming this is debug code?

packages/extensions/openapi-to-cadl/test/analyzeText/cadl-output/models.cadl
```diff
@@ -6,16 +6,27 @@ using Azure.Core;
 
 namespace Azure.Language.Authoring;
 
-enum ProjectKindKnownValues {
+enum ProjectKind {
```
tjprescott, line 9: Seems like there should be an `@fixed` enum in here soemwhere.

tjprescott: Nevermind I found one elsewhere.

packages/extensions/openapi-to-cadl/test/anomalyDetector/cadl-output/models.cadl
```diff
@@ -495,7 +483,7 @@ the anomaly score of the subsequent point.
   alignPolicy?: AlignPolicy;
 
   @doc("Model status. One of CREATED, RUNNING, READY, and FAILED.")
-  status?: ModelStatus;
+  status?: modelStatus;
```
tjprescott, line 486: This seems wrong? TypeSpec model names should be PascalCase, so this should be `ModelStatus`.

packages/extensions/openapi-to-cadl/test/bodyComplex/cadl-output/models.cadl
```diff
@@ -4,35 +4,26 @@ using Cadl.Rest;
 
 namespace AutoRestComplexTestService;
 
-enum CMYKColorsKnownValues {
-  Cyan: "cyan",
+enum CMYKColors {
```
tjprescott, line 7: This will raise a warning in Azure because it isn't PascalCase ("CmykColors").

packages/extensions/openapi-to-cadl/src/transforms/transform-service-information.ts
joheredi: Oh yeah good catch thanks :)

packages/extensions/openapi-to-cadl/test/anomalyDetector/cadl-output/models.cadl
joheredi: So this is coming from removing `prenamer` and it is taking the name straight from Swagger. Prenamer was generating a few `AutoGenerated` names and also some weird Enum definitions, for example, it was creating some aliases that didn't make sense.

For example, it was generating
image

Where we'd expect  this, which is what swagger has:
image

packages/extensions/openapi-to-cadl/test/bodyComplex/cadl-output/models.cadl
joheredi: This is the way it is defined in Swagger, if the editor/compiler rise Warnings I think it is okay to let Typespec authors fix it, rather than getting into the business of autogenerating names in the right casing.

Do we also have warnings for model names?

packages/extensions/openapi-to-cadl/test/bodyComplex/cadl-output/models.cadl
```diff
@@ -50,94 +41,86 @@ model Error {
   message?: string;
 }
 
-model IntWrapper {
+model int_wrapper {
```
tjprescott, line 44: This will also issue a warning, but I suspect this was because of OpenAI?

packages/extensions/openapi-to-cadl/test/bodyComplex/cadl-output/models.cadl
```diff
@@ -50,94 +41,86 @@ model Error {
...
-model DictionaryWrapper {
+model dictionary_wrapper {
   @doc("Dictionary of <string>")
   defaultProgram?: Record<string>;
 }
 
-model Siamese {
-  ...Cat;
+model siamese {
+  ...cat;
```
tjprescott, line 108: Yeah, again this seems like a regression. It should be `...Cat`

tjprescott (changes requested):
There seems to be a regression around model names, which need to be PascalCased. There are a bunch of lowercased ones.

docs/extensions/readme.md
```diff
@@ -1082,8 +1082,6 @@ Some requests like creating/deleting a resource cannot be carried out immediatel
 
 When `x-ms-long-running-operation-options` is specified, there should also be a `x-ms-long-running-operation: true` specified.
 
-See [Azure RPC Spec](https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/Addendum.md#asynchronous-operations) for asynchronous operation notes.
```
joheredi, line 1085: This is making CI fail because the link is private

packages/extensions/openapi-to-cadl/test/bodyComplex/cadl-output/models.cadl
joheredi: Added some pretransform renames to help with these

packages/extensions/openapi-to-cadl/test/anomalyDetector/cadl-output/models.cadl
joheredi: Added pretransform to fix this casing

tjprescott (approved):

PR #4738, opened 2023-05-23
[OpenAPI to TypeSpec] - Support primitive types as array elements

joheredi:
We had limited the support of array elements to Model, however we should also support primitive types.

We are also updating so that it generates TypeSpec compatible with compiler version 0.44.0


packages/extensions/openapi-to-cadl/src/transforms/transform-object.ts
```diff
@@ -25,7 +25,7 @@ import { transformValue } from "../utils/values";
 
 const cadlTypes = new Map<SchemaType, string>([
   [SchemaType.Date, "plainDate"],
-  [SchemaType.DateTime, "zonedDateTime"],
+  [SchemaType.DateTime, "offsetDateTime"],
```
timotheeguerin, line 28:
```suggestion
  [SchemaType.DateTime, "offsetDateTime"],
```

so I think we want to recommend people to use `utcDateTIme`. Even though this change is probably safer maybe we should still try to use `utcDateTime` with a todo to go double check?

catalinaperalta: I agree with Tim's idea. Since service teams will typically use utcDateTime, we should use that and let them check it if they need to update it.

joheredi: Makes sense

packages/extensions/openapi-to-cadl/src/emiters/emit-cadl-config.ts
```diff
@@ -3,8 +3,8 @@ import { formatFile } from "../utils/format";
 
 export async function emitCadlConfig(filePath: string): Promise<void> {
   const content = `
-  emitters:
-    "@azure-tools/typespec-autorest": true
+  emit:
+    - "@azure-tools/typespec-autorest": true
```
catalinaperalta, line 7: This should be updated:
```suggestion
    - "@azure-tools/typespec-autorest"
```

packages/extensions/openapi-to-cadl/test/analyzeText/tsp-output/package.json
```diff
@@ -4,13 +4,13 @@
   "description": "The language service API is a suite of natural language processing (NLP) skills built with best-in-class Microsoft machine learning algorithms. The API can be used to analyze unstructured text for tasks such as sentiment analysis, key phrase extraction, language detection and question answering. Further documentation can be found in <a href=\"https://docs.microsoft.com/en-us/azure/cognitive-services/language-service/overview\">https://docs.microsoft.com/en-us/azure/cognitive-services/language-service/overview</a>.",
   "license": "MIT",
   "dependencies": {
-    "@typespec/compiler": "^0.41.0",
```
catalinaperalta, line 7: We dont need to output a package.json anymore since the specs repo relies on a global package.json now

catalinaperalta: We dont need to include this change in this PR it's just a heads up so we remove it eventually

joheredi: Makes sense I have added this change, you can still generate with --include-package, but the default is false.

timotheeguerin (approved):

PR #4761, opened 2023-07-18
Fix formatting

aelij:
Table did not render correctly


timotheeguerin:
Hi @aelij thanks for the contribution sorry for now getting to it earlier. Seems liek the markdown files needs formatting now but otherwise the change looks good

PR #4842, opened 2023-12-15
Fix typo in autorest-configuration-schema.ts

glahaye:
description: "Reservered for future use." -> description: "Reserved for future use."


timotheeguerin:
Hey sorry, it means it needs a change log

PR #4863, opened 2024-01-16
Use Auth shcemes

### LINKED ISSUE #4791: [openapi-to-tsp] Emit authentication defintion
For swagger definitions like this one: 

``` 
"securityDefinitions": {
    "azure_auth": {
      "flow": "implicit",
      "authorizationUrl": "https://login.microsoftonline.com/common/oauth2/authorize",
      "scopes": {
        "user_impersonation": "impersonate your user account"
      },
      "type": "oauth2",
      "description": "Azure Active Directory OAuth2 Flow."
    }
  },
```

Could we support emitting the corresponding TypeSpec definition with @useAuth?

joheredi:
Fixes #4791

### DIFF
--- a/packages/extensions/openapi-to-typespec/src/generate/generate-service-information.ts
+++ b/packages/extensions/openapi-to-typespec/src/generate/generate-service-information.ts
@@ -1,4 +1,4 @@
-import { TypespecProgram, EndpointParameter } from "../interfaces";
+import { TypespecProgram, EndpointParameter, Auth } from "../interfaces";
 import { getOptions } from "../options";
 import { generateDocs } from "../utils/docs";
 import { getNamespace } from "../utils/namespace";
@@ -13,6 +13,7 @@ export function generateServiceInformation(program: TypespecProgram) {
     definitions.push(`@armProviderNamespace`);
   }
 
+  serviceInformation.authentication?.map((auth) => generateUseAuth(auth, definitions));
   definitions.push(`@service({
     title: "${serviceInformation.name}"
   })`);
@@ -90,4 +91,18 @@ function getEndpointParameters(endpoint: string) {
   }
 
   return params;
+}
+
+function generateUseAuth(auth: Auth | undefined, statements: string[]): void {
+  if(!auth) {
+    return;
+  }
+  if(auth.kind === "AadOauth2Auth") {
+    const scopes = `[${auth.scopes.map(s => `"${s}"`).join()}]`
+    statements.push(`@useAuth(AadOauth2Auth<${scopes}>)`);
+  }
+
+  // TODO: Add support for other AAD auth types
+
+  return;
 }
--- a/packages/extensions/openapi-to-typespec/src/generate/generate-service-information.ts
+++ b/packages/extensions/openapi-to-typespec/src/generate/generate-service-information.ts
@@ -94,12 +94,16 @@ function getEndpointParameters(endpoint: string) {
 }
 
 function generateUseAuth(auth: Auth | undefined, statements: string[]): void {
-  if(!auth) {
+  if (!auth) {
     return;
   }
-  if(auth.kind === "AadOauth2Auth") {
-    const scopes = `[${auth.scopes.map(s => `"${s}"`).join()}]`
+  if (auth.kind === "AadOauth2Auth") {
+    const scopes = `[${auth.scopes.map((s) => `"${s}"`).join()}]`;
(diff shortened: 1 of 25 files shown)


packages/extensions/openapi-to-typespec/src/interfaces.ts
```diff
@@ -63,13 +63,31 @@ export interface TypespecResource {
   response: TypespecDataType;
 }
 
+export interface AadOauth2AuthFlow {
+  kind: "AadOauth2Auth";
+  scopes: string[];
```
catalinaperalta, line 68: I think some of these interfaces might need the auth url and token url

packages/extensions/openapi-to-typespec/src/generate/generate-service-information.ts
```diff
@@ -91,3 +92,17 @@ function getEndpointParameters(endpoint: string) {
...
   return params;
 }
+
+function generateUseAuth(auth: Auth | undefined, statements: string[]): void {
+  if(!auth) {
+    return;
+  }
+  if(auth.kind === "AadOauth2Auth") {
+    const scopes = `[${auth.scopes.map(s => `"${s}"`).join()}]`
+    statements.push(`@useAuth(AadOauth2Auth<${scopes}>)`);
```
catalinaperalta, line 102: What happens if there's an auth url specified?

catalinaperalta: Also will this support definitions like the following: https://github.com/Azure/azure-rest-api-specs/blob/e69839a006ffc8902412cc38228fd600942c2f73/specification/cognitiveservices/ContentSafety/main.tsp#L17-L22 ?

joheredi: It will but it wouls generate 2 @useAuth decorators instead of the union

packages/extensions/openapi-to-typespec/src/interfaces.ts
joheredi: I couldn't see those in M4, it seems to assume the login.microsoft url

packages/extensions/openapi-to-typespec/test/arm-agrifood/tsp-output/main.tsp
```diff
@@ -20,6 +20,7 @@ using Azure.Core;
 using Azure.ResourceManager;
 using TypeSpec.Versioning;
 @armProviderNamespace
+@useAuth(AadOauth2Auth<["user_impersonation"]>)
```
tadelesh, line 23: it seems we could not get scope info from m4 for arm?

packages/extensions/openapi-to-typespec/src/generate/generate-service-information.ts
catalinaperalta: Multiple auth decorators are not supported :/ Only the first one would be used. Here is the issue I opened for the linter to warn against having multiple useAuth decorators: https://github.com/microsoft/typespec/issues/2787

catalinaperalta: We'd need to do the | syntax when generating

joheredi: Great! Thanks for catching this :)

packages/extensions/openapi-to-typespec/test/arm-agrifood/tsp-output/main.tsp
joheredi: not sure I follow, this seems to be the scope defined in the swagger
image

catalinaperalta (approved):
This is a great improvement for adding auth to the converter. Are there plans to capture more complex flows like in this example: https://github.com/Azure/azure-rest-api-specs/blob/27bae6b2e0b1283f806213ed90e47c5db8c5d0a8/specification/logz/resource-manager/Microsoft.Logz/preview/2020-10-01-preview/logz.json#L17-L26 ?

catalinaperalta:
> This is a great improvement for adding auth to the converter. Are there plans to capture more complex flows like in this example: https://github.com/Azure/azure-rest-api-specs/blob/27bae6b2e0b1283f806213ed90e47c5db8c5d0a8/specification/logz/resource-manager/Microsoft.Logz/preview/2020-10-01-preview/logz.json#L17-L26 ?

I understand if this is possibly already captured by the AadOauth2Auth type

PR #4868, opened 2024-01-23
Add feature to converter

### LINKED ISSUE #4861: Support self-defined common-type resources
ref: https://github.com/Azure/typespec-azure/issues/104#issuecomment-1888024030

tadelesh:
1. support flatten
2. support custom resource, fix #4861 
3. fix doc multi-line issue
4. fix naming escape issue
5. fix duration default value problem


packages/extensions/openapi-to-typespec/src/generate/generate-arm-resource.ts
```diff
@@ -25,9 +26,48 @@ export function generateArmResource(resource: TspArmResource): string {
...
+    );
+    definitions.push(
+      `#suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "For backward compatibility"`,
+    );
+    definitions.push(`@includeInapplicableMetadataInPayload(false)`);
+
+    if (!getArmCommonTypeVersion()) {
+      if (resource.baseModelName) {
+        definitions.push(`model ${resource.name} extends ${resource.baseModelName} {`);
+      } else {
```
joheredi, line 51: I wonder if we could abstract all of this new logic into a function and just call the function here to help readability and maintenance in the future.

joheredi (approved):

PR #4869, opened 2024-01-23
Update tspconfig for data-plane

### LINKED ISSUE #4787: [openapi-to-tsp] Improve tspconfig.yaml
Can we emit a tspconfig.yaml with the following configuration for typespec-autorest: 

```
"@azure-tools/typespec-autorest":
    azure-resource-provider-folder: "data-plane"
    emitter-output-dir: "{project-root}/.."
    examples-directory: "examples"
    output-file: "{azure-resource-provider-folder}/{service-name}/{version-status}/{version}/ContosoWidget.json"
```

This config is the default necessary for checking in specs in the azure rest api specs repo. Maybe this needs a flag to specify that we do this only for Azure specs?

joheredi:
Update data-plane tspconfig to comply with the azure-rest-spec repo requirements

fixes #4787 4787

### DIFF
--- a/packages/extensions/openapi-to-typespec/src/emiters/emit-typespec-config.ts
+++ b/packages/extensions/openapi-to-typespec/src/emiters/emit-typespec-config.ts
@@ -1,12 +1,20 @@
+import { pascalCase } from "@azure-tools/codegen";
+import { TypespecProgram } from "interfaces";
 import { getSession } from "../autorest-session";
 import { getOptions } from "../options";
 import { formatFile } from "../utils/format";
 
-export async function emitTypespecConfig(filePath: string): Promise<void> {
+export async function emitTypespecConfig(filePath: string, programDetails: TypespecProgram): Promise<void> {
   const session = getSession();
   const { isArm } = getOptions();
   let content = `emit:
-  - "@azure-tools/typespec-autorest"
+  - "@azure-tools/typespec-autorest":
+    azure-resource-provider-folder: "data-plane"
+    emitter-output-dir: "{project-root}/.."
+    examples-directory: "examples"
+    output-file: "{azure-resource-provider-folder}/{service-name}/{version-status}/{version}/${pascalCase(
+      programDetails.serviceInformation.name,
+    )}.json"
   # Uncomment this line and add "@azure-tools/typespec-python" to your package.json to generate Python code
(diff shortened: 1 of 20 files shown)


packages/extensions/openapi-to-typespec/test/anomalyDetector/tsp-output/tspconfig.yaml
```diff
@@ -1,5 +1,9 @@
 emit:
-  - "@azure-tools/typespec-autorest"
+  - "@azure-tools/typespec-autorest":
```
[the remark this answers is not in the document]
joheredi: good catch! Thanks :)

packages/extensions/openapi-to-typespec/src/emiters/emit-typespec-config.ts
```diff
@@ -1,12 +1,20 @@
...
-export async function emitTypespecConfig(filePath: string): Promise<void> {
+export async function emitTypespecConfig(filePath: string, programDetails: TypespecProgram): Promise<void> {
   const session = getSession();
   const { isArm } = getOptions();
   let content = `emit:
-  - "@azure-tools/typespec-autorest"
+  - "@azure-tools/typespec-autorest":
+    azure-resource-provider-folder: "data-plane"
+    emitter-output-dir: "{project-root}/.."
+    examples-directory: "examples"
```
catalinaperalta, line 14: Same feedback as Tim about where this should go. 
Also, I think this line should be:
```suggestion
    examples-directory: "{project-root}/examples"
```

packages/extensions/openapi-to-typespec/src/emiters/emit-typespec-config.ts
```diff
@@ -1,12 +1,20 @@
...
+    output-file: "{azure-resource-provider-folder}/{service-name}/{version-status}/{version}/${pascalCase(
+      programDetails.serviceInformation.name,
+    )}.json"
```
catalinaperalta, line 17: Do we want to just simplify this and leave it as openapi.json?

joheredi: Makes sense I like it better

timotheeguerin (approved):

PR #4896, opened 2024-02-28
Fix crash with duplicate paths

### LINKED ISSUE #4882: Fatal error in lintdiff execution resulting in a healthy automated check when it is expected to fail
**Before filling a bug**

- [ X] have you checked the [faq](https://github.com/Azure/autorest/tree/main/docs/faq.md) for known issues.
- [ X] have you checked existing issues

In the following run, there was a fatal error encountered. the expectation in such cases is that the fatal error translate into an error for the automated check. However, in this case it resulted in the automated swagger lintdiff check being marked as healthy

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3491008&view=logs&j=0574a2a6-2d0a-5ec6-40e4-4c6e2f70bea2&t=80c3e782-49f0-5d1c-70dd-cbee57bdd0c7

```
Please provide as much information as you can. This would be:
- OpenAPI files having the issues
- Autorest command used
```

**Expected behavior**
The automated check for swagger lintdiff should have failed.

**Additional context**

timotheeguerin:
fix #4882 too

### DIFF
--- a/packages/extensions/core/src/app.ts
+++ b/packages/extensions/core/src/app.ts
@@ -351,7 +351,9 @@ async function main() {
     // - doing the inversion (instanceof Error) doesn't reliably work since that seems to return false on Errors marshalled from safeEval
     if (e instanceof Exception) {
       logger.log({ level: "error", message: e.message });
-      exitCode = e.exitCode;
+      if (e.exitCode) {
+        exitCode = e.exitCode;
+      }
     }
     if (e !== false) {
       logger.log({ level: "error", message: `!${e}` });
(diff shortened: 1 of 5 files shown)


packages/extensions/core/src/app.ts
```diff
@@ -351,7 +351,9 @@ async function main() {
     // - doing the inversion (instanceof Error) doesn't reliably work since that seems to return false on Errors marshalled from safeEval
     if (e instanceof Exception) {
       logger.log({ level: "error", message: e.message });
-      exitCode = e.exitCode;
+      if (e.exitCode) {
```
mikeharder, line 354: Can you explain why this check is needed?  Might be obvious but I don't see it.

PR #4907, opened 2024-03-12
Inlcude csharp rename decorator when openai-to-typespec generating tsp

RodgeFu:
Fixes https://github.com/Azure/autorest.csharp/issues/4236.
Generate csharp renaming decorator when openai-to-typespec generating tsp. (include rename of resource, resource.property, model, model.property, enum, enum.member and operation name)

related autorest.csharp change can be found at https://github.com/Azure/autorest.csharp/pull/4380


packages/extensions/openapi-to-typespec/test/schemaRegistry/readme.md
```diff
@@ -1,6 +1,6 @@
 ```yaml
 input-file:
-  - https://raw.githubusercontent.com/Azure/azure-rest-api-specs/main/specification/schemaregistry/data-plane/Microsoft.EventHub/stable/2021-10/schemaregistry.json
```
tadelesh, line 3: can we change to a fixed commit to prevent further problem?

packages/extensions/openapi-to-typespec/src/emiters/emit-client.ts
```diff
@@ -34,8 +38,14 @@ function generateClient(program: TypespecProgram) {
         .filter((r) => r !== "")
         .join("\n\n")
     : "";
+
+  const enums = models.enums
+    .map(generateEnumClientDecorator)
+    .filter((r) => r !== "")
+    .join("\n\n");
+
   if (objects === "" && armResources === "") {
```
tadelesh, line 47:
```suggestion
  if (objects === "" && armResources === "" && enums === "") {
```

need also check enums should be empty if there is no renaming,

packages/extensions/openapi-to-typespec/src/emiters/emit-main.ts
```diff
@@ -33,6 +33,7 @@ function getArmServiceInformation(program: TypespecProgram) {
     `import "@azure-tools/typespec-azure-core";`,
     `import "@azure-tools/typespec-azure-resource-manager";`,
     `import "./models.tsp";`,
+    `import "./client.tsp";`,
```
tadelesh, line 36:
```suggestion
```

do not import client.tsp in main.tsp. instead, import main.tsp in client.tsp and compile with client.tsp.

packages/extensions/openapi-to-typespec/src/utils/paging.ts
```diff
@@ -74,7 +74,7 @@ export function markPagination(codeModel: CodeModel) {
       if (!isPageableOperation(operation)) {
         continue;
       }
-      const itemName = paginationExtension.itemName || "value";
+      const itemName = (paginationExtension.itemName as string) || "value";
```
tadelesh, line 77: from https://github.com/Azure/autorest/tree/main/docs/extensions#x-ms-pageable, we could see the default property name should be `value`. i don't think we should do the lowercase, it should be a wrong swagger definition.

packages/extensions/openapi-to-typespec/src/emiters/emit-main.ts
tadelesh: and if there is nothing to generate for client.tsp, we will not generate it.

packages/extensions/openapi-to-typespec/test/schemaRegistry/readme.md
RodgeFu: resolve by using fixed url of original spec

packages/extensions/openapi-to-typespec/src/emiters/emit-client.ts
RodgeFu: fixed.thx

packages/extensions/openapi-to-typespec/src/utils/paging.ts
RodgeFu: reverted

packages/extensions/openapi-to-typespec/src/emiters/emit-main.ts
RodgeFu: removed. thanks

packages/extensions/openapi-to-typespec/convert.ps1
```diff
@@ -14,6 +14,10 @@ param(
     [string]
     # Specifies the swagger config file, not the swagger json, but the readme config.
     $swaggerConfigFile,
+    [Parameter(Mandatory)]
+    [string]
+    # Specifies the corresponding autorest.md file in azure-sdk-for-net repo used by autorest.csharp codegen.
+    $csharpAutorestFile,
```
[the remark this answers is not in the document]
RodgeFu: updated as discussed

tadelesh (approved):

tadelesh:
another concern: if i change the script to use this latest converter version without using the latest csharp codegen version, there will be no error, right? is it backward compatible?

packages/extensions/openapi-to-typespec/convert.ps1
```diff
@@ -10,25 +10,23 @@
 #>
 
 param(
-    [Parameter(Mandatory)]
+    [Parameter(Mandatory = $true, HelpMessage = "Specifies the swagger config file (not the swagger json, but the readme config) or autorest.md file in the azure-sdk-for-net repo if .net related configuration is expected to be included.")]
+    [string] 
+    $ConfigFile,
```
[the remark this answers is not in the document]
RodgeFu: the change reverted, thanks.

RodgeFu:
> another concern: if i change the script to use this latest converter version without using the latest csharp codegen version, there will be no error, right? is it backward compatible?

There should be no change if a normal swagger config is used. The change should only have impact when an autorest.md with csharp related rename configuration is used. and all test passed. thanks.

PR #4915, opened 2024-03-15
Replace "new Buffer()" with "Buffer.from()"

mikeharder:
- Removes deprecation warning


timotheeguerin (approved):
Doesn't look like any CI is running anymore

PR #4920, opened 2024-03-19
Use 1es pipeline templates

hallipr:


benbp (approved):

timotheeguerin (approved):

.prettierignore
```diff
@@ -335,6 +335,9 @@ packages/libs/codemodel/.resources
 **/test/scenarios/expected/**/*.json
 packages/extensions/core/test/plugins/*/expected/*.json
 
+# ensure rush doesn't catch on any yml files
+**/*.yml
```
timotheeguerin, line 339: hhm do you really want to disable formatting of yaml files? is there an issue?

hallipr: check-format is failing on the new 1es-redirect.yml template, but it doesn't say what the problem is.

`rush format` just crashes

hallipr:
```
Starting "rush check-format"


> /mnt/vss/_work/1/s/packages/extensions/core/node_modules/.bin/prettier --list-different --config .prettierrc.yml --ignore-path .prettierignore **/*.{ts,js,cjs,mjs,json,yml,yaml,cadl,tsp,md}
eng/pipelines/templates/1es-redirect.yml

ERROR: Files above are not formatted correctly. Run `rush format` and commit the changes.

```

timotheeguerin: hhm let me take a look, formatting the content of the file copied works fine for me, maybe we are using an old prettier version that has a bug

timotheeguerin: `rush format` formatted the file fine, i pushed the change, can you share the crash if it happens again?

hallipr: I've removed the yml ignore lines as the yaml is now pretty.

PR #4932, opened 2024-03-26
Transition `autorest.v2` release pipelines to `1es-templates`

scbedd:


scbedd:
The [test run](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3640674&view=logs&j=bc38e8b8-e027-53cb-48e7-2adbd1070eca) failed as expected. It was really trying to push the version that was already there, and we failed as expected. This is g2g.

.scripts/publish.yaml
```diff
@@ -6,28 +6,41 @@
...
-    publish-release --token $(azuresdk-github-pat) --repo autorest --owner azure --name autorest-core-$v --tag autorest-core-$v --notes='v2 prerelease build' --prerelease --editRelease false --assets ./src/autorest-core/microsoft.azure-autorest-core-$v.tgz --target_commitish $(Build.SourceBranchName)
-    rc=$?; if [ $rc -ne 0 ]; then exit $rc ; fi 
-    cd ..
+extends:
+  template: /eng/pipelines/templates/1es-redirect.yml
+  parameters:
+    stages:
+    - stage: Publish_Stage
+      variables:
+        - template: /eng/pipelines/templates/variables/image.yml
```
benbp, line 15: If you're using the stage/variable expansion model from Patrick, then wouldn't we also want to move the image.yml inclusion inline to 1es-redirect?

benbp (approved):

.scripts/publish.yaml
scbedd: /shrug. it works.

PR #4934, opened 2024-03-29
[v2] Report full error for schema violations

### LINKED ISSUE #4933: [v2] Report full error for schema violations
When reporting schema violations, I think we should include the full dump of the error object, not just the message.

https://github.com/Azure/autorest/blob/6a1a09115ff1389ebee194e1889bf155c7546f27/src/autorest-core/lib/pipeline/schema-validation.ts#L25-L33

Replace `error.message` with something like `JSON.stringify(error)`.

mikeharder:
- Fixes #4933

## Before
```
ERROR: Schema violation: Data does not match any schemas from 'oneOf'
```

## After
```
ERROR: Schema violation: Data does not match any schemas from 'oneOf'
{
  code: 'ONE-OF-MISSING',
  message: "Data does not match any schemas from 'oneOf'",
  inner: [
    {
      code: 'OBJECT-MISSING-REQUIRED-PROPERTY',
      message: 'Missing required property: parameters',
      description: 'Describes the format of an example defined using the x-ms-examples extension.'
    },
    {
      code: 'OBJECT-MISSING-REQUIRED-PROPERTY',
      message: 'Missing required property: $ref',
    }
  ]
}
```

### DIFF
--- a/src/autorest-core/lib/pipeline/schema-validation.ts
+++ b/src/autorest-core/lib/pipeline/schema-validation.ts
@@ -11,6 +11,7 @@ import { Parse } from '../ref/yaml';
 import { CreatePerFilePlugin, PipelinePlugin } from "./common";
 import { Channel } from "../message";
 import { OperationAbortedException } from '../exception';
+import { inspect } from 'util';
 
 export function GetPlugin_SchemaValidator(): PipelinePlugin {
   const validator = new SchemaValidator({ breakOnFirstError: false });
@@ -23,7 +24,8 @@ export function GetPlugin_SchemaValidator(): PipelinePlugin {
     const errors = await new Promise<{ code: string, params: string[], message: string, path: string }[] | null>(res => validator.validate(obj, extendedSwaggerSchema, (err, valid) => res(valid ? null : err)));
     if (errors !== null) {
       for (const error of errors) {
-        const errorString = JSON.stringify(error, null, 2);
+        // Replace '_' with '-' to avoid output formatter interpreting as italics
+        const errorString = inspect(error).replaceAll('_', '-');
 
         config.Message({
           Channel: Channel.Error,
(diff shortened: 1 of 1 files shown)


src/autorest-core/lib/pipeline/schema-validation.ts
```diff
@@ -23,12 +23,14 @@ export function GetPlugin_SchemaValidator(): PipelinePlugin {
     const errors = await new Promise<{ code: string, params: string[], message: string, path: string }[] | null>(res => validator.validate(obj, extendedSwaggerSchema, (err, valid) => res(valid ? null : err)));
     if (errors !== null) {
       for (const error of errors) {
+        const errorString = JSON.stringify(error, null, 2);
```
timotheeguerin, line 26: this might fail if the error is not serializable as json(Contains circular reference)

you could use `util.inspect` from node which lets you specify a max depth

src/autorest-core/lib/pipeline/schema-validation.ts
```diff
@@ -23,12 +24,15 @@ export function GetPlugin_SchemaValidator(): PipelinePlugin {
...
+        // Replace '_' with '-' to avoid output formatter interpreting as italics
+        const errorLog = inspect(error, { depth: 3 }).replaceAll('_', '-');
+
         config.Message({
           Channel: Channel.Error,
           Details: error,
           Plugin: "schema-validator",
           Source: [{ document: fileIn.key, Position: { path: parseJsonPointer(error.path) } as any }],
-          Text: `Schema violation: ${error.message}`
+          Text: `Schema violation: ${error.message}\n${errorLog}`
```
mikeharder, line 35: @timotheeguerin: Do you think we should always log this?  Or maybe only when `error.code == ONE_OF_MISSING`?

The details are certainly helpful for `ONE_OF_MISSING`, but I wonder if they might be overkill for all schema violations?

On the other hand, in the case of schema violations, probably the more information we can show the caller the better?

timotheeguerin: In autorest v3 I have a much better log when there is schema errors but here I would say just do the cheapest unless you want to try to cherry pick those pieces

timotheeguerin (approved):

PR #4945, opened 2024-04-10
Change enum definition to latest grammer

### LINKED ISSUE #4943: [openapi-to-typespec] Use `union` instead of `enum`
Azure specs should use the `union` syntax to declare extensible enums instead of the `enum` keyword. We should have the converter output the correct format.

cc @joheredi

pshao25:
Fix #4943


pshao25:
Please hold on this until architects have a decision.

packages/extensions/openapi-to-typespec/test/analyzeText/tsp-output/models.tsp
```diff
@@ -8,91 +8,101 @@ using Azure.Core;
...
-  EmptyRequest,
-  MissingInputDocuments,
-  InvalidDocument,
-  ModelVersionIncorrect,
-  InvalidDocumentBatch,
-  UnsupportedLanguageCode,
-  InvalidCountryHint,
+union InnerErrorCode {
+  string,
+  "InvalidRequest",
```
timotheeguerin, line 65:
```suggestion
  InvalidRequest: "InvalidRequest",
```
the correct way should be with a name otherwise the variants have no names

PR #4958, opened 2024-04-23
Support for anyObject

### LINKED ISSUE #4957: AnyObject as body is not correctly generated to typespec
For what generated for arm-networkanalytics
```
listRolesAssignments is ArmResourceActionSync<
    DataProduct,
    AnyObject,
    ListRoleAssignments
  >;
```
The `AnyBody` from [swagger ](https://github.com/Azure/azure-rest-api-specs/blob/2bb95006b1eaffeae16fcbc20f5d9844309a80a0/specification/networkanalytics/resource-manager/Microsoft.NetworkAnalytics/stable/2023-11-15/networkanalytics.json#L1170)is an object body.

Expected:
```
listRolesAssignments is ArmResourceActionSync<
    DataProduct,
    {},
    ListRoleAssignments
  >;
```

### LINKED ISSUE #4959: parent resource should be a model in the generated tsp
In test packages\extensions\openapi-to-typespec\test\arm-playwrighttesting, the parent resource for Quota is a string.
```
@parentResource("SubscriptionLocationResource")
model Quota is ProxyResource<QuotaProperties> {
  /**
   * The quota name.
   */
  @path
  @key("name")
  @segment("quotas")
  @visibility("read")
  name: QuotaNames;
}
```
Expected: it should be a model.

pshao25:
Fix #4957
Fix #4959


packages/extensions/openapi-to-typespec/src/transforms/transform-arm-resources.ts
```diff
@@ -37,7 +37,8 @@ export function isGeneratedResourceObject(name: string): boolean {
 }
 
 export function replaceGeneratedResourceObject(name: string): string {
-  return generatedResourceObjects.get(name) ?? name;
+  const mappingName = generatedResourceObjects.get(name) ?? name;
+  return mappingName == "AnyObject" ? "{}" : mappingName;
```
tadelesh, line 41: This is only for the renaming part of resource. I suggest to handle the `AnyObject` in the object conversion part.

pshao25: Here it is `{}` because the template only accept `Model|void`, but my intuitive tells me it is `unknown` at first. I think we should do this change as late as possible in case in other scenarios it should be `unknown` or `void`, etc.

tadelesh: Let's sync offline.

tadelesh (approved):

PR #4989, opened 2024-07-08
[Openapi to Typespec] Detect Arm and Fix unsupported char type

### LINKED ISSUE #4991: [openapi-to-typespec] Automatically detect arm
We should be able to use the spec folder name `resource-management` to detect if the conversion should have --arm

joheredi:
* Currently SchemaType.Char is unsupported, mapping it to string to fix the problem
* Automatically detect when a swagger is ARM
Fixes #4991


packages/extensions/openapi-to-typespec/src/options.ts
```diff
@@ -31,8 +31,21 @@ export function getGuessResourceKey(session: Session<CodeModel>) {
...
-  return isArm !== false;
+  if (session.configuration["isArm"] !== undefined) {
+    // If isArm is explicitly set, use it.
+    return Boolean(session.configuration["isArm"]);
+  }
+
+  const inputs = session.configuration["inputFileUris"] as string[];
+
+  for (const input of inputs) {
+    if (input.includes("") && input.includes("resource-manager")) {
```
mikeharder, line 42: Why `input.includes("")`?

mikeharder (approved):

packages/extensions/openapi-to-typespec/src/options.ts
joheredi: Oh no thanks for catching it

catalinaperalta (approved):

PR #5004, opened 2024-08-20
Note that final-state-via is only for POST

jhendrixMSFT:


docs/extensions/readme.md
```diff
@@ -1102,6 +1102,8 @@ final-state-schema | A `ref` to the schema of the final result | For languages t
 
 The polling mechanism in itself remains unchanged, the only impact of this option could be to do an additional final GET, or skip a final GET.
 
+**NOTE** `final-state-via` is _only_ applicable to `POST` operations. Setting this for other operations will have _no effect_.
```
MaryGao, line 1105: For JS final-state-via would take effective with two conditions, first it is in PATCH and POST operations secondly it must be in an AzureAsyncOperation operation.

I also note @lmazuel mentioned in Python we have similar conditions.

```
Python does not honor final-state-via for anything else than POST. Final-State-Via was designed initially as a way to disambiguate when a service team was returning both AzureAsyncOperation and Location at the same time for the initial response on a POST call. It was never intended to be used outside of this context 
```

Issue #4995, opened 2024-08-01
Replace .net dependency from openapi-to-typespec


pshao25:
Currently we depend on .net generator to calculate resource information from swagger. We should remove this dependency, and use ts code instead.

PR #5008, opened 2024-08-28
Remove dependency of .net generator from tsp converter

pshao25:
Fix https://github.com/Azure/autorest/issues/4995

What still left is:
1. Multi path cases depend on TypeSpec design: "arm-apimanagement", "arm-compute", "arm-machinelearningservices"
2. Expanded cases depend on TypeSpec design: "arm-dns"
3. Deal with directives: an example is arm-storage.


packages/extensions/openapi-to-typespec/src/resource/constants.ts
```diff
@@ -0,0 +1,9 @@
+export const ProvidersSegment = "/providers/";
+export const ManagementGroupPath = "/providers/Microsoft.Management/managementGroups/{managementGroupId}";
+export const ManagementGroupScopePrefix = "/providers/Microsoft.Management/managementGroups";
+export const ResourceGroupPath = "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}";
+export const ResourceGroupScopePrefix = "/subscriptions/{subscriptionId}/resourceGroups";
+export const SubscriptionPath = "/subscriptions/{subscriptionId}";
+export const SubscriptionScopePrefix = "/subscriptions";
+export const TenantPath = "";
+export const TenantScopePrefix = "/tenants";
```
ArcturusZhang, line 9: i kind of doubt if we have such cases with this prefix.
For these hard-coded paths, could we have a link of where this comes from? such as a link to the code in our csharp generator.

pshao25: Just borrowed from .net generator. But I think we could refine it.

packages/extensions/openapi-to-typespec/src/resource/utils.ts
```diff
@@ -0,0 +1,76 @@
...
+export function getResourceKeySegment(path: string): string {
+  const segments = path.split("/");
+  return segments[segments.length - 2];
+}
+
+export function getScopePath(path: string): string {
+  const pathToLower = path.toLowerCase();
+
+  const index = pathToLower.lastIndexOf(ProvidersSegment);
+  if (index === 0 && pathToLower.startsWith(ManagementGroupScopePrefix.toLowerCase())) return ManagementGroupPath;
```
live1206, line 46: can be simplified to `pathToLower.startsWith(ManagementGroupScopePrefix.toLowerCase())`
And also, consider to put the exact matches in front and take care of `providers` in the end.
Shall we throw otherwise?

packages/extensions/openapi-to-typespec/src/resource/utils.ts
```diff
@@ -0,0 +1,76 @@
...
+} from "./constants";
+
+export function getResourceType(path: string): string {
+  const index = path.lastIndexOf(ProvidersSegment);
+  if (index < 0 || path.substring(index + ProvidersSegment.length).includes("/") === false) {
+    const pathToLower = path.toLowerCase();
+    if (pathToLower.startsWith(ResourceGroupScopePrefix.toLowerCase())) return "Microsoft.Resources/resourceGroups";
+    if (pathToLower.startsWith(SubscriptionScopePrefix.toLowerCase())) return "Microsoft.Resources/subscriptions";
+    if (pathToLower.startsWith(TenantScopePrefix.toLowerCase())) return "Microsoft.Resources/tenants";
+    throw `Path ${path} doesn't have resouce type`;
```
live1206, line 20: typo: resource

packages/extensions/openapi-to-typespec/src/resource/resource-equivalent.ts
```diff
@@ -0,0 +1,65 @@
...
+  for (const property of getAllProperties(schema)) {
+    if (property.flattenedNames) continue;
+
+    if (property.serializedName === "location" && isStringSchema(property.schema)) {
+      isLocationFound = true;
+    } else if (property.serializedName === "tags" && isDictionarySchema(property.schema)) {
+      isTagsFound = true;
+    }
+  }
+  return isLocationFound && isTagsFound;
```
live1206, line 43: should this be `isResource` as well?

packages/extensions/openapi-to-typespec/src/resource/find-parent.ts
```diff
@@ -0,0 +1,171 @@
...
+    const trimmedRequestPath = requestPath.substring(providerIndexOfRequest);
+
+    if (!pathIncludes(trimmedResourceRequestPath, trimmedRequestPath)) continue;
+    if (getResourceDataSchema(operationSet) !== itemType) continue;
+
+    candidates.push(operationSet);
+  }
+
+  if (candidates.length === 0) return false;
+  const bestOne = candidates.sort((a, b) => b.RequestPath.split("/").length - a.RequestPath.split("/").length)[0];
```
live1206, line 53: consider to put L32 here

packages/extensions/openapi-to-typespec/src/resource/find-parent.ts
```diff
@@ -0,0 +1,171 @@
...
+    resourceCollectionMethodCache.set(bestOne, [operation]);
+  }
+  return true;
+}
+
+export function getResourceCollectionOperations(set: OperationSet): Operation[] {
+  return resourceCollectionMethodCache.get(set) ?? [];
+}
+
+export function setParentOfOtherOperation(
```
live1206, line 67: what is other operation?

tadelesh (commented):
why we still want to keep `resources.json`, it is kind of a middle output from .net codegen. since we have the resource detection logic, then we could compose the typespec resource model directly.

pshao25:
> why we still want to keep `resources.json`, it is kind of a middle output from .net codegen. since we have the resource detection logic, then we could compose the typespec resource model directly.

We can definitely delete it, and it is actual my next step. In this PR, I don't want to do take this step too big, therefore to keep this file so that we could see the diff clearly.

packages/extensions/openapi-to-typespec/src/resource/find-parent.ts
pshao25: Action is an example.

packages/extensions/openapi-to-typespec/src/resource/resource-equivalent.ts
pshao25: Yes, TrackedResource is a Resource too.

packages/extensions/openapi-to-typespec/src/resource/utils.ts
```diff
@@ -0,0 +1,76 @@
...
+    .split("/")
+    .reduce((result, current, index) => {
+      if (index === 0 || index % 2 === 1) return result === "" ? current : `${result}/${current}`;
+      else return result;
+    }, "");
+}
+
+export function getResourceKey(path: string): string {
+  const segments = path.split("/");
+  return segments[segments.length - 1].replace(/^\{(.+)\}$/, "$1");
```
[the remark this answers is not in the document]
pshao25: Then nothing is replaced. So keep the last segment value, which is a constant.

live1206 (approved):

Issue #4968, opened 2024-05-21
EXEC : error | error : spawn EINVAL


archerzz:
**Before filling a bug**

- [x] have you checked the [faq](https://github.com/Azure/autorest/tree/main/docs/faq.md) for known issues.
- [x] have you checked existing issues

**Describe the bug**

Starting from May 17, there were autorest errors in one of our pipeline: https://dev.azure.com/azure-sdk/internal/_build?definitionId=5990&_a=summary
```
fatal | Process() cancelled due to failure
EXEC : error | error : spawn EINVAL [D:\a\_work\1\s\azure-sdk\sdk\resourcehealth\Azure.ResourceManager.ResourceHealth\src\Azure.ResourceManager.ResourceHealth.csproj]
error | Autorest completed with an error. If you think the error message is unclear, or is a bug, please declare an issues at https://github.com/Azure/autorest/issues with the error message you are seeing.
C:\Users\cloudtest\.nuget\packages\microsoft.azure.autorest.csharp\3.0.0-alpha.20240519.1\buildMultiTargeting\Microsoft.Azure.AutoRest.CSharp.targets(57,5): error MSB3073: The command "npx autorest@ --max-memory-size=8192 --skip-csproj --skip-upgrade-check --version=3.9.7 D:\a\_work\1\s\azure-sdk\sdk\resourcehealth\Azure.ResourceManager.ResourceHealth\src/autorest.md --use=C:\Users\cloudtest\.nuget\packages\microsoft.azure.autorest.csharp\3.0.0-alpha.20240519.1\buildMultiTargeting\../tools/net7.0/any/ --clear-output-folder=true --shared-source-folders="D:\a\_work\1\s\azure-sdk\eng\/../sdk/core/Azure.Core/src/Shared/;C:\Users\cloudtest\.nuget\packages\microsoft.azure.autorest.csharp\3.0.0-alpha.20240519.1\buildMultiTargeting\../content/Generator.Shared/" --output-folder=D:\a\_work\1\s\azure-sdk\sdk\resourcehealth\Azure.ResourceManager.ResourceHealth\src/Generated --namespace=Azure.ResourceManager.ResourceHealth" exited with code 1. [D:\a\_work\1\s\azure-sdk\sdk\resourcehealth\Azure.ResourceManager.ResourceHealth\src\Azure.ResourceManager.ResourceHealth.csproj]
```

[...]

PR #5009, opened 2024-08-30
Update node install instruction 

timotheeguerin:


piyushsh:
Hi @timotheeguerin 
Has autorest been fixed to support Node.js V18+?
Currently there is an error thrown by autorest when run with Node.js v18+?
I see there is [an issue](https://github.com/Azure/autorest/issues/4968) logged for this as well.

mikeharder:
> Hi @timotheeguerin Has autorest been fixed to support Node.js V18+? Currently there is an error thrown by autorest when run with Node.js v18+? I see there is [an issue](https://github.com/Azure/autorest/issues/4968) logged for this as well.

@piyushsh: Autorest should work on all supported LTS versions of Node, which is currently Node 18|20|22.  Is #4968 the only problem on newer Node versions?

mikeharder (approved):

PR #5022, opened 2024-09-24
Fix issues in converter tool

pshao25:
1. update ruleset in tspconfig.yaml
2. Some operations are not calculated as resource operations
3. Some duplicate operations in the same interface
4. interface Operations extends Azure.ResourceManager.Operations should be generated according to swagger
5. Take variables into consideration when calculating parent.


packages/extensions/openapi-to-typespec/src/transforms/transform-arm-resources.ts
```diff
@@ -704,8 +711,24 @@ function getTspOperations(armSchema: ArmResourceSchema): [TspArmResourceOperatio
   return [tspOperations, normalOperations];
 }
 
-function getOperationName(name: string): string {
-  return _.lowerFirst(_.last(name.split("_")));
+const nameCache: { [resourceName: string]: Set<string> } = {};
+
+function getOperationName(resourceName: string, operationId: string): string {
+  let operationName = _.lowerFirst(_.last(operationId.split("_")));
+  if (resourceName in nameCache) {
```
live1206, line 718: This part seems strange to me, with the cache in place, we still need to do some calculation from the cache out value.

packages/extensions/openapi-to-typespec/src/resource/utils.ts
```diff
@@ -79,8 +79,21 @@ export function isSingleton(set: OperationSet): boolean {
...
   const lowerPath2 = path2.toLowerCase();
-  // TO-DO: escape the variable case
-  return lowerPath1.includes(lowerPath2);
+  const segments1 = lowerPath1.split("/");
+  const segments2 = lowerPath2.split("/");
+  if (segments2.length > segments1.length) return false;
+
+  for (let index = 0; index < segments2.length; ++index) {
+    if (isVariable(segments1[index])) {
+      if (!isVariable(segments2[index])) return false;
```
live1206, line 88: Is it because the variable name can be different from the paths?

packages/extensions/openapi-to-typespec/src/transforms/transform-arm-resources.ts
pshao25: Yes, if the name is in cache, it means this name is already taken, so we need another name.

packages/extensions/openapi-to-typespec/src/resource/utils.ts
pshao25: Yes

packages/extensions/openapi-to-typespec/src/transforms/transform-arm-resources.ts
```diff
@@ -704,8 +711,24 @@ function getTspOperations(armSchema: ArmResourceSchema): [TspArmResourceOperatio
   return [tspOperations, normalOperations];
 }
 
-function getOperationName(name: string): string {
-  return _.lowerFirst(_.last(name.split("_")));
+const existingNames: { [resourceName: string]: Set<string> } = {};
+// TO-DO: Figure out a way to create a new name if the name exists
```
live1206, line 715: I think we want to align the logic with .NET SDK generator

live1206 (approved):

packages/extensions/openapi-to-typespec/src/transforms/transform-arm-resources.ts
pshao25: Yes, let's see later.

PR #5027, opened 2024-10-15
upgrade tsp compiler to 0.61

pshao25:
1. Upgrade tsp compiler to 0.61
2. Fix paging response issues by:
2.1 if user customize the paging response then use the customized mode.
2.2 if user defined model is the same as ResourceListResult, then replace it by ResourceListResult
2.3 if interface name conflicts with model name, then rename interface name
3. Support customize LRO
4. Support ArmProviderActionAsync template.


packages/extensions/openapi-to-typespec/src/interfaces.ts
```diff
@@ -222,6 +222,18 @@ export interface TspArmResourceOperationBase extends WithDoc, WithFixMe, WithSup
   customizations?: string[];
 }
 
+export interface TspArmProviderActionOperation extends WithDoc, WithSummary {
+  kind: "ArmProviderActionAsync";
+  name: string;
+  action: string | undefined;
```
live1206, line 228: can it just be `action?: string`?

packages/extensions/openapi-to-typespec/src/utils/type-mapping.ts
```diff
@@ -1,3 +1,6 @@
+import { isObjectSchema, Response, SchemaResponse } from "@autorest/codemodel";
```
live1206, line 1: this is not in use

packages/extensions/openapi-to-typespec/src/generate/generate-operations.ts
```diff
@@ -59,6 +71,55 @@ export function generateOperation(operation: TypespecOperation, operationGroup?:
...
+  const doc = generateDocs(operation);
+  const summary = generateSummary(operation);
+  const statements: string[] = [];
+  summary && statements.push(summary);
+  statements.push(doc);
+  const templateParameters = [];
+  const responses = [...new Set(operation.responses)];
+  // Workaround for array response, refactor later.
+  const response =
+    responses.length === 1 && responses[0].endsWith("[]") ? `{@body _: ${responses[0]}}` : responses.join("|");
```
live1206, line 84: what if it is not array?

packages/extensions/openapi-to-typespec/src/interfaces.ts
```diff
@@ -222,6 +222,18 @@ export interface TspArmResourceOperationBase extends WithDoc, WithFixMe, WithSup
   customizations?: string[];
 }
 
+export interface TspArmProviderActionOperation extends WithDoc, WithSummary {
```
live1206, line 225: add a TODO to consolidate this with other options?

live1206 (approved):

packages/extensions/openapi-to-typespec/src/generate/generate-operations.ts
pshao25: Then there will not have the `{@body}` wrapper.

packages/extensions/openapi-to-typespec/src/interfaces.ts
pshao25: We can.

packages/extensions/openapi-to-typespec/src/interfaces.ts
```diff
@@ -222,6 +222,19 @@ export interface TspArmResourceOperationBase extends WithDoc, WithFixMe, WithSup
   customizations?: string[];
 }
 
+// TO-DO: consolidate with other templates
+export interface TspArmProviderActionOperation extends WithDoc, WithSummary {
+  kind: "ArmProviderActionAsync";
+  name: string;
+  action?: string;
+  responses: string[] | undefined;
```
live1206, line 230: there are more cases in the same contract

PR #5050, opened 2025-01-06
Fix several issues in converter

pshao25:
1. `ResourceListResult` should be only replaced in arm.
2. Properties in TrackedResource/ProxyResource should not be in resource model.

### DIFF
--- a/common/changes/@autorest/openapi-to-typespec/2025-01-06-07-05.json
+++ b/common/changes/@autorest/openapi-to-typespec/2025-01-06-07-05.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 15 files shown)


common/changes/@autorest/openapi-to-typespec/2025-01-06-07-05.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@autorest/openapi-to-typespec",
+      "comment": "Fix several issues in converter",
```
live1206, line 5: consider to put the actual fix in the comment

live1206 (approved):

PR #5058, opened 2025-01-22
Add Azure.Core import if using Oauth2

pshao25:
As titled.

### DIFF
--- a/packages/extensions/openapi-to-typespec/src/emiters/emit-main.ts
+++ b/packages/extensions/openapi-to-typespec/src/emiters/emit-main.ts
@@ -1,9 +1,9 @@
-import { getMainImports } from "../utils/imports";
 import { getSession } from "../autorest-session";
 import { generateServiceInformation } from "../generate/generate-service-information";
 import { TypespecProgram } from "../interfaces";
 import { getOptions } from "../options";
 import { formatTypespecFile } from "../utils/format";
+import { getMainImports } from "../utils/imports";
 import { Metadata } from "../utils/resource-discovery";
 const packageInfo = require("../../package.json");
 
--- a/packages/extensions/openapi-to-typespec/src/emiters/emit-main.ts
+++ b/packages/extensions/openapi-to-typespec/src/emiters/emit-main.ts
@@ -1,3 +1,4 @@
+import { getMainImports } from "../utils/imports";
 import { getSession } from "../autorest-session";
 import { generateServiceInformation } from "../generate/generate-service-information";
 import { TypespecProgram } from "../interfaces";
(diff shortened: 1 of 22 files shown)


packages/extensions/openapi-to-typespec/src/emiters/emit-main.ts
```diff
@@ -37,18 +38,19 @@ function getHeaders() {
...
-    `import "./routes.tsp";`,
-    ``,
-    `using TypeSpec.Rest;`,
-    `using TypeSpec.Http;`,
-    `using TypeSpec.Versioning;`,
-  ];
-  const content = generateServiceInformation(program);
-
-  return [...imports, content].join("\n");
+    ...new Set<string>([`import "@typespec/rest";`, `import "@typespec/http";`, ...modules, `import "./routes.tsp"`]),
```
live1206, line 43: why do we need to wrap a set of string within the array?
And since we have `join("\n")` in L53, why do we need to have them here?

packages/extensions/openapi-to-typespec/src/emiters/emit-main.ts
```diff
@@ -37,18 +38,19 @@ function getHeaders() {
...
+  ].join("\n");
+  const namespaces = [
+    ...new Set<string>([
+      `using TypeSpec.Rest;`,
+      `using TypeSpec.Http;`,
+      `using TypeSpec.Versioning;`,
+      ...namespacesSet,
+    ]),
+  ].join("\n");
+  return [imports, "\n", namespaces, "\n", generateServiceInformation(program)].join("\n");
```
live1206, line 53: no spread of imports here?

pshao25: No need

packages/extensions/openapi-to-typespec/src/emiters/emit-main.ts
live1206: When get modules from getMainImports, the implementation already uses `Set<string>` to deal with the duplication removal.

packages/extensions/openapi-to-typespec/src/emiters/emit-main.ts
live1206: right.

packages/extensions/openapi-to-typespec/src/emiters/emit-main.ts
pshao25: Changed the inner `Set` to `Array`.

live1206 (approved):

PR #5065, opened 2025-01-27
Resource model envelope properties

pshao25:
1. If the common type is not in v3-v6, set it to v3, otherwise we cannot extend the resource model from arm library.
2. If the property in the resource model can be replaced by the envelop property from arm library, then use the envelope property, otherwise remain the same.

### DIFF
--- a/packages/extensions/openapi-to-typespec/src/generate/generate-enums.ts
+++ b/packages/extensions/openapi-to-typespec/src/generate/generate-enums.ts
@@ -2,17 +2,19 @@ import { TypespecEnum } from "../interfaces";
 import { getOptions } from "../options";
 import { generateDecorators } from "../utils/decorators";
 import { generateDocs } from "../utils/docs";
-import {
-  generateSuppressionForDocumentRequired,
-  generateSuppressionForNoEnum,
-  generateSuppressions,
-} from "../utils/suppressions";
+import { generateSuppressionForNoEnum, generateSuppressions, getSuppresssionWithCode } from "../utils/suppressions";
 
 export function generateEnums(typespecEnum: TypespecEnum) {
   const { isFullCompatible } = getOptions();
   const definitions: string[] = [];
   const doc = generateDocs(typespecEnum);
-  definitions.push(doc.length > 0 || !isFullCompatible ? doc : `${generateSuppressionForDocumentRequired()}\n`);
+  definitions.push(
+    doc.length > 0 || !isFullCompatible
+      ? doc
+      : `${generateSuppressions([
+          getSuppresssionWithCode("@azure-tools/typespec-azure-core/documentation-required"),
+        ])}\n`,
+  );
 
   const isExtensible = typespecEnum.isExtensible && !["ApiVersion"].includes(typespecEnum.name);
   if (!isExtensible && isFullCompatible) definitions.push(`${generateSuppressionForNoEnum()}\n`);
@@ -37,7 +39,10 @@ export function generateEnums(typespecEnum: TypespecEnum) {
             const doc = generateDocs(m);
             const kv = `"${m.name}": ${m.value}`;
             if (doc.length > 0 || !isFullCompatible) return `${doc}${kv}`;
-            else return `${generateSuppressionForDocumentRequired()}\n${kv}`;
+            else
+              return `${generateSuppressions([
+                getSuppresssionWithCode("@azure-tools/typespec-azure-core/documentation-required"),
+              ])}\n${kv}`;
           })
           .join(", ")}
     }\n\n`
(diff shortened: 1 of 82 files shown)


packages/extensions/openapi-to-typespec/test/arm-analysisservices/tsp-output/models.tsp
```diff
@@ -74,19 +74,6 @@ union ProvisioningState {
   Scaling: "Scaling",
 }
 
-/**
```
pshao25, line 77: Expected. We changed the behavior to setting common type version v3. so that we could leverage the models from arm library.

packages/extensions/openapi-to-typespec/test/arm-apimanagement/tsp-output/ApiManagementServiceResource.tsp
```diff
@@ -26,18 +26,25 @@ model ApiManagementServiceResource
...
-  zones?: string[];
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "For backward compatibility"
+  identity?: ApiManagementServiceIdentity;
+
+  /**
+   * ETag of the resource.
+   */
+  #suppress "@azure-tools/typespec-azure-resource-manager/arm-resource-invalid-envelope-property" "For backward compatibility"
+  @visibility("read")
+  etag?: string;
```
pshao25, line 45: Need to confirm if it could be replaced by `EntityTagProperty`, since in the `EntityTagProperty`, the name is `eTag`. Capital "T".

packages/extensions/openapi-to-typespec/test/arm-authorization/tsp-output/DenyAssignment.tsp
```diff
@@ -13,6 +13,7 @@ namespace Azure.ResourceManager.Authorization;
 /**
  * Deny Assignment
  */
+@extensionResource
```
pshao25, line 16: Expected. It is extension resource.

packages/extensions/openapi-to-typespec/test/arm-test/tsp-output/ArrayDefault.tsp
```diff
@@ -20,8 +20,7 @@ model ArrayDefault
   ...ResourceNameParameter<
     Resource = ArrayDefault,
     KeyName = "name",
-    SegmentName = "arraydefaults",
-    NamePattern = "^[a-zA-Z0-9-]{3,24}$"
+    SegmentName = "arraydefaults"
```
pshao25, line 23: Default value for `NamePattern` is `"^[a-zA-Z0-9-]{3,24}$"`

packages/extensions/openapi-to-typespec/src/generate/generate-enums.ts
```diff
@@ -37,7 +39,10 @@ export function generateEnums(typespecEnum: TypespecEnum) {
             const doc = generateDocs(m);
             const kv = `"${m.name}": ${m.value}`;
             if (doc.length > 0 || !isFullCompatible) return `${doc}${kv}`;
-            else return `${generateSuppressionForDocumentRequired()}\n${kv}`;
+            else
+              return `${generateSuppressions([
+                getSuppresssionWithCode("@azure-tools/typespec-azure-core/documentation-required"),
```
live1206, line 44: consider to declare a const string "@azure-tools/typespec-azure-core/documentation-required" to avoid the duplicates

packages/extensions/openapi-to-typespec/src/utils/common-type-mapping.ts
```diff
@@ -0,0 +1,186 @@
...
+  if (!isObjectSchema(schema) || schema.properties?.length !== 5) return false;
+
+  let nameFound = false;
+  let tierFound = false;
+  let sizeFound = false;
+  let familyFound = false;
+  let capacityFound = false;
+
+  for (const property of schema.properties) {
+    if (property.serializedName === "name" && property.required && property.schema.type === "string") {
```
live1206, line 15: Wondering how we define the replacement.
Here we check if schema contains all properties of Sku, but what if there is more properties in schema, can we replace it with Sku?

And also, instead of hard-code all the property name and types, is it possible to get the defined schemas and compare them instead?

pshao25: > Wondering how we define the replacement.

It should have same count of properties. For each property, same name, same type, same optionality. (I think we should consider more like visibility, pattern, etc. Will consider as an on demand improvement.)

> Here we check if schema contains all properties of Sku, but what if there is more properties in schema, can we replace it with Sku?

No. (We can improve this function like `containsSku` and generate like `{ otherproperty: string; ...Sku }` if satisfied)

> And also, instead of hard-code all the property name and types, is it possible to get the defined schemas and compare them instead?

No unless TypeSpec provides an API like `getProperties`. We discussed with TypeSpec team before for this. I think it's difficult for them to realize.

live1206: > No unless TypeSpec provides an API like getProperties. We discussed with TypeSpec team before for this. I think it's difficult for them to realize.

I mean we can create objects instead of hard-code properties, so that the comparison logic will be more general.

packages/extensions/openapi-to-typespec/test/arm-apimanagement/tsp-output/ApiManagementServiceResource.tsp
pshao25: This will be fixed in the next PR too.

PR #5069, opened 2025-02-07
Envelop property replacement with case sensitive

pshao25:
When we decide if a property will be replaced by an envelop property, we check name, optionality, readonly, and type. If the difference between names is only case, then we still replace with envelope property but add encodedName to keep its original serialized name.


packages/extensions/openapi-to-typespec/src/utils/library-type-mapping.ts
```diff
@@ -0,0 +1,250 @@
...
+}
+
+interface LibraryObjectType extends LibraryTypeBase {
+  type: "object";
+  properties: Record<string, LibraryProperty>;
+}
+
+interface LibraryArrayType extends LibraryTypeBase {
+  type: "array";
+  elementType: () => LibraryType;
```
live1206, line 36: can this just be 
```
elementType: LibraryType;
``
instead of function?

File: packages/extensions/openapi-to-typespec/src/utils/library-type-mapping.ts [hunk 2]
```diff
@@ -0,0 +1,250 @@
...
+
+interface LibraryEnumType extends LibraryTypeBase {
+  type: "enum";
+  sealed: boolean;
+  values: Record<string, string>;
+}
+
+interface LibraryDictionaryType extends LibraryTypeBase {
+  type: "dictionary";
+  valueType: () => LibraryType;
```
live1206, line 52: no need to be function?

live1206 (approved):

packages/extensions/openapi-to-typespec/src/utils/library-type-mapping.ts
```diff
@@ -0,0 +1,250 @@
...
+    if (!isObjectSchema(schema)) return false;
+    if (schema.properties?.length !== Object.keys(expected.properties).length) return false;
+
+    for (const property of schema.properties ?? []) {
+      if (property.serializedName in expected.properties) {
+        const expectedProperty = expected.properties[property.serializedName];
+        if ((property.required ?? false) !== expectedProperty.required) return false;
+
+        if (expectedProperty.type === "primitive") {
+          if (!expectedProperty.schema(property.schema)) return false;
```
live1206, line 73: seems like the schema function is always property.schema.type == expectedProperty.name?
can we get rid of the schema function?

PR #5105, opened 2025-05-07
Fix common type replacement logic

### LINKED ISSUE #5100: [openapi-to-tsp]: Common Type Replacement
Currently, the common type reference from M4 will be inlined and replaced by the definition itself. This means conversion tool don't have knowledge whether the model is a user defined model or a model from common type.

Expectation: if the model is user defined, we should convert to a normal model; if the model is from common type, we should convert to some reference to arm library.

pshao25:
Fix #5100

A known issue after this is: if user refers to a common type, we will use the common type in TypeSpec lib, however, if this common type has inner type, I cannot tell whether it is from common type, therefore I'll still generate it.

### DIFF
--- a/packages/extensions/openapi-to-typespec/src/autorest-session.ts
+++ b/packages/extensions/openapi-to-typespec/src/autorest-session.ts
@@ -5,6 +5,8 @@ let _session: Session<CodeModel>;
 let _armCommonTypeVersion: ArmCommonTypeVersion | undefined;
 let _userSetArmCommonTypeVersion: string;
 
+const commonTypeModels: string[] = [];
+
 export function setSession(session: Session<CodeModel>): void {
   _session = session;
 }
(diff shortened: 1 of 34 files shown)


packages/extensions/openapi-to-typespec/src/autorest-session.ts
```diff
@@ -5,6 +5,8 @@ let _session: Session<CodeModel>;
 let _armCommonTypeVersion: ArmCommonTypeVersion | undefined;
 let _userSetArmCommonTypeVersion: string;
 
+const commonTypeModels: string[] = [];
```
live1206, line 8: can we use Set<string>?

packages/extensions/openapi-to-typespec/src/autorest-session.ts
```diff
@@ -14,7 +16,21 @@ export function getSession(): Session<CodeModel> {
 }
 
 export function setArmCommonTypeVersion(version: string): void {
-  _userSetArmCommonTypeVersion = version;
+  if (_userSetArmCommonTypeVersion === undefined) {
+    _userSetArmCommonTypeVersion = version;
+  }
+}
+
+export function setArmCommonTypeModel(model: string): void {
```
live1206, line 24:
```suggestion
export function addArmCommonTypeModel(model: string): void {
```

packages/extensions/openapi-to-typespec/src/autorest-session.ts
```diff
@@ -14,7 +16,21 @@ export function getSession(): Session<CodeModel> {
...
+  if (!commonTypeModels.includes(model)) {
```
live1206, line 25: if we use Set<string>, no need to check includes

packages/extensions/openapi-to-typespec/src/utils/model-generation.ts
```diff
@@ -72,3 +82,64 @@ export function generateTemplateModel(templateModel: TypespecTemplateModel): str
 export function generateAdditionalProperties(properties: TypespecObjectProperty[]): string {
   return properties.map((p) => getModelPropertyDeclarations(p).join("\n")).join(";");
 }
+
+const commonTypeMapping: Record<string, string> = {
+  Resource: "Azure.ResourceManager.CommonTypes.Resource",
```
live1206, line 87: consider to remove the namespace if they are always the same

pshao25: I want to keep this in case they have different namespace.

PR #5120, opened 2025-05-21
Fix regression with json path failing on eval error

timotheeguerin:
A few fix:
- make sure the openapi to tsp use the local version of core and autorest to not catch error after the fact
- Fix the json path issue


packages/libs/datastore/src/json-path/json-path.ts
```diff
@@ -22,19 +21,6 @@ interface JsonPathResult {
   pointer: string;
 }
 
-// Override the vm used in jsonpath to use our safeEval and ignore errors.
-const safeEval = createSandbox();
```
timotheeguerin, line 26: I don't think we need that as json path fixed their vulnerability

mikeharder (approved):

